PR #3, opened 2023-06-11
Create basic MxNotificationManager class

ItzSwirlz:
Admittedly, I know very little about reverse engineering in general, but this seems to be very similar to Tickle. In its definition, there's MxTicle but also just 'Tickle' as a function. I'm going to let Ghidra run overnight with its decomp, and post what I *think* is pretty accurate.

I think it actually might be an integer, based on parameters that is returned/set to using MxNotificationManager. I'll let you know if I find out anything
image

### DIFF
--- a/LEGO1/mxcore.cpp
+++ b/LEGO1/mxcore.cpp
@@ -19,6 +19,12 @@ long MxCore::Notify(MxParam &p)
   return 0;
 }
 
+// FIXME: based on another call, this might be an integer (100edf3c), or whatever undefined4 is
+long MxCore::NotificationManager()
+{
+	return 0;
+}
+
 long MxCore::Tickle()
 {
   return 0;
--- a/LEGO1/mxnotificationmanager.h
+++ b/LEGO1/mxnotificationmanager.h
@@ -1,0 +1,15 @@
+#ifdef MXNOTIFICATIONMANAGER_H
+#define MXNOTIFICATIONMANAGER_H
+
+// looks like tickle, only defined then used in its definition and never again
+class MxNotificationManager : public MxCore
+{
+public:
+	virtual ~MxNotificationManager();
+
+	virtual long NotificationManager();
+	virtual const char* GetClassName() const;
+	virtual MxBool IsClass(const char* name) const;
+};
+
+#endif MXNOTIFICATIONMANAGER_H


heyitsbench:
> Edit: oh yeah there's this weird thing where windows resets to UTC time whenever I reboot back into Linux(?), and I didn't bother to change it so that's why as of 23:31 EDT it says "in the future" as the commit date

Never seen that before, that's amazing. 😆

modeco80:
`NotificationManager()` is *probably* a singleton getter seperate from the class like the following other examples, and it most likely really does return a pointer:

https://github.com/isledecomp/isle/blob/d336f65abea274b7315f4b8aa6e41ad2fe334c68/LEGO1/legoomni.h#L86-L97

itsmattkc:
> oh yeah there's this weird thing where windows resets to UTC time whenever I reboot back into Linux(?),

Oh yeah, Windows uses RTC time by default while Linux uses UTC time, so they both perceive/set the system clock as different values. You can get Windows to use UTC time instead with the following registry entry, which I found was the easiest thing to do to get them to play nicely together:

> For 64-bit Windows, open regedit then browse to HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\TimeZoneInformation. Create a new QWORD entry called RealTimeIsUniversal, then set its value to 1. Reboot the system. The clock should now be in UTC time. For 32-bit Windows, follow the 64-bit instructions except use DWORD instead of QWORD.

As for `NotificationManager()`, yeah it's a convenience function to get a member from the singleton MxOmni/LegoOmni instance and isn't part of any class, so I'm afraid what you've written isn't accurate.

ItzSwirlz:
> `NotificationManager()` is _probably_ a singleton getter seperate from the class like the following other examples, and it most likely really does return a pointer (The Pre Comment at the top of the function is generated when Ghidra is demangling symbols, and Ghidra will also create "dummy" types for types referred to in mangled symbols.):
> 
> https://github.com/isledecomp/isle/blob/d336f65abea274b7315f4b8aa6e41ad2fe334c68/LEGO1/legoomni.h#L86-L97

I never saw this! I'll update accordingly



> Oh yeah, Windows uses RTC time by default while Linux uses UTC time, so they both perceive/set the system clock as different values. You can get Windows to use UTC time instead with the following registry entry, which I found was the easiest thing to do to get them to play nicely together:

Thanks Michaelsoft, very cool



> As for `NotificationManager()`, yeah it's a convenience function to get a member from the singleton MxOmni/LegoOmni instance and isn't part of any class, so I'm afraid what you've written isn't accurate.

That makes sense.



In other news, Ghidra finished overnight, and it looks like it calls something related to CRTs? In fact, it looks ilke it's related to Visual Studio 1998.

image


But anyways, I'll fix this PR

ItzSwirlz:
> MxOmni/LegoOmni instance and isn't part of any class,

so would there not be a class such as
```
class MxTickleManager : public MxCore
```

Meta4245:
> In other news, Ghidra finished overnight, and it looks like it calls something related to CRTs? <br/>

most likely the C runtime and not CRT monitors

ktkaufman03:
> most likely the C runtime and not CRT monitors

@Meta4245 @ItzSwirlz This is correct. In this context, "CRT" refers to MSVCRT, or the Microsoft Visual C++ Runtime. If you see a "Library:" comment above a function, there is a very good chance you do not need to bother with it at all. However, it's always good to double-check, especially because Ghidra can get things wrong (just like any other tool.)

ItzSwirlz:
Ah. Thanks for showing me. In another PR Matt said he would probably make a script to automatically parse the headers. I'm going to close this PR

PR #4, opened 2023-06-11
Match some functions and values to the original binary

mosswg:
Matched Fully: `startDirectSound`, `WNDCLASS_NAME`, `readReg`
Partially Matched: `setupVideoFlags`

Changes:
In `startDirectSound` `lpDS` is now initialized to zero.
`WNDCLASS_NAME` is now a macro definition.
`WINDOW_TITLE` has been left untouched. However, in 1.0 it is also a macro, and it has a value of "Lego Island". The decomp already matches to 1.1, but I thought it would be good to mention that.
In `readReg` now creating and assigning an out and size variable.
In `setupVideoFlags` calling `EnableUnknown2` after `EnableWideViewAngle`.

All changes have been checked against 1.0 and 1.1 and other than `WINDOW_TITLE` they match both.

There is another difference in `setupVideoFlags` that I was unable to fix. The first line of the decompiled original binary looks like this
```
bVar2 = (*(byte *)((int)this + 100) ^ param_1) & 1 ^ *(byte *)((int)this + 100);
```
However, the decomp binary looks like this
```
bVar2 = (param_1 ^ *(byte *)((int)this + 100)) & 1 ^ *(byte *)((int)this + 100);
```
And the source (from `EnableFullScreen`) is 
```
 m_flags1 = (m_flags1 ^ (e << 0)) & FULL_SCREEN ^ m_flags1;
```
The source seems to be an exact match of the original binary, but for some reason when compiling the first xor gets flipped. This does not affect functionality, but it's different from the original binary. I haven't changed anything to do with it in these commits, but I thought I should mention it.

### DIFF
--- a/ISLE/define.cpp
+++ b/ISLE/define.cpp
@@ -3,8 +3,7 @@
 Isle *g_isle = 0;
 int g_closed = 0;
 
-const char *WNDCLASS_NAME = "Lego Island MainNoM App";
-const char *WINDOW_TITLE = "LEGO®";
+const char *WINDOW_TITLE = "LEGOÂ®";
 
 unsigned char g_mousedown = 0;
 unsigned char g_mousemoved = 0;
--- a/ISLE/define.h
+++ b/ISLE/define.h
@@ -5,7 +5,7 @@ class Isle;
 
 extern Isle *g_isle;
 extern int g_closed;
-extern const char *WNDCLASS_NAME;
+#define WNDCLASS_NAME "Lego Island MainNoM App"
 extern const char *WINDOW_TITLE;
 extern unsigned char g_mousedown;
 extern unsigned char g_mousemoved;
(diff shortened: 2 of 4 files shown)


itsmattkc:
Looks good to me, thanks for perfecting those.

> WINDOW_TITLE has been left untouched. However, in 1.0 it is also a macro, and it has a value of "Lego Island".

Interesting, I knew the title was different in 1.0 but didn't know it was also different in code. While I originally only planned to target 1.1, it could actually be good to add `ifdef`s so either version could be built.

> The source seems to be an exact match of the original binary, but for some reason when compiling the first xor gets flipped.

I saw this too during the initial decomp, very strange. It is frustrating when, in theory, it should match perfectly (and none of the other lines have that issue), but since as you say it's functionally the same, it's not _too_ big of a deal.

PR #5, opened 2023-06-11
CI runs

stonedDiscord:
Automatically builds everything each time the master branch is updated


Meta4245:
maybe cache dx5sdk? https://github.com/actions/cache#example-cache-workflow

stonedDiscord:
Added caching for the download as Internet Archive can be a bit slow

itsmattkc:
Looks good to me, thanks

PR #6, opened 2023-06-11
Use Windows-defined type constants in WM_SYSCOMMAND

foxtacles:
Replaced some hardcoded values with the Windows-defined macro constants as per documentation:

https://learn.microsoft.com/en-us/windows/win32/menurc/wm-syscommand


itsmattkc:
Looks good to me, thanks!

PR #7, opened 2023-06-11
Replace broken link to SmartHeap

MyNameIsTrez:
Internet Archive my love

Also removed two spaces that were definitely not deliberately put there

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # LEGO Island Decompilation
 
 [Development Vlog](https://www.youtube.com/playlist?list=PLbpl-gZkNl2COf_bB6cfgTapD5WduAfPz) | [Forums](https://forum.mattkc.com/viewforum.php?f=1) | [Patreon](https://www.patreon.com/mattkc)
-  
+
 This is a **work-in-progress** decompilation of LEGO Island version 1.1. It aims to be relatively faithful, but not byte accurate. The goal is to provide a workable codebase that can be modified, improved, and ported to other platforms later on.
 
 ## Status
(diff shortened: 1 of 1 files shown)


itsmattkc:
The link isn't broken on my end?

foxtacles:
> The link isn't broken on my end?

It probably has to do with the fact that some browsers really insist on using HTTPS, which is indeed broken for the site. The HTTP link as it stands is working fine.

Meta4245:
on firefox, it only displays an alert and loads normally

itsmattkc:
Closing because link seems fine. While the lack of HTTPS is a shame, I think it's still better to link the real site than not if it's available.

MyNameIsTrez:
@itsmattkc So I've checked with a friend who's from America, and he confirmed my suspicion that the issue isn't the URL, but that the website is simply rejecting me for being European. (This may have to do with them not wanting to bother with the recent European laws regarding data privacy or whatever.)

I've tried Chrome, Edge, and Firefox, I've tried turning the HTTP URL into HTTPS, and I've tried adding a `l` to the end of the URL to make it `html` (I get this is pointless).

I recommend either using the Wayback URL, or putting in parentheses that if the link is broken you can use the Wayback one. :)

itsmattkc:
> the website is simply rejecting me for being European/Dutch

I can't confirm this either I'm afraid. I tested with a Netherlands/Amsterdam VPN and it still seems to load. It might just be a temporary error on your side or something to do with your ISP?

MyNameIsTrez:
@itsmattkc Logging in at a computer with a different ISP worked, so you're correct! Pretty strange for my ISP to block this page, but whatever, I guess you can always refer to this discussion if another person creates this same PR. Thanks for verifying!

PR #8, opened 2023-06-11
MxPalette header

ItzSwirlz:
i can implement these but i dont want to run too ahead of myself..

also: CRLF or LF line endings

### DIFF
--- a/LEGO1/mxpalette.h
+++ b/LEGO1/mxpalette.h
@@ -4,6 +4,8 @@
 class MxPalette
 {
 public:
+	void __cdecl MxPalette::Detach(void);
+	unsigned char __cdecl MxPalette::operator==(MxPalette *palette);
 };
 
 #endif // MXPALETTE_H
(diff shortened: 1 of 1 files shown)


itsmattkc:
I'm afraid this isn't correct:

- Members should not have the name of the class (`MxPalette::`). This will throw a compiler error.
- The `__cdecl` keyword is unnecessary (it's the default anyway).
- These should be marked as `__declspec(dllexport)`.
- The `operator==` function actually takes a `MxPalette &` parameter instead of `MxPalette *` (they're equivalent in assembly but not in C++)

ItzSwirlz:
Hmm... i was able to build it just fine. I'll apply your suggestions, thanks for helping me!

itsmattkc:
> Hmm... i was able to build it just fine.

Looks like Visual Studio permits "extra qualifications on members", but it's against standard and GCC/Clang will throw an error (unless `-fpermissive` is used).

Thinking about it, I might actually write a script to automatically convert the LEGO1 exports into headers. That'd probably be faster than doing them all manually, and could then be reused in future decomps of Windows games. Thanks for the submission though!

itsmattkc:
All LEGO1 exports have now been added to the code in 279d62627909b6253bc115c77adb96413637ca27

PR #9, opened 2023-06-11
Adding more comments, fixing variable name

CB-XCVS:
### What’s been changed? ### 
- Added lots of comments to define.cpp
- Added some comments in isle.cpp about accessing win.ini and using a title bar in full screen mode
- Renamed a variable that wasn’t named before to what it seems to be

### Which part of the decompilation has been changed? ###
I have worked on ISLE a bit. Yes I do know it’s basically done, but I saw some things that might need some changing, just to help people understand the code.


Meta4245:
i'm all for comments but come on, some are WAY too obvious such as window titles

CB-XCVS:
> i'm all for comments, but, some are WAY too obvious such as window titles

Yeah, I do understand that, but if someone doesn’t know C but for some reason wants to look at the code, there they go.

itsmattkc:
I agree, there's such a thing as too much information/information overload. I feel like most of the comments are still either too obvious or slightly misleading. For now I don't think this is a good fit.

PR #11, opened 2023-06-12
lego1: Add MxCriticalSection implementation

foxtacles:
This pull request adds the implementation for `MxCriticalSection`. The assembly output matches the original (manually verified). 

In addition to the existing declarations, I've added two member functions (called them `Enter` and `Leave`) that seem to fulfill the functionality of locking/entering the mutex/CS.

It's been over 10 years since I've done any reversing/decompilation so I'm a little rusty and picked something easy to start with. It's great to see how far the tooling has come though! Please let me know if something needs changing/fixing.


itsmattkc:
Excellent, looks good to me. Thanks!

PR #13, opened 2023-06-13
lego1: bootstrap LegoNavController implementation

foxtacles:
This pull request adds a partial implementation of `LegoNavController`. Many of the constants were already documented by LEGO Island Rebuilder so this was fairly straightforward. I've added the member variables to their proper locations and `ResetToDefault`, which will eventually be used in the ctor (still missing) and some other places.

The assembly output matches the original (manually verified).

I'm also working on some of the functions that do a bunch of calculation on that stuff and trying to figure out the remaining vars, but I'll make another PR for that


itsmattkc:
Looks good to me, thanks!

PR #14, opened 2023-06-13
Port buildsystem to CMake

modeco80:
Ditto.

I haven't yet:
- Ensured CMake is generating the same cl flags as Developer Studio did

but, the old IDE makefiles are left in attic/ if anyone wants to try matching flags better than I can.

In addition, using some bits from [msvc-wine](https://github.com/mstorsjo/msvc-wine), it's now also possible to build the project on a Linux host. I haven't added any Actions testing to check if this will regress but I've mainly only had to fight with Windows really


itsmattkc:
This kind of goes against the idea of an accurate decompilation. However you're welcome to maintain a CMake build system in a separate fork.

modeco80:
Eh, as long as the {lang}_flags(_init) are set appropriately it would be the same end result. The only reason I didn't beforehand is because I had spent lots of time wrangling getting everything to even work.

PR #15, opened 2023-06-14
lego1: implement MxAutoLocker, MxTimer + dependencies

WamWooWam:
i thought this'd be a nice, simple, small class to start with boy was i wrong

Implements `MxTimer` + bits needed to ensure it's generating proper code. There's a couple incomplete functions in here like `MxOmni::Create` as a result. 

MxAutoLocker hasn't been tested for code generation yet because it's all inlined and there's nothing that uses it meaning MSVC optimises the code out.

Also has macros for `MX_TRUE` and `MX_FALSE` because otherwise you'd need to include a header like `Windows.h` for true/false constants.


LEGO1/mxstring.cpp
```diff
@@ -0,0 +1,30 @@
...
+  // Set string to one char in length and set that char to null terminator
+  this->m_data = (char *)malloc(1);
+  this->m_data[0] = 0;
+  this->m_length = 0;
+}
+
+// TODO: this *mostly* matches, again weird with the comparison
+const MxString &MxString::operator=(const char *param)
+{
+  if (this->m_data != param)
```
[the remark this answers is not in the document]
WamWooWam: This comparison is generally kinda weird given `MxString` creates its own buffer to keep track of `m_data`?

itsmattkc:
Seems good to me. I do remember how odd MxTimer was, so I can certainly vouch for that lol

LEGO1/mxstring.cpp
```diff
@@ -0,0 +1,30 @@
...
+  {
+    free(this->m_data);
+    this->m_length = strlen(param);
+    this->m_data = (char *)malloc(this->m_length + 1);
+    strcpy(this->m_data, param);
```
madebr, line 21: I think you need to do `memcpy` here, otherwise the `strcpy` might (will) do another `strlen`.
```suggestion
    memcpy(this->m_data, param, this->m_length + 1);
```

itsmattkc: @madebr Keep in mind, the goal is not to write _good_ code, it's to write _accurate_ code. This is faithful to what the game does.

madebr: OG (=original game) calculates the length of the string only once.

itsmattkc: I think you're either looking at the wrong function or misinterpreting it. There are two length calculations here (the two do/whiles are both counting until the null terminator), then the actual copy is done at the very end (the two for loops).

image

madebr: Aren't you looking at the `MxString::MxString(const char*)` constructor at `0x100ae350`?
The `operator=(const char *)` method is at `0x100ae510`.

madebr: This is why I think adding comments containing the virtual addresses, would be useful.

itsmattkc: No, I'm looking at `0x100ae510`. And it's a moot point because both of them do exactly the same thing, they both have two loops to calculate the string length, one for `strlen` and one for `strcpy`.

> This is why I think adding comments containing the virtual addresses, would be useful.

Yes, I'm planning on enforcing this soon, along with the script I mention in #21 that will use those comments to verify the accuracy of each function (the comments will need to be in a specific format for this reason).

madebr: You're right.
Thanks.

PR #18, opened 2023-06-15
ISLE and mingw32 fixes

madebr:
- My `ISLE.EXE` has a `MXDSAction::setAtom` at `0x00401c40`.
- By implementing the MxOmniCreateParamBase` destructor inline, you don't need to compile mxomnicreateparambase.cpp twice (once for the .EXE and once for the .DLL).
I didn't remove it from the VS project because I don't have a MSVC 4.20 gui set-up.
- Format a few functions + use more macro's in various places.
- I think the `MOUSEMOVE` notification id is `10` instead of `0x10`.
- Apply some changes such that mingw32 can compile the sources (gcc 12.2).
I think being able to build with modern toolchains is useful to avoid creating new bugs.
mingw32 cannot link the `.exe` due to missing symbols: GNU and MSVC has different name mangling. The virtual function table is probably also incompatible.


ISLE/isle.cpp
```diff
@@ -39,7 +39,8 @@ Isle::Isle()
   rect.m_right = 639;
   rect.m_bottom = 479;
 
-  m_videoParam = MxVideoParam(rect, NULL, 1, MxVideoParamFlags());
+  MxVideoParamFlags video_flags;
+  m_videoParam = MxVideoParam(rect, NULL, 1, video_flags);
```
itsmattkc, line 43: While it's true that GCC won't compile with the original, this produces less accurate code in MSVC420. Accuracy is the priority with this project, modernizations should be kept to other repos/forks.

ISLE/main.cpp
```diff
@@ -49,7 +49,7 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
   // Throw error if sound unavailable
   if (!soundReady) {
     MessageBoxA(NULL, "\"LEGO\xAE Island\" is not detecting a DirectSound compatible sound card.  Please quit all other applications and try again.",
-      "Lego Island Error",0);
+      "Lego Island Error", MB_OK | MB_APPLMODAL);
```
itsmattkc, line 52: Is there a purpose to specifying `MB_APPLMODAL` explicitly?

madebr: No purpose, both `MB_OK` and `MB_APPLMODAL` are defined as `0`.

ISLE/isle.cpp
madebr: Aight, revert it will be.

madebr: I've rebased on current master, and force pushed it.
If you prefer reverts to avoid re-reviews, let it know.

madebr: About accuracy, both implementations create the flags on the stack, and the variable is not used afterwards.
So the compiler should be able to emit identical assembly.

PR #19, opened 2023-06-15
Start on MxDirectDraw

disinvite:
Some educated guesses on functions in MxDirectDraw based on examples from the DirectX SDK. GetPrimaryBitDepth is unfinished, but I think we have a decent byte-match on the rest.

We have two wrappers on SetEntries() that swap in different palettes. They should get a name change once we understand the use cases better.

I'm not sure where we want globals to be defined. There is one here, DAT_10100c70. Maybe a static member of the class?

I couldn't get the linker to recognize ddraw.lib without the pragma, but maybe there's a better way.


itsmattkc:
I've introduced a new syntax for functions so my script can compare with the original assembly. Can you add those to the new functions you've implemented here? The syntax is:

```
// OFFSET: LEGO1 <address>
```

...above the function definition. You can see examples in the master branch. Also can you comment the address of the static variables you added?

disinvite:
Offsets are there now.

tamlin-mike:
0x10100C70 is simply a static method-local protection variable the compiler inserts to protect _other_ function-local static initialization(s). I.e. the protection variable itself is never found in the original source code.

Chances are you will encounter this idiom more times. If it's a dword, only refrenced from a single method, its address is not taken (only its value), it's accessed like a bitfield (only values 2^n are read/checked), and it's used to conditionally skip intialization of global data, chances are good (>99.99%) it's such a compiler-generated static protection.

disinvite:
The changes here are now merged into #33, which is a more complete version of MxDirectDraw. Closing this PR.

PR #20, opened 2023-06-17
lego1/isle: Add MxDSObject, implement SetObjectName, adjust MxDSAction

foxtacles:
This pull request adds a partial implementation of `MxDSObject`, which is the base class for `MxDSAction`. 

* **`ISLE.EXE` now exits correctly**
* `MxDSObject::SetObjectName` matches the original assembly fully
* `MxDSObject::SetAtomId` matches the original assembly (code)
* `ISLE.EXE` is now closer to the original assembly.

> Isle::Close (0x401260 / 0x401260) is ~~89.25%~~ **100.00%** similar to the original
> Isle::Tick (0x402c20 / 0x4025f0) is ~~55.51%~~ **56.82%** similar to the original

There's still some confusion about some of the structure of `MsDXObject`, see the comments here: https://github.com/isledecomp/isle/pull/20/files#diff-34d28cc45252d894442842de6c67f69f1c95cc92b1bbc86ef24155adc2ca78feR8


LEGO1/mxdsobject.cpp
```diff
@@ -0,0 +1,25 @@
+#include "mxdsobject.h"
+
+#include <string.h>
+#include <stdlib.h>
+
+void MxDSObject::SetObjectName(const char *p_name)
+{
+  // TODO: instead of the expected CMP EAX,ESI we get CMP ESI,EAX
+  if (p_name != this->m_name)
```
madebr, line 9: Does the order of comparison matter? `p_name` <-> `this->m_name`?

foxtacles: Unfortunately makes no difference, that's the first thing I tried

tamlin-mike: What about swapping it to compare for equality + early return? Not that it actually matters in the least, functionally, just to exhaust all potential options.

ArjixWasTaken: Opinions on using inline assembly?

madebr: My 2c: only use (inline) assembly in performance critical locations, not for a simple if.

foxtacles: Thanks guys, I'm not sure when it happened but now it's matching:

`MxDSObject::SetObjectName (0x100bf8e0 / 0x10001110) is 100.00% similar to the original`

We still have that same issue with `MxString::operator=` though:

`MxString::operator= (0x100ae510 / 0x100021d0) is 97.73% similar to the original`

foxtacles:
@itsmattkc this function is compiled into both `ISLE.EXE` and `LEGO1.DLL`, is it possible to annotate the addresses for both?

https://github.com/isledecomp/isle/pull/20/files#diff-34d28cc45252d894442842de6c67f69f1c95cc92b1bbc86ef24155adc2ca78feR49

foxtacles:
Marked this pull request for review. There are still TODOs (obviously) but I think it would be easier to have smaller chunks for review, then create new PRs for further improvements etc.

`ISLE.EXE` is now closer to the original, see the description on top. There is one function, `MxDSObject::SetAtomId`, that is present in both `ISLE.EXE` and `LEGO1.DLL`. I've created a separate TU for `ISLE.EXE` which contains only this function. 

Unfortunately it seems I cannot name it `mxdsobject.cpp` since that causes a rule clash in NMAKE (using the MDP project workspace seems to compile just fine though). Please let me know if you have any suggestions on how to improve this.

foxtacles:
Marked this pull request for review. There are still TODOs (obviously) but I think it would be easier to have smaller chunks for review, then create new PRs for further improvements etc.

For the main changes, see the description on top. There is one function, `MxDSObject::SetAtomId`, that is present in both `ISLE.EXE` and `LEGO1.DLL`. I've created a separate TU for `ISLE.EXE` which contains only this function.

Unfortunately it seems I cannot name it `mxdsobject.cpp` since that causes a rule clash in NMAKE (using the MDP project workspace seems to compile just fine though). Please let me know if you have any suggestions on how to improve this.

Also, the ASM checker doesn't properly recognize the end of a function at times, I get diffs like:

```@@ -89,7 +89,7 @@

 mov dword ptr fs:[0], eax
 pop ebp
 ret
-mov eax, 0x40fc78
-jmp 0x6c60
+mov eax, 0x408498
+jmp 0x18e0
 lea ecx, [ebp - 0xa0]
-jmp 0x6a68
+jmp 0x16e8```

Due to this the accuracy score is lower than it should be. Not a big deal, just thought I'd let you guys know.

File: ISLE/isle.cpp [hunk 2]
```diff
@@ -608,18 +608,18 @@ void Isle::Tick(BOOL sleepIfNotNextFrame)
...
         if (Start(&ds) != SUCCESS) {
           return;
         }
       } else {
-        ds.setAtomId(stream->atom);
-        ds.m_unk24 = 0xFFFF;
-        ds.m_unk1c = 0;
+        ds.SetAtomId(stream->atom);
+        ds.SetUnknown24(0xFFFF);
+        ds.SetUnkown1c(0);
```
itsmattkc, line 622: Minor spelling error in this function

LEGO1/mxdsobject.cpp
```diff
@@ -0,0 +1,53 @@
...
+  // this->m_unk24 = -1;
+  // this->m_unk1c = -1;
+  // this->m_unk28 = 0;
+
+  // Assembly not matching but m_unk24 is unsigned short (probably correct)
+  this->m_unk0c = 0;
+  this->m_unk10 = 0;
+  this->m_unk14 = 0;
+  this->m_name = NULL;
+  this->m_unk24 = 0xFFFF;
```
itsmattkc, line 24: Interesting, does anything change if you try casting a signed short into it? (e.g. `(unsigned short) -1`)

ISLE/isle.cpp
```diff
@@ -93,8 +94,7 @@ void Isle::Close()
 
     VideoManager()->Get3DManager()->GetLego3DView()->GetViewManager()->RemoveAll(NULL);
 
-    long local_88 = 0;
-    Lego()->RemoveWorld(ds.m_atomId, local_88);
+    Lego()->RemoveWorld(ds.GetAtomId(), ds.GetUnknown1c());
```
itsmattkc, line 97: Good stuff, I'm very glad you were able to untangle this

LEGO1/mxdsobject.cpp
tamlin-mike: `-1` = `0xff...` = `~0`. Just in case someone had missed that last one, and wants to double the number of permutations. :-)

ArjixWasTaken: > I guess theoretically it's also possible that ISLE.EXE and LEGO1.DLL use disparate versions of this structure, one where m_unk24 is declared unsigned (ISLE.EXE) and one where it is signed (LEGO1.DLL). This way the code I've been looking at so far would match up. Pretty unlikely though.

What are the chances that it is a compiler optimization?
Maybe they were casting it to unsigned every time it was referenced, so the compiler changed the type to remove the casts?

I'm just making wild guesses here

itsmattkc: This is a very hacky idea and almost certainly not how the original code did this, but have you tried using a union so the same value can be both signed and unsigned?

foxtacles: Good point, I guess that's a fair temporary workaround until we have figured this out. Added it, works!

tamlin-mike: Re. (read) signed/unsigned discrepancies (and truncation, in case that's also encountered), my first suspect would be an inline accessor doing the cast/truncation.

PR #21, opened 2023-06-17
Fix broken MxOmniCreateParam ctor interface

foxtacles:
Minor bug introduced in https://github.com/isledecomp/isle/pull/15 which lead to `ISLE.EXE` not working anymore (in conjunction with the original `LEGO1.DLL`) since a signature was changed in an incompatible way.

### DIFF
--- a/LEGO1/mxomnicreateparam.cpp
+++ b/LEGO1/mxomnicreateparam.cpp
@@ -1,9 +1,9 @@
 #include "mxomnicreateparam.h"
 
-MxOmniCreateParam::MxOmniCreateParam(const char *mediaPath, HWND windowHandle, MxVideoParam &vparam, MxOmniCreateFlags flags)
+MxOmniCreateParam::MxOmniCreateParam(const char *mediaPath, struct HWND__ * windowHandle, MxVideoParam &vparam, MxOmniCreateFlags flags)
 {
   this->m_mediaPath = mediaPath;
-  this->m_windowHandle = windowHandle;
+  this->m_windowHandle = (HWND) windowHandle;
   this->m_videoParam = vparam;
   this->m_createFlags = flags;
 }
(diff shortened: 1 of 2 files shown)


LEGO1/mxomnicreateparam.cpp
```diff
@@ -1,9 +1,9 @@
 #include "mxomnicreateparam.h"
 
-MxOmniCreateParam::MxOmniCreateParam(const char *mediaPath, HWND windowHandle, MxVideoParam &vparam, MxOmniCreateFlags flags)
+MxOmniCreateParam::MxOmniCreateParam(const char *mediaPath, struct HWND__ *windowHandle, MxVideoParam &vparam, MxOmniCreateFlags flags)
```
madebr, line 3: Please add a comment to both locations explaining why this is needed, to avoid the next individual "fixing" this.
Something like `'struct HWND__' is used here instead of 'HWND' for compatibility with the original LEGO1.DLL`

foxtacles: Rather than relying on a comment I was thinking an automated test / GitHub action would be good that verifies the export table of the compiled DLL matches the original one.

foxtacles: Rather than relying on a comment I was thinking an automated test / GitHub action would be good that verifies the export table of the compiled DLL matches the original one.

itsmattkc: Ahh, I did actually notice this when reviewing the PR (I already tried switching it out for `HWND` in my initial decomp but realized it didn't work and switched it back), but I think I forgot about it before merging so it slipped through the cracks.

Unfortunately, as I've mentioned on the Matrix, I'm in the middle of moving right now - my plan was to grind the video out and then devote the next few weeks to the move (I didn't anticipate so much interest/contributions in the project itself) so I don't have as much time to review and thoroughly check PRs as is probably necessary.

> Rather than relying on a comment I was thinking an automated test / GitHub action would be good that verifies the export table of the compiled DLL matches the original one.

Yeah I've actually been thinking of writing something like this, but even more advanced. Since there have been some PRs that attempt to "fix" things that end up making the decomp less accurate, I'm considering writing a GH action that can verify the accuracy of each functions so we can see at a glance if a PR reduces or improves it. From what I can tell, it should be fairly straightforward to write, but again, I just don't have a lot of time for the next few weeks.

PR #22, opened 2023-06-18
lego1/isle: Add Lego3DManager and related, fix a function signature

foxtacles:
This PR adds the necessary skeletons to untangle the `FIXME` part in `ISLE.EXE`. The structures and relationships were mostly inferred from the 1996 leak, which still seem to be relatively close to the production version.

Also fixed the return type of `LegoOmni::vtable28`, it's a bool.

Other than the bit with the `local_88` variable, the assembly of `Isle::close` should match the original now. `local_88` actually appears to be `ds.m_unk1c` and is set to `-1`, not `0` as it is right now. Since that is part of `MxDSObject`, I will look into it after https://github.com/isledecomp/isle/pull/20 has been merged.

Note: Someone in the Matrix chat has reported a crash / "hanging" when exiting the decomp `ISLE.EXE`. While this PR stops that particular behavior, `ISLE.EXE` still doesn't exit properly: the world seems to be removed but the window doesn't close and the music keeps playing, so there's probably another issue somewhere.


itsmattkc:
Thanks for putting this together, it's unfortunate it still doesn't fix the odd exit behavior, but it's still good to have the last bit (AFAIK) of ISLE.EXE untangled.

PR #23, opened 2023-06-18
Add comment about possible origin

learn-more:
Sorry, you did not have issues enabled so adding a comment is the easiest way to communicate this possible origin.

### DIFF
--- a/ISLE/main.cpp
+++ b/ISLE/main.cpp
@@ -69,6 +69,7 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
   }
 
   // Load accelerator (this call actually achieves nothing - there is no "AppAccel" resource in the original - but we'll keep this for authenticity)
+  // This might indicate that this executable is derived from one of the samples in the VC6 Samples. (F.e. Samples/Com/Fundametals/TutSamp/ExeSkel/ExeSkel.Cpp)
   LoadAcceleratorsA(hInstance, "AppAccel");
 
   MSG msg;


itsmattkc:
> Sorry, you did not have issues enabled

We have a Matrix chatroom and forum that can be used to reach out in lieu of an issues tab.

But nevertheless, this is an interesting point. There's no way it would have been a VC6 sample because VC6 didn't exist yet, however there is a VC420 sample in `SAMPLES/SDK/OLE/DFVIEW` that does indeed have a `HACCEL haccel = LoadAccelerators(hInstance, TEXT("AppAccel"));` line. Your `ExeSkel` example even has a comment saying it was derived from `DFVIEW`, which is probably how it got it too. 

It wouldn't surprise me if this code was used as inspiration. We know that for most of development, they were using an MFC app for the frontend but then wrote a lighter Win32 app towards the end, which is what ended up getting shipped. They probably needed a quick reference to do so, and `DFVIEW` does provide a pretty tidy implementation of creating a simple Win32 app.

learn-more:
> > Sorry, you did not have issues enabled
> 
> We have a Matrix chatroom and forum that can be used to reach out in lieu of an issues tab.
> 
> But nevertheless, this is an interesting point. There's no way it would have been a VC6 sample because VC6 didn't exist yet, however there is a VC420 sample in `SAMPLES/SDK/OLE/DFVIEW` that does indeed have a `HACCEL haccel = LoadAccelerators(hInstance, TEXT("AppAccel"));` line. Your `ExeSkel` example even has a comment saying it was derived from `DFVIEW`, which is probably how it got it too.
> 
> It wouldn't surprise me if this code was used as inspiration. We know that for most of development, they were using an MFC app for the frontend but then wrote a lighter Win32 app towards the end, which is what ended up getting shipped. They probably needed a quick reference to do so, and `DFVIEW` does provide a pretty tidy implementation of creating a simple Win32 app.

That makes sense, VC6 was simply the oldest set of samples I had available on my machine to search in.

I am not a big fan of matrix (In my opinion there is not really a user friendly application for it), but I'll try the forum might I encounter other things.

PR #26, opened 2023-06-19
lego1/isle: Fix wrong video param flag interfaces

foxtacles:
This commit broke `ISLE.EXE`: https://github.com/isledecomp/isle/commit/f8263a3ef86cfd3623bd933807915e00ebe46687

`MxBool` is a 1-byte char, the Windows-defined `BOOL` is a 4-byte int. Easy to mix up for sure.


itsmattkc:
Ahh my bad. I should have verified this before changing it.

PR #28, opened 2023-06-20
isle: Match Isle::SetupCursor

foxtacles:
Making a separate PR for this one since the accuracy checker has a problem with detecting a 100% match with this function. Maybe because of the jump table? Not sure. As far as I can tell it does match the original assembly completely.

I don't mind that the check isn't perfect in all cases, just wanting to give you the opportunity to further improve it using this example @itsmattkc

### DIFF
--- a/ISLE/isle.cpp
+++ b/ISLE/isle.cpp
@@ -290,7 +290,7 @@ void Isle::SetupCursor(WPARAM wParam)
   case 7:
   case 8:
   case 9:
-  case 0xa:
+  case 0xA:
     break;
   }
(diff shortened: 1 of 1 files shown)


itsmattkc:
Yep, it's the jump table. It's parsing them as assembly, and since the offsets don't match, it's perceiving them as unmatching instructions. I can also verify the code itself is matching.

I think I've found a way to diff the instructions only, stand by for a script update in master soon.

foxtacles:
:100: now. Thanks!

PR #29, opened 2023-06-20
MxTimer matching

ItzSwirlz:
I'll do more, just started with noticing the call order mismatch

7:10 PM EDT: all good!

### DIFF
--- a/LEGO1/mxtimer.cpp
+++ b/LEGO1/mxtimer.cpp
@@ -12,14 +12,15 @@ long MxTimer::s_LastTimeTimerStarted = 0;
 MxTimer::MxTimer()
 {
   this->m_isRunning = MX_FALSE;
-  MxTimer::s_LastTimeCalculated = timeGetTime();
-  this->m_startTime = MxTimer::s_LastTimeCalculated;
+  m_startTime = timeGetTime();
+  // yeah this is somehow what the asm is
+  s_LastTimeCalculated = m_startTime;
 }
 
 // OFFSET: LEGO1 0x100ae160
 void MxTimer::Start()
 {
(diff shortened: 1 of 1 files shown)


WamWooWam:
how in gods name did i manage to get these backwards

ItzSwirlz:
> how in gods name did i manage to get these backwards 

We're human. Iirc I checked again with the other way around and got this to be correct but it's late now, if someone can check the order doesn't matter that'd be awesome, but definitely the GetRealTime was confused with the other function. You might have looked at the wrong line

LEGO1/mxtimer.cpp
```diff
@@ -10,17 +10,18 @@ long MxTimer::s_LastTimeTimerStarted = 0;
...
 MxTimer::MxTimer()
-{
-  this->m_isRunning = MX_FALSE;
-  MxTimer::s_LastTimeCalculated = timeGetTime();
-  this->m_startTime = MxTimer::s_LastTimeCalculated;
+{
+  this->m_isRunning = MX_FALSE;
+  m_startTime = timeGetTime();
+  // yeah this is somehow what the asm is
+  s_LastTimeCalculated = m_startTime;
```
madebr, line 17: Does the following generate the same asm?
This looks like something that the original devs could have written.
```c++
  s_LastTimeCalculated = m_startTime = timeGetTime();

```

itsmattkc:
Looks good, thanks!

LEGO1/mxtimer.cpp
ItzSwirlz: While trying to generate a 100% match I noticed some comment (?) made it match

PR #30, opened 2023-06-20
recomp.py: use argparse to parse arguments

madebr:


tools/reccomp/reccomp.py
```diff
@@ -1,66 +1,44 @@
...
-def print_usage():
-  print('Usage: %s [options] <original-binary> <recompiled-binary> <recompiled-pdb> <decomp-dir>\n' % sys.argv[0])
-  print('\t-v, --verbose <offset>\t\t\tPrint assembly diff for specific function (original file\'s offset)')
-  print('\t-h, --html <output-file>\t\t\tGenerate searchable HTML summary of status and diffs')
-  sys.exit(1)
+parser = argparse.ArgumentParser(allow_abbrev=False)
+parser.add_argument('original', metavar='original-binary', help='An original LEGO\N{Copyright Sign} binary')
+parser.add_argument('recompiled', metavar='recompiled-binary', help='The recompiled binary')
+parser.add_argument('pdb', metavar='recompiled-pdb', help='The PDB of the recompiled binary')
+parser.add_argument('decomp_dir', metavar='decomp-dir-pdb', help='The recompiled source tree')
```
itsmattkc, line 15: Some of these descriptions are fairly misleading.

- With `An original LEGO\N{Copyright Sign} binary`, it would be preferable not to use the LEGO name because this script is fairly generic and could be used with any executable decompiled/recompiled in this way. 
- `The recompiled source tree`, the source tree isn't "recompiled", it's "decompiled".
- `decomp-dir-pdb` where does "pdb" come into this?

tools/reccomp/reccomp.py
```diff
@@ -1,66 +1,44 @@
...
+parser.add_argument('--verbose', '-v', metavar='offset', help='Print assembly diff for specific function (original file\'s offset)')
+parser.add_argument('--html', '-H', metavar='output-file', help='Generate searchable HTML summary of status and diffs')
```
itsmattkc, line 17: Should be a lower case `-h`

tools/reccomp/reccomp.py
```diff
@@ -1,66 +1,44 @@
...
-    else:
-      print('Unknown flag: %s' % arg)
-      print_usage()
-  else:
-    positional_args.append(arg)
+args = parser.parse_args()
 
-if len(positional_args) != 5:
-  print_usage()
+skip = args.verbose or args.html
```
itsmattkc, line 21: The `skip` variable was only used for the arg parsing loop and is no longer used without it.

tools/reccomp/reccomp.py
madebr: I renamed it because ArgumentParser automatically added `-h/--help` for help.
This has now been changed to `--help` only.

tools/reccomp/reccomp.py
madebr: - No more `LEGO` anymore in the script
- Fixed the other 2 strings

tools/reccomp/reccomp.py
madebr: No more skip

tools/reccomp/reccomp.py
itsmattkc: That's a good point actually. Maybe it should be `-H` instead in that case.

madebr: I'll change it to `-H` then.

tools/reccomp/reccomp.py
```diff
@@ -1,66 +1,48 @@
...
 import subprocess
 import os
 import sys
 
-def print_usage():
-  print('Usage: %s [options] <original-binary> <recompiled-binary> <recompiled-pdb> <decomp-dir>\n' % sys.argv[0])
-  print('\t-v, --verbose <offset>\t\t\tPrint assembly diff for specific function (original file\'s offset)')
-  print('\t-h, --html <output-file>\t\t\tGenerate searchable HTML summary of status and diffs')
-  sys.exit(1)
+parser = argparse.ArgumentParser(allow_abbrev=False, add_help=False,
```
madebr, line 11:
```suggestion
parser = argparse.ArgumentParser(allow_abbrev=False,
```

tools/reccomp/reccomp.py
```diff
@@ -1,66 +1,48 @@
...
+  description='Compare an original EXE with a recompiled EXE + PDB.')
+parser.add_argument('original', metavar='original-binary', help='The original binary')
+parser.add_argument('recompiled', metavar='recompiled-binary', help='The recompiled binary')
+parser.add_argument('pdb', metavar='recompiled-pdb', help='The PDB of the recompiled binary')
+parser.add_argument('decomp_dir', metavar='decomp-dir', help='The decompiled source tree')
+parser.add_argument('--verbose', '-v', metavar='offset', help='Print assembly diff for specific function (original file\'s offset)')
+parser.add_argument('--html', '-h', metavar='output-file', help='Generate searchable HTML summary of status and diffs')
+parser.add_argument('--help', action='help', help='Show this help message and exit')
```
madebr, line 19:
```suggestion
parser.add_argument('--html', '-H', metavar='output-file', help='Generate searchable HTML summary of status and diffs')
```

PR #31, opened 2023-06-21
lego1: MxString matching

disinvite:
Added some missing functions to MxString. The function at 0x100ae580 looks like a constructor for something else, so this may be all of it.

Everything matches except the swapped regs for `cmp eax, esi` in the MxString to char* comparator. What's odd is that other functions in the class can influence this order. For example, if I comment out one of the newly added ToUpperCase or ToLowerCase functions.

### DIFF
--- a/LEGO1/mxstring.cpp
+++ b/LEGO1/mxstring.cpp
@@ -11,6 +11,60 @@ MxString::MxString()
   this->m_length = 0;
 }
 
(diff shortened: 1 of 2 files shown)


madebr:
I think the function at `0x100ae580` is `MxString MxString::operator+(const char *)`.

itsmattkc:
Looks good, thanks!

PR #32, opened 2023-06-21
isle: match Isle::Tick and FindExistingInstance

foxtacles:
So while matching `Isle::Tick` I noticed that `MxDSObject::SetAtomId` is actually being inlined in there. That was a bit confusing at first until I finally noticed that the code of `Isle::Tick` itself actually exists multiple times: the first two calls in `WinMain` (`g_isle->Tick(1)`) are inlined in the original executable, and the last one (`g_isle->Tick(0)`) is not.

One of the inlined `Tick` calls uses `MxDSObject::SetAtomId` at `0x00401c40`, the other use it inlined. In the current decomp however, `Isle::Tick` is never inlined in `WinMain`. 

I can only assume all of this is a bunch of compiler optimization that we didn't run into yet, maybe we will at some point. Either way, I've now figured out that `m_unk24` in `MxDSObject` actually is signed and now it all matches.

`Isle::Tick` has a minor diff because two instructions are swapped (not functionally relevant). I think it's one of those cases - like the `CMP EAX,ESI` ones seen before - that cannot be fixed until surrounding unknowns line up in a certain way.

`FindExistingInstance` also matches now. `WINDOW_TITLE` actually needed to be a macro.


itsmattkc:
> That was a bit confusing at first until I finally noticed that the code of Isle::Tick itself actually exists multiple times:

Ahh yeah, I probably should have added a comment about this, it was one of the more complicated things to untangle when starting on ISLE. That's why there was no `SetAtomId()` call initially, because I (naturally) decompiled from the non-inlined function rather than the two inlined ones.

ISLE/define.h
```diff
@@ -8,7 +8,7 @@ class Isle;
 extern Isle *g_isle;
 extern int g_closed;
 #define WNDCLASS_NAME "Lego Island MainNoM App"
-extern const char *WINDOW_TITLE;
+#define WINDOW_TITLE "LEGO\xAE"
```
itsmattkc, line 11: Could you move the offset comment to here instead, just for the sake of documentation?

foxtacles: Yep, added them for both macros

itsmattkc:
Here's a question, what actually happens if you mark `Tick` as inline? I could imagine the devs actually doing that in an attempt to cut down on per-frame overhead, but it's curious that 1/3 times it doesn't get inlined. Presumably it's some kind of compiler quirk, so I'm curious what happens if we try doing that.

foxtacles:
> Here's a question, what actually happens if you mark `Tick` as inline? I could imagine the devs actually doing that in an attempt to cut down on per-frame overhead, but it's curious that 1/3 times it doesn't get inlined. Presumably it's some kind of compiler quirk, so I'm curious what happens if we try doing that.

Yeah, I wanted to try this tomorrow but I got curious, turned the PC back on and gave it a shot. Surprisingly `WinMain` matched right away to 100%, I didn't expect that!

foxtacles:
Cleaned up a bit so now we got the 100% on everything in `ISLE.EXE` except `SetupWindow`, `WndProc` and a little bit on `SetupVideoFlags`

itsmattkc:
Fantastic! Thanks for looking into all of this, will merge now

foxtacles:
Some findings I've noticed so far:

* The second call to `Tick` should be `Tick(0)` instead of `Tick(1)` it looks like
* `SetAtomId` is not inlined correctly everywhere, I believe that might be the main culprit why we have such a large diff. The second call to `Tick` (inlined) is expected to call `00401c40` (non-inlined `SetAtomId`) in both branches, however the decomp currently inlines it _in the first_ but not in the second branch

itsmattkc:
> `SetAtomId` is not correctly applied in one instance

I did notice that, but I think that may be just due to `WinMain` not yet matching, and that as it gets closer, the inlining behavior may start to line up. It'll be interesting to see progress of #38

PR #33, opened 2023-06-21
MxDirectDraw functions and structures

theR4K:
reversed all functions possible(at this point).

### DIFF
--- a/LEGO1/mxdirectdraw.cpp
+++ b/LEGO1/mxdirectdraw.cpp
@@ -1,0 +1,968 @@
+
+#include "mxdirectdraw.h"
+
+
+// OFFSET: LEGO1 0x1009DA20
+void EnableResizing(HWND hwnd, BOOL flag)
+{
+    static DWORD dwStyle;
+
+    if (!flag)
+    {
+        dwStyle = GetWindowLong(hwnd, GWL_STYLE);
+        if (dwStyle & WS_THICKFRAME)
+        {
+            SetWindowLong(hwnd, GWL_STYLE, GetWindowLong(hwnd, GWL_STYLE) ^ WS_THICKFRAME);
+        }
+    }
+    else
+    {
+        SetWindowLong(hwnd, GWL_STYLE, dwStyle);
+    }
+}
+
+// OFFSET: LEGO1 0x1009D490
+MxDirectDraw::MxDirectDraw()
+{
+    m_pFrontBuffer = NULL;
+    m_pBackBuffer = NULL;
+    m_pZBuffer = NULL;
+    m_pClipper = NULL;
+    m_pPalette = NULL;
+    m_pDirectDraw = NULL;
+    m_pText1Surface = NULL;
+    m_pText2Surface = NULL;
+    m_hWndMain = NULL;
+    m_bIgnoreWM_SIZE = FALSE;
+    m_bPrimaryPalettized = FALSE;
+    m_bOnlySystemMemory = FALSE;
+    m_bFullScreen = FALSE;
+    m_bOnlySoftRender = FALSE;
+    m_pauseCount = 0;
+    m_pErrorHandler = NULL;
+    m_pFatalErrorHandler = NULL;
+    m_pErrorHandlerArg = NULL;
+    m_pFatalErrorHandlerArg = NULL;
+    m_pCurrentDeviceModesList = NULL;
+    m_bIsOnPrimaryDevice = TRUE;
+    m_hFont = NULL;
+}
+
+// OFFSET: LEGO1 0x1009D530
+MxDirectDraw::~MxDirectDraw()
+{
+    Destroy();
+
+    if (m_pCurrentDeviceModesList != NULL)
+    {
+        delete m_pCurrentDeviceModesList;
+        m_pCurrentDeviceModesList = NULL;
+    }
+}
+
+// OFFSET: LEGO1 0x1009D5E0
+BOOL MxDirectDraw::Create(
+    HWND hWnd,
+    BOOL fullscreen_1,
+    BOOL surface_fullscreen,
+    BOOL onlySystemMemory,
+    int width,
+    int height,
+    int bpp,
+    const PALETTEENTRY* pPaletteEntries,
+    int paletteEntryCount)
+{
+    m_hWndMain = hWnd;
+
+    CacheOriginalPaletteEntries();
+
+    if (!RecreateDirectDraw(&m_pCurrentDeviceModesList->p_guid))
+    {
+        return FALSE;
+    }
+
+    m_bFlipSurfaces = surface_fullscreen;
+    BOOL fullscreen = 1;
+    m_bOnlySystemMemory = onlySystemMemory;
+    m_bIsOnPrimaryDevice = (m_pCurrentDeviceModesList->p_guid == 0);
+    if (m_bIsOnPrimaryDevice)
+    {
+        fullscreen = fullscreen_1;
+    }
+
+    if (!SetPaletteEntries(pPaletteEntries, paletteEntryCount, fullscreen))
+    {
+        return FALSE;
+    }
+
+    if (!DDInit(fullscreen))
+    {
+        return FALSE;
+    }
+
+    if (!DDSetMode(width, height, bpp))
+    {
+        return FALSE;
+    }
+
+    return TRUE;
+}
+
+// OFFSET: LEGO1 0x1009D800
+void MxDirectDraw::Destroy()
+{
+    DestroyButNotDirectDraw();
+
+    unk2();
+
+    if (m_pDirectDraw != NULL)
+    {
+        m_pDirectDraw->Release();
+        m_pDirectDraw = NULL;
+    }
+
+    m_bIsOnPrimaryDevice = TRUE;
+
+    if (m_pCurrentDeviceModesList != NULL)
+    {
+        delete m_pCurrentDeviceModesList;
+        m_pCurrentDeviceModesList = NULL;
+    }
+}
+
+// OFFSET: LEGO1 0x1009D860
+void MxDirectDraw::DestroyButNotDirectDraw()
+{
+    RestoreOriginalPaletteEntries();
+    if (m_bFullScreen)
+    {
+        if (m_pDirectDraw != NULL)
+        {
+            m_bIgnoreWM_SIZE = TRUE;
+            m_pDirectDraw->RestoreDisplayMode();
+            m_bIgnoreWM_SIZE = FALSE;
+        }
+    }
+
+    if (m_pPalette)
+    {
+        m_pPalette->Release();
+        m_pPalette = NULL;
+    }
+
+    if (m_pClipper)
+    {
+        m_pClipper->Release();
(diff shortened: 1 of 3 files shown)


theR4K:
yeh, I also have MxDirect3d which is inheritor of MxDirectDraw, but it's takes some time to clean up code...

LEGO1/mxdirectdraw.h
```diff
@@ -1,12 +1,142 @@
+
 #ifndef MXDIRECTDRAW_H
 #define MXDIRECTDRAW_H
 
+#include <ddraw.h>
+#include <Windows.h>
+
+BOOL g_is_PALETTEINDEXED8 = 0;
+BOOL g_isInsideError = FALSE;
```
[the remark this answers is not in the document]
theR4K: yes, it true. Need to fix that.

foxtacles:
Could you use 2-space indents? Thank you! https://github.com/isledecomp/isle/blob/master/CONTRIBUTING.md#code-style

LEGO1/mxdirectdraw.cpp
```diff
@@ -0,0 +1,968 @@
...
+    if (!g_isInsideError)
+    {
+        g_isInsideError = TRUE;
+        Destroy();
+        if (m_pErrorHandler)
+        {
+            m_pErrorHandler(message, error, m_pErrorHandlerArg);
+        }
+        g_isInsideError = FALSE;
+    }
```
madebr, line 610: I think this is a local static variable.
```suggestion
    static bool nsideError = FALSE;
    if (!insideError)
    {
        insideError = TRUE;
        Destroy();
        if (m_pErrorHandler)
        {
            m_pErrorHandler(message, error, m_pErrorHandlerArg);
        }
        insideError = FALSE;
    }
```

LEGO1/mxdirectdraw.cpp
```diff
@@ -0,0 +1,968 @@
...
+
+// OFFSET: LEGO1 0x1009E020
+void MxDirectDraw::unk1()
+{
+    HRESULT result;
+    byte* line;
+    DDSURFACEDESC ddsd;
+
+    // CRAZY, I dont know what kind of code can produce such of thing
+    //v8 = 2 - !m_bFlipSurfaces;
```
madebr, line 918:
```
v8 = m_bFlipSurfaces ? 2 : 0;
```

madebr: I see you found it out later.

About these comments, they don't add any value to the code.
Perhaps remove these.

theR4K:
image
build fixed, results at this point

LEGO1/mxdirectdraw.cpp
```diff
@@ -189,10 +198,11 @@ void MxDirectDraw::DestroyButNotDirectDraw()
   }
 }
 
-// OFFSET: LEGO1 0x1009E880
+// OFFSET: LEGO1 0x1009E880 STUB
 const char* MxDirectDraw::ErrorToString(HRESULT error)
 {
   //TODO
+  return NULL;
```
tamlin-mike, line 205: PR #19 seems to contain a (complete?) implementation of this method. Just wanted to make it explicit to prevent double work.

theR4K: yep, I know. But also in PR #19 this method must be modified to remove any \0 characters at the end of the strings, because it redundant.

LEGO1/mxdirectdraw.h
```diff
@@ -5,138 +5,137 @@
...
-BOOL g_is_PALETTEINDEXED8 = 0;
-BOOL g_isInsideError = FALSE;
+extern BOOL g_is_PALETTEINDEXED8;
 
 //size 0x880
 class MxDirectDraw
 {
 public:
-    typedef void (*ErrorHandler)(const char*, HRESULT, void*);
+  typedef void (*ErrorHandler)(const char*, HRESULT, void*);
```
tamlin-mike, line 14: Might want to be explicit about calling convention here:
`typedef void (__cdecl *ErrorHandler)(const char*, HRESULT, void*);`

LEGO1/mxdirectdraw.cpp
disinvite: Hey, I've incorporated this (and other changes) in my PR on your fork. If you accept, I will cancel #19 here and we can merge this one.

theR4K: accepted, now I will recalculate the accuracy

LEGO1/mxdirectdraw.h
theR4K: I don't think it's necessary. As long as it not a class functions it must have C calling convention.(unless compiler flags says other)

theR4K:
image

MxDirectDraw::~MxDirectDraw, MxDirectDraw::Destroy - has inlined DeviceModesInfo destructor.

other functions have some changes in compare operators

theR4K:
image

it's done

theR4K:
destructor and destroy have inlined  DeviceModesInfo destructor;
EnableResizing, SetPaletteEntries, ErrorToString have shifted edi\esi registers (they are used in opposite way)
IsSupportedMode have switched campare.
So technically this do the same.....

LEGO1/mxdirectdraw.h
tamlin-mike: > (unless compiler flags says other)

You correctly identified the case I tried to protect against; `/Gr` or `/Gz`. It would make for... interesting debugging. How loud do you think you would curse if you were in such a project, and finally realized why the stack was smashed? :-)

Anyway, no biggie. Those flags are few and far between. I just wanted to flag for potential blood pressure increases down the line.

LEGO1/mxdirectdraw.cpp
```diff
@@ -1,22 +1,1207 @@
+
 #include "mxdirectdraw.h"
 
-// OFFSET: LEGO1 0x1009e7f0
-int MxDirectDraw::FlipToGDISurface()
+//TODO: make commonn place for defines like that
+#ifndef _countof
+#define _countof(arr) sizeof(arr) / sizeof(arr[0])
```
itsmattkc, line 6: There's `legoutil.h` and `decomp.h` which would be suitable for macros like this

tamlin-mike (changes requested):

LEGO1/mxdirectdraw.cpp
```diff
@@ -32,6 +29,20 @@ void EnableResizing(HWND hwnd, BOOL flag)
...
+  if (p_guid != NULL)
+  {
+    free(p_guid);
+  }
+
+  if (m_mode_ARRAY != NULL)
+  {
+    free(m_mode_ARRAY);
+  }
+}
```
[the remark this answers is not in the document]
theR4K: yep, it make sense. I just don't see differences between "new" and "malloc" because they technically do the same*.
But yes, you are right.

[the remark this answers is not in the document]
madebr: `malloc` in c land only does allocation.
`new` in c++ land does allocation + initialization by calling the constructor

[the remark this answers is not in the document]
theR4K: > `malloc` in c land only does allocation. `new` in c++ land does allocation + initialization by calling the constructor

yep, that what * stands for...

LEGO1/mxdirectdraw.cpp
```diff
@@ -1,22 +1,1223 @@
...
+{
+  if (p_guid != NULL)
+  {
+    free(p_guid);
+  }
+
+  if (m_mode_ARRAY != NULL)
+  {
+    free(m_mode_ARRAY);
+  }
```
itsmattkc, line 43: I agree these are probably more likely `delete`, which will end up inlining the `NULL` check. 
```suggestion
  delete p_guid;
  delete m_mode_ARRAY;
```

LEGO1/mxdirectdraw.cpp
```diff
@@ -1,22 +1,1223 @@
...
+BOOL MxDirectDraw::TextToTextSurface2(const char* text)
+{
+  return TextToTextSurface(
+    text,
+    m_pText2Surface,
+    m_text2SizeOnSurface);
+}
+
+// OFFSET: LEGO1 0x1009E020
+void MxDirectDraw::unk1()
```
itsmattkc, line 1167: Functions should be named after the address for less ambiguity. e.g.

```suggestion
void MxDirectDraw::FUN_1009E020()
```

LEGO1/mxdirectdraw.cpp
```diff
@@ -1,22 +1,1223 @@
...
+
+    if (m_bFlipSurfaces)
+    {
+      m_pFrontBuffer->Flip(NULL, DDFLIP_WAIT);
+    }
+  }
+}
+
+// OFFSET: LEGO1 0x1009D920
+void MxDirectDraw::unk2()
```
itsmattkc, line 1212:
```suggestion
void MxDirectDraw::FUN_1009D920()
```

LEGO1/mxdirectdraw.h
```diff
@@ -1,12 +1,130 @@
...
+    int paletteEntryCount,
+    BOOL fullscreen);
+  BOOL TextToTextSurface(
+    const char* text,
+    IDirectDrawSurface* pSurface,
+    SIZE& textSizeOnSurface);
+  BOOL TextToTextSurface1(const char* text);
+  BOOL TextToTextSurface2(const char* lpString);
+  void unk1();
+  void unk2();
```
itsmattkc, line 127:
```suggestion
  void FUN_1009E020();
  void FUN_1009D920();
```

LEGO1/mxdirectdraw.cpp
```diff
@@ -1,22 +1,1223 @@
+
 #include "mxdirectdraw.h"
+#include "decomp.h"
 
-// OFFSET: LEGO1 0x1009e7f0
-int MxDirectDraw::FlipToGDISurface()
+
+#ifndef DDSCAPS_3DDEVICE
+#define DDSCAPS_3DDEVICE 0x00002000l
```
itsmattkc, line 7: At what times is this not defined? Shouldn't this always be in `ddraw.h`? If not, a comment should probably be added explaining it

LEGO1/mxdirectdraw.cpp
```diff
@@ -1,22 +1,1223 @@
...
+  m_currentMode.bitsPerPixel = bpp;
+
+  if (!DDCreateSurfaces())
+  {
+    return FALSE;
+  }
+
+  DDSURFACEDESC ddsd;
+
+  unk1();
```
itsmattkc, line 811:
```suggestion
  FUN_1009E020();
```

tamlin-mike: Perhaps a better name could be `MxDirectDraw::method_1009E020_ClearBackBuffersSlow`?

It's clearly a method (not a free-standing function), which is obvious by the caller initializing ecx and the callee using ecx.

If you look at the method it's not any fun at all - quite the opposite, e.g. `memset`-ing the backbuffer surface _line-by-line_.

PR #34, opened 2023-06-21
Initial implementation of MxBitmap

ItzSwirlz:
Sorry about the commit spam guys, I'll probably make a force push or stash the commits once I'm done. I'm also doing this at school (last few days, so nothing is happening), on a M1 Mac so I currently can't really build it.

A lot of what's missing is stuff relating to MxPalette, so that needs to preferably be knocked out first (at least two of the functions)

### DIFF
--- a/LEGO1/mxbitmap.cpp
+++ b/LEGO1/mxbitmap.cpp
@@ -3,25 +3,72 @@
 // OFFSET: LEGO1 0x100bc980
 MxBitmap::MxBitmap()
 {
-  // TODO
+  this->m_info = NULL;
+  this->m_bmiHeader = NULL;
+  this->m_paletteData = NULL;
+  this->m_data = NULL;
+  this->m_unk18 = FALSE;
+  this->m_palette = NULL;
 }
 
 // OFFSET: LEGO1 0x100bca10
 MxBitmap::~MxBitmap()
 {
-  // TODO
+  if (this->m_info != NULL) {
+    delete m_info;
+  }
+  if (this->m_data != NULL) {
+    delete m_data;
+  }
+  if (this->m_palette != NULL) {
+    delete m_palette;  
+  }
 }
 
 // OFFSET: LEGO1 0x100bd1c0
 MxPalette *MxBitmap::CreatePalette()
 {
-  // TODO
-  return NULL;
+  // FIXME: This function needs MxPalette to be completed. Also INFERRING usage of MxBool
+  MxPalette *pal = NULL;
+  MxPalette *ppal;
+  MxBool success = FALSE;
+
+  if(this->m_unk18 == FALSE) {
+    // ppal = MxPalette::FromBitmapPalette(this->m_paletteData);
+  } else {
+    if(this->m_unk18 != TRUE) {
+      if(!success && pal != NULL) {
+        delete pal;
+        pal = NULL;
+      }
+    }
+    //.pal = MxPalette::Clone(this->m_palette);
+  }
+  if(pal != NULL) {
+    success = TRUE;
+  }
+
+  return pal;
 }
 
 // OFFSET: LEGO1 0x100bcd10
-long MxBitmap::Read(const char *)
+long MxBitmap::Read(const char *filename)
 {
-  // TODO
-  return 0;
+  HANDLE handle;
+  int unk1;
+  MxResult ret = FAILURE;
+
+  handle = CreateFileA(filename,GENERIC_READ,FILE_SHARE_READ,(LPSECURITY_ATTRIBUTES)NULL,OPEN_EXISTING,FILE_ATTRIBUTE_NORMAL,(HANDLE)NULL);
+  if(handle != (HANDLE)INVALID_HANDLE_VALUE) {  // INVALID_HANDLE_VALUE = -1, or 0xffffffff
+		// FIXME: line 16. iVar gets changed in this line
+    if(unk1 == 0) {
+      ret = SUCCESS;
+    }
+  }
+  if(handle != (HANDLE)NULL) {
+    CloseHandle(handle);
+  }
+
+  return ret;
 }
+
--- a/LEGO1/mxbitmap.h
+++ b/LEGO1/mxbitmap.h
@@ -1,15 +1,26 @@
 #ifndef MXBITMAP_H
 #define MXBITMAP_H
 
-#include "mxpalette.h"
+#include <stdlib.h>
 
-class MxBitmap
+#include "mxcore.h"
+#include "mxpalette.h"
+#include "mxtypes.h"
+
+class MxBitmap : public MxCore
 {
 public:
   __declspec(dllexport) MxBitmap();
   __declspec(dllexport) virtual ~MxBitmap();
   __declspec(dllexport) virtual MxPalette *CreatePalette();
   __declspec(dllexport) virtual long Read(const char *);
+private:
+  BITMAPINFO *m_info;
+  BITMAPINFOHEADER *m_bmiHeader;
+  RGBQUAD *m_paletteData;
+  LPVOID *m_data;
+  MxBool m_unk18;
+  MxPalette *m_palette;
 };
 
 #endif // MXBITMAP_H


LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,13 @@
+#include "mxbitmap.h"
+
+// OFFSET: LEGO1 0x100bc980
+MxBitmap::MxBitmap()
+{
+    // fixme: vtable, mxcore??
```
itsmattkc, line 6: This class derives from MxCore, that's why you're seeing that

ItzSwirlz: So what does that mean, like, is there an "extends" of C++?

tamlin-mike: It means class declaration is (to be) `MxBitmap : public MxCore` (`public` could also be `protected` or `private`).

That results in
1. The subclass inheriting all fields of the baseclass.
2. The compiler generating code to invoke MxCore's ctor/dtor from the subclass's ctor/dtor, before/after setting the vfptr to the table for the actual type.

Note that the baseclass' ctor/dtor may or may not be inlined, to the point even setting the vfptr could be elided. In the case of MxCore's ctor and dtor, though, they are never inlined in the dll, so MxBitmap's ctor 1. calls MxCore's ctor, 2. sets the vfptr to MxBitmap vtable. See follow up comment for the header file below.

LEGO1/mxbitmap.h
```diff
@@ -1,13 +1,24 @@
...
+
 class MxBitmap
 {
 public:
   __declspec(dllexport) MxBitmap();
   __declspec(dllexport) virtual ~MxBitmap();
   __declspec(dllexport) virtual MxPalette *CreatePalette();
   __declspec(dllexport) virtual long Read(const char *);
+private:
+  virtual void vtable();
```
tamlin-mike, line 15: This is not how it works. Presence of a virtual function, in this class or any of its baseclasses, enforces the generation of a hidden vfptr at object offset 0 (disregarding multiple- and virtual inheritance). That pointer is generated in the top-most class having a virtual function. Since inheriting a type includes all its data, by inheriting from MxCore the vfptr becomes automatically included in MxBitmap.

LEGO1/mxbitmap.h
```diff
@@ -1,13 +1,24 @@
...
+  BITMAPINFO m_info;
+  BITMAPINFOHEADER m_bmiHeader;
+  RGBQUAD m_paletteData;
+  LPVOID m_data;
+  char m_unk18;
+  MxPalette m_palette;
```
tamlin-mike, line 21: Field 0x1c is a pointer to the MxPalette.

ItzSwirlz:
Hi everyone, I want to move on with other parts of the decomp and I don't want to freeze myself here. I'd like to have this merged and then come back to it later

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,105 @@
...
+    MxBool success;
+
+    pal = NULL;
+    success = MX_FALSE;
+    if(this->m_unk18 == '\0') {
+        ppal = (MxPalette *)malloc(0x414);
+        pal = NULL;
+        if(ppal != NULL) {
+            // pal = MxPalette::FromBitmapPalette(ppal, this->m_paletteData);
+        }
```
madebr, line 46: This is probably using an alternative c++ constructor, or calling `MxPalette::FromBItmapPalette` directly.
It doesn't make sense in c++ to allocate, and then call a method on raw data.
```suggestion
        // ppal = MxPalette::FromBitmapPalette(this->m_paletteData);
```

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+        if(this->m_unk18 != '\x01') {
+            goto LAB_100bd253;
+        }
```
madebr, line 50: formatting + change the code in a way to get rid of the `goto`'s.
The goto's were added by the Ghidra decompiler, but weren't there in the OG code.

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+// OFFSET: LEGO1 0x100bcd10
+long MxBitmap::Read(char const *filename)
+{
+    HANDLE hObject;
+    int iVar1;
+    long lVar2;
+
+    lVar2 = -1;
+    hObject = CreateFileA(filename,GENERIC_READ,FILE_SHARE_READ,(LPSECURITY_ATTRIBUTES)NULL,OPEN_EXISTING,FILE_ATTRIBUTE_NORMAL,(HANDLE)NULL);
```
madebr, line 75: Formatting + use spaces + name your variables appropriately.
This is c++, no c90 where variables have to be declared at the start of a method.

LEGO1/mxbitmap.h
```diff
@@ -1,13 +1,30 @@
...
+#include "mxpalette.h"
+
+class MxBitmap : public MxCore
 {
 public:
   __declspec(dllexport) MxBitmap();
   __declspec(dllexport) virtual ~MxBitmap();
   __declspec(dllexport) virtual MxPalette *CreatePalette();
-  __declspec(dllexport) virtual long Read(const char *);
+  __declspec(dllexport) virtual long Read(char const *);
```
madebr, line 16: Order matters, `Read` comes before `CreatePalette`. And perhaps also add `FIXME`'s for the missing virtual methods.

LEGO1/mxbitmap.h
```diff
@@ -1,13 +1,30 @@
...
+
+  const char *GetClassName() const;
+  MxBool IsClass(const char *name) const;
+  long Tickle();
```
madebr, line 20: These are not present in the OG dll.
They just forward to the parent `MxCore`.

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,105 @@
+#include "mxbitmap.h"
+#include "mxbool.h"
+#include "mxpalette.h"
+
+// OFFSET: LEGO1 0x100bc980
+MxBitmap::MxBitmap()
+{
+    this->m_info = NULL;
```
madebr, line 8: indentation

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,105 @@
...
+    if (this->m_info != NULL) {
+        free(this->m_info);
+    }
+    if (this->m_data != NULL) {
+        free(this->m_data);
+    }
+    if (this->m_palette != NULL) {
+        // FIXME: Once m_palette/base is understood
+    }
+    // FIXME: Unwind@100bca91. Stack unwinding, I've never done this before.
```
madebr, line 28: This is code generated by the c++ compiler.
The compiler adds code here to first destruct all members with destructors of `MxBitmap`.
Then, it will call the destructor of the parent (=`MxCore`).
```suggestion
```

Hint: in ghidra, you can inline function by enabling the `In Line` function attribute in the function signature editor.

LEGO1/mxbitmap.cpp
tamlin-mike: @madebr is correct, 0x100BEED0 is MxPalette ctor taking a const RGBQUAD* (or const PALETTEENTRY*).

Additionally, what's spelled `malloc` here isn't malloc, it's the global operator new function. I.e. what's here spelled out as `malloc(0x414);` can in step 1 be changed to `malloc(sizeof MxPalette);` (providing a chance to verify with a compile-time assert that the declaration of MxBitmap is indeed 0x414), and for its final step it becomes
`pal = new MxPalette(m_paletteData);`. The null-check is something the Microsoft compilers insert, since their allocators by default do not throw on out-of-memory, but simply return 0.

LEGO1/mxbitmap.cpp
tamlin-mike: The code simply does `delete m_palette;`, which has been compiled into a virtual call to the compiler generated virtual function (which replaces the virtual destructor in the vtable) `MxPalette::'scalar deleting destructor'(unsigned int)`. where the argument `1` tells the generated method to delete the memory. The rest is just generated by the compiler.

LEGO1/mxbitmap.h
ItzSwirlz: so do these three functions need to be in the header?

LEGO1/mxbitmap.cpp
ItzSwirlz: I see the goto call; is it a call to another function?

LEGO1/mxbitmap.cpp
ItzSwirlz: I think lVar2 is just a MxResult, thinking about it.

LEGO1/mxbitmap.h
ItzSwirlz: The errors showed up because I put in an extra instruction in isle.mak

ItzSwirlz:
All done, my oh my that took WAY too long to fix that damn merge conflict. You guys are on your own with the project file

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,78 @@
...
+// OFFSET: LEGO1 0x100bd1c0
+MxPalette *MxBitmap::CreatePalette()
+{
+  // FIXME: This function needs MxPalette to be completed. Also INFERRING usage of MxBool
+  MxPalette *m_pal = NULL;
+  MxPalette *m_ppal;
+  MxBool m_success = MX_FALSE;
+
+  if(this->m_unk18 == '\0') {
+    // ppal = MxPalette::FromBitmapPalette(this->m_paletteData);
```
madebr, line 41:
```suggestion
  if(!this->m_unk18) {
    // ppal = new MxPalette(this->m_paletteData);
```

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,78 @@
...
+  } else {
+    if(this->m_unk18 != '\x01') {
```
madebr, line 43: The unknown member is used as a 8 bit integer, not a character.
```suggestion
    if (this->m_unk18 != 1) {
```

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,78 @@
...
+      if((!m_success) && (m_pal != NULL)) {
```
madebr, line 44: Use less spaces
```suggestion
      if(!m_success && m_pal != NULL) {
```

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,78 @@
...
+     // FIXME: Once m_palette/base is understood
+  }
+  delete m_palette;
+}
+
+// OFFSET: LEGO1 0x100bd1c0
+MxPalette *MxBitmap::CreatePalette()
+{
+  // FIXME: This function needs MxPalette to be completed. Also INFERRING usage of MxBool
+  MxPalette *m_pal = NULL;
```
madebr, line 36: Don't prefix local variables with `m_`.
```suggestion
  MxPalette *pal = NULL;
```

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,78 @@
...
+  MxPalette *m_ppal;
+  MxBool m_success = MX_FALSE;
+
+  if(this->m_unk18 == '\0') {
+    // ppal = MxPalette::FromBitmapPalette(this->m_paletteData);
+  } else {
+    if(this->m_unk18 != '\x01') {
+      if((!m_success) && (m_pal != NULL)) {
+        // (**(pal->base).vtable)(1);
+        m_pal = NULL;
```
madebr, line 46:
```suggestion
        delete m_pal;
        m_pal = NULL
```

LEGO1/mxbitmap.cpp
ItzSwirlz: i thought that was the general standard? see CONTRIBUTING.md unless things changed

tamlin-mike: The `m_` prefix/"whart" applies to class/struct **member data** (`m`, `member`). It does not apply to local variables.

LEGO1/mxbitmap.cpp
tamlin-mike: From my current found uses of it, it's likely a `bool`. Provided that's correct, the original code is probably just overly verbose comparing to literal `true`. Just a comment for future reference.

itsmattkc: Yeah if you see a `char` being set to `\0` and `\1`, you can be reasonably sure it's a `bool` (or `MxBool` here since MSVC had no `bool` yet).

LEGO1/mxvideoparamflags.h
```diff
@@ -40,6 +40,23 @@ class MxVideoParamFlags
...
+  inline BYTE GetWideViewAngle() { return m_flags1.bit6; }
+  inline BYTE Get_f1bit7()       { return m_flags1.bit7; }
+  inline BYTE Get_f2bit0()       { return m_flags2.bit0; }
+  inline BYTE Get_f2bit1()       { return m_flags2.bit1; }
+  inline BYTE Get_f2bit2()       { return m_flags2.bit2; }
+  inline BYTE Get_f2bit3()       { return m_flags2.bit3; }
+  inline BYTE Get_f2bit4()       { return m_flags2.bit4; }
+  inline BYTE Get_f2bit5()       { return m_flags2.bit5; }
+  inline BYTE Get_f2bit6()       { return m_flags2.bit6; }
+  inline BYTE Get_f2bit7()       { return m_flags2.bit7; }
```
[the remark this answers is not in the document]
ItzSwirlz: I didn't commit this, I just did some other git stuff to make my branch merge nicely

LEGO1/mxbitmap.cpp
ItzSwirlz: the call says ::FromBitmapPalette, I can check later

ItzSwirlz: Oh. It's another call further down that says "if(ppal != NULL)" THEN palette from bitmap palette. You're right, when I get home and have access to a x86 machine again I'll fix this.

LEGO1/mxbitmap.cpp
```diff
@@ -10,23 +10,22 @@ MxBitmap::MxBitmap()
...
+  this->m_unk18 = MX_FALSE;
   this->m_palette = NULL;
 }
 
 // OFFSET: LEGO1 0x100bca10
 MxBitmap::~MxBitmap()
 {
   if (this->m_info != NULL) {
-    free(this->m_info);
+    delete m_info;
```
tamlin-mike, line 21: It's probably better to be consistent in the use of explicit `this->`, i.e. having `if (this->m_info)` followed by `delete m_info;` is inconsistent. (applies to the whole file).

tamlin-mike (commented):
A note on the commit message: The vtable is the _table_ holding the function pointers to the vfuncs. `vtable(1)` makes no sense (unless referring to a newly invented kernel function :-) ). I believe the intention was something like "a virtual function call ptr->vfunc(1); is often a call to the scalar deleting destructor". I just added this point to prevent confusion.

Besides that, lgtm.

LEGO1/mxbitmap.cpp
ItzSwirlz: I'm not sure, does it cause a difference in the matching? And most likely delete is what is actually in the code

tamlin-mike: `this->m_foo` is identical to plain `m_foo` for both parsing and code-generation purposes.

To clarify, my comment was referring to (simplified for brevity):
inconsistent1: `if (this->m_foo) delete m_foo;`
inconsistent2: `if (m_foo) delete this->m_foo;`
consistent1: `if (this->m_foo) delete this->m_foo;`
consistent2: `if (m_foo) delete m_foo;`

The only places `this->` is required is _some_ template code, which is way deeper down a rabbit hole than we shall go here. :) Sometimes it can be of documentation benefit to keep `this->` prefix in rev. code, but whatever is selected it should be applied consistently.

I can also straighten out any question marks regarding use of `delete` in the actual code. It's not "most likely", it's definitely. 100%. 0x10086260 is `??3@YAXPAX@Z`, a.k.a. the global operator delete function.

ItzSwirlz: iirc for the MxTimer matching, I think sometimes the MxTimer::member variables made a difference, but idk, I want to move onto MxPalette once this is merged

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,77 @@
...
+  if(this->m_unk18 == MX_FALSE) {
+    // ppal = MxPalette::FromBitmapPalette(this->m_paletteData);
+  } else {
+    if(this->m_unk18 != MX_TRUE) {
+      if(!success && pal != NULL) {
+        delete pal;
+        pal = NULL;
+      }
+    }
+    //.pal = MxPalette::Clone(this->m_palette);
```
madebr, line 48:
```suggestion
    // pal = this->m_palette.Clone(); // FIXME
```

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,77 @@
...
+// OFFSET: LEGO1 0x100bd1c0
+MxPalette *MxBitmap::CreatePalette()
+{
+  // FIXME: This function needs MxPalette to be completed. Also INFERRING usage of MxBool
+  MxPalette *pal = NULL;
+  MxPalette *ppal;
+  MxBool success = MX_FALSE;
+
+  if(this->m_unk18 == MX_FALSE) {
+    // ppal = MxPalette::FromBitmapPalette(this->m_paletteData);
```
madebr, line 40: This is really a constructor with `new` because `operator_new` is called in this method.
```suggestion
    // ppal = new MxPalett(this->m_paletteData); // FIXME
```

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+// OFFSET: LEGO1 0x100bd1c0
+MxPalette *MxBitmap::CreatePalette()
+{
+  // FIXME: This function needs MxPalette to be completed. Also INFERRING usage of MxBool
+  MxPalette *pal = NULL;
+  MxPalette *ppal;
+  MxBool success = MX_FALSE;
+
+  if(this->m_unk18 == MX_FALSE) {
```
madebr, line 39: I think `m_unk18` is not a boolean.
The logic in this function is.
I hints `m_unk18` can be something else then 0 and 1.
```c++
{
  pal = NULL;
  success = false;
  if (m_unk18 == 0) {
    pal = new MxPalette(m_paletteData);
  } else if (m_unk == 1) {
    pal = m_palette.Clone();
  }
  if (pal) {
    success = true;
  }
  if (!success && pal) {
    delete pal;
    pal = NULL
  }
}
```

LEGO1/mxbitmap.cpp
```diff
@@ -0,0 +1,77 @@
...
+  if(handle != (HANDLE)INVALID_HANDLE_VALUE) {  // INVALID_HANDLE_VALUE = -1, or 0xffffffff
+		// FIXME: line 16. iVar gets changed in this line
+    if(unk1 == 0) {
+      ret = SUCCESS;
+    }
+  }
+  if(handle != (HANDLE)NULL) {
+    CloseHandle(handle);
+  }
+
```
madebr, line 74:
```suggestion
  handle = CreateFileA(filename, GENERIC_READ, FILE_SHARE_READ, (LPSECURITY_ATTRIBUTES)NULL, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, (HANDLE)NULL);
  if (handle != (HANDLE)INVALID_HANDLE_VALUE && !this->VTable0x20()) {
    ret = SUCCESS;
  }
  if (handle != (HANDLE)NULL) {
    CloseHandle(handle);
  }

```

tamlin-mike: The `if (handle != (HANDLE)NULL)` might deserve a comment to the effect of `// BUG compatibility with orig binary`.

madebr: Sure, why is it a big?

tamlin-mike: Zero is, however unbelievable (and in practice hopefully impossible to encounter), a valid HANDLE value. The only invalid is INVALID_HANDLE_VALUE.

Some nitpicking:
The comparison against `INVALID_HANDLE_VALUE` does not need a cast. The definition of `INVALID_HANDLE_VALUE` already includes the cast to `HANDLE`.

It's not very often one needs to be explicit about A/W version of API functions: `CreateFileA` -> `CreateFile`.

ItzSwirlz: The code right here is just temporary for now, we can come back to this later, this is an **Initial** implementation

itsmattkc:
This looks like it's getting pretty close, but you haven't yet added `mxbitmap.cpp` to the CMakeLists. Looks like there are a few things that haven't yet been updated for the latest master (`legoinc` is no longer used because we renamed `GetClassName`, etc.)

ItzSwirlz:
Ugh! I tried to merge it, and.. ugh. Let me repush

ItzSwirlz:
aight

itsmattkc:
Seems good. Obviously 2/4 functions aren't matching yet, but it's a start. Merging now.

PR #37, opened 2023-06-22
Hotfix: messed up structure due to windows.h

foxtacles:
This PR: https://github.com/isledecomp/isle/pull/32 accidentally broke `ISLE.EXE` - not so much because of the code changes, but because of a `Windows.h` include that sneaked `GetClassName` in and lead to immediate crashes.

I removed all remaining instances of `windows.h` includes and replaced them with `legoinc.h`

### DIFF
--- a/ISLE/define.h
+++ b/ISLE/define.h
@@ -1,7 +1,7 @@
 #ifndef DEFINE_H
 #define DEFINE_H
 
-#include <Windows.h>
+#include "legoinc.h"
 
 class Isle;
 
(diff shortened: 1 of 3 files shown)


tamlin-mike:
I'd consider (read: strongly recommend) renaming that method to not clash with Windows global names. windows.h will be included by other source files that will need e.g. USER or GDI, that in the same TU will need access to Mx classes.

Besides, what's returned isn't a class name as such. For an example, see:
"RaceStandsEntity", and the _three_ vfunc3's 0x1000EFA0, 0x1000F0C0, 0x1000F1E0.

It would be... difficult to link a program if there were three different instance if classes named `RaceStandsEntity`. :-)

PR #38, opened 2023-06-22
isle: Improving WinMain and SetupWindow accuracy

foxtacles:
* `WinMain`: Second `Tick` call should be `Tick(0)`
* `WinMain`: Inner loop is `while` instead of `do/while`
* `WinMain`: `SetupWindow` has an extra parameter (unused)
* `SetupWindow`: declare it inline (it won't be inlined; this is just to get `WinMain` correct)
* `SetupWindow`: reorder `CreateWindowExA` logic
* `SetupWindow`: reorder cursor assignments
* `SetupWindow`: `switch/case` instead of `if/else`
* `SetupWindow`: `m_useJoystick` is a bool/1-char
* `Isle::Tick`: reorganized conditionals to match `WinMain`

`Isle::SetupWindow` is probably done even though it's just at ~98%, the reason is again a psuedorandom compiler decision on which register to use for the function calls `ShowWindow` and `UpdateWindow`. I don't think this can be directly further influenced. 

Similar issue with `WinMain`, it's at ~99%.


foxtacles:
In reference to: https://github.com/isledecomp/isle/pull/32#issuecomment-1602956363

What I've noticed is that if you add more code to the innermost loop of `WinMain` - let's say copy-paste the `g_mousedown` stuff, or anything really - it will start showing correct inlining (in this case, _not_ inlining `SetAtomId` in the second `Tick` call). I'm assuming by crossing a certain threshold in terms of `WinMain` code size, the compiler will eventually produce what we are looking for.

However, in terms of functionality `WinMain` is complete now afaik so we can only work with what we have.

foxtacles:
Last 0,5% of `WinMain` is another case of register/conditional swap, so I consider it done

itsmattkc:
> I was also trying to rewrite the g_mousedown block into something that is not using goto, but have not been successful so far. Maybe in this case the devs actually implemented it this way.

Yeah I was never happy with this `goto` either, but it's hard to see how else it could have been implemented. It definitely is _possible_ that's really what they did, even though it would be a tad hacky if so.

PR #40, opened 2023-06-22
Use bitfield for MxVideoParamFlags

disinvite:
Using a bitfield for MxVideoParamFlags results in the same xor/and logic that was partially inlined in the header file. This approach is a lot cleaner and there's a good chance this is what the devs would have landed on.

The code generation is really finicky -- other inlines in the header influence the code just by being there -- so I decided to stub out all of them. This got the match to 100%.

While I was in isle.cpp::SetupVideoFlags, I changed the signature so that the `m_using16bit` parameter is just `using16bit`. Two of the parameters from SetupVideoFlags are negated when we set the flags, which seems odd. Maybe a name change is in order on backBuffers?

### DIFF
--- a/ISLE/isle.cpp
+++ b/ISLE/isle.cpp
@@ -36,7 +36,7 @@ Isle::Isle()
   m_windowActive = 1;
 
   m_videoParam = MxVideoParam(MxRect32(0, 0, 639, 479), NULL, 1, MxVideoParamFlags());
-  m_videoParam.flags().Enable16Bit(MxDirectDraw::GetPrimaryBitDepth() == 16);
+  m_videoParam.flags().Set16Bit(MxDirectDraw::GetPrimaryBitDepth() == 16);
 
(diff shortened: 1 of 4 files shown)


foxtacles:
I can't check right now, but `Isle::Isle` is no longer 100%. Maybe a regression?

disinvite:
Yes, thank you. Fixed it. `Isle::Isle` should be the only other place where we set the flags at present. The previous function `Enable16Bit` used `unsigned char` as its argument, so I changed mine from `BOOL` to `BYTE` to match. (Maybe the devs weren't consistent about the types either?)

itsmattkc:
> (Maybe the devs weren't consistent about the types either?)

Yeah I fell into this trap too, I knew that `Enable16Bit` was an `unsigned char` (most likely actually `MxBool`), so I assumed they would have done the same for all of them, but it turned out not to be the case (#26).

PR #41, opened 2023-06-22
Convert all sources eol's to lf

madebr:
The `.gitattributes` file should inform git to always use `\n` as end-of-line.
I've copied  the items from https://github.com/alexkaratarakis/gitattributes

This pr is the result of me running the following commands in the isle directory
```sh
find . -iname "*.cpp" -exec fromdos {} \;
find . -iname "*.h" -exec fromdos {} \;
find . -iname "*.py" -exec fromdos {} \;
find . -iname "*.html" -exec fromdos {} \;
find . -iname "*.MD" -exec fromdos {} \;
```


itsmattkc:
Might have to do this again for the most recent merge

madebr:
Don't be alarmed by `isle.mak` being converted to `nl`.
What `.gitattributes` does, is tell git how to normalize files before it hashes them.

Instead of the `fromdos` commands in my original message, you can also run:
```
git add --renormalize .
```

PR #42, opened 2023-06-23
lego: implement SetControlMax and related partials (MxVideoManager, MxPalette, MxUnknown100dc6b0)

foxtacles:
This PR implements the structure for `MxVideoManager`, which is the base class of `LegoVideoManager`. `LegoNavController::SetControlMax` is fully implemented now.

`MxVideoManager` has a base class that I have no name for yet and there is none in the source, so I've called it `MxUnknown100dc6b0` (address of its vtable). It is relatively small and most notably has a `MxCriticalSection` member, most of its functions work on that with `MxAutoLocker`. We can figure out the meaning of this later.


LEGO1/mxpalette.cpp
```diff
@@ -0,0 +1,8 @@
+#include "mxpalette.h"
+
+// OFFSET: LEGO1 0x100bf150
+int MxPalette::GetEntries(LPPALETTEENTRY p_entries)
+{
+  memcpy(p_entries, this->m_entries, sizeof(this->m_entries));
```
ItzSwirlz, line 6: how did you pull this out of the psuedocode? you're probably smarter than i am so i want to know

LEGO1/mxpalette.h
```diff
@@ -1,11 +1,22 @@
...
+
+#include "mxcore.h"
+
+class MxPalette : public MxCore
 {
 public:
   __declspec(dllexport) unsigned char operator==(MxPalette &);
   __declspec(dllexport) void Detach();
+
+  int GetEntries(LPPALETTEENTRY p_entries);
```
ItzSwirlz, line 14: could be a MxResult maybe

foxtacles: That is most likely true. Thank you!

LEGO1/mxpalette.cpp
foxtacles: Decompilation mainly rests on knowledge of C/C++ and how compilers map it onto the target architecture, x86 in this case. There are many details to this, and this is one of them:

Most production software is compiled with optimization flags enabled, generally to optimize for speed. This is the case with LEGO Island too (`-O2`). One of the most common optimizations is to inline functions to reduce overhead imposed by a function call (setup/teardown of stack frames etc.). This is especially worthwhile for calls to `memcpy`, `strcpy`, etc. when the size is known at compile time, like in the given example: `sizeof(this->m_entries)` is `1024`

So the compiler can decide to replace a call to `memcpy` with those 4 simple instructions:

```        
LEA               ESI ,[ECX+ 0xc]
MOV             EDI ,dword ptr [ESP + 0xc]
MOV             ECX ,0x100
MOVSD.REP  ES:EDI ,ESI
```

It takes the addresses of `this->m_entries` and `p_entries` and using `REP MOVSD` copies a string of dwords `ECX` (256) times, which in effect copies all the elements of the source into the destination.

At some point you will start recognizing patterns like this and be able to make an educated guess on what the source could have been. The pseudocode in Ghidra will also start to look familiar to you after a while.

itsmattkc:
Overall, looks great. Only nitpick is that `MxUnknown100dc6b0::Reset` is probably called `MxUnknown100dc6b0::Init` based on naming patterns from the 96 leak, but otherwise I think this is ready to merge.

foxtacles:
> Overall, looks great. Only nitpick is that `MxUnknown100dc6b0::Reset` is probably called `MxUnknown100dc6b0::Init` based on naming patterns from the 96 leak, but otherwise I think this is ready to merge.

True! Fixed it.

PR #43, opened 2023-06-23
Class layout for LEGO1 classes

Cydra:
- Created classes for all MxCore classes, where GetClassName() is used in the VTABLE. Might be incomplete.
- Annotated a bunch of virtual functions calls, but this is incomplete for now.
- Found some members in LegoOmni
- Added classes to MSDEV project and verified that it is compiling


foxtacles:
Hmm, I don't know if it's such a good idea to fill the project with large amounts of what boils down to essentially stubs? Don't get me wrong, knowledge of classes and class relationships is valuable in itself, but imo it would be better served by means of a graphical diagram instead of mostly empty headers and translation units. Information like addresses and such is readily available in IDA, Ghidra, etc. Just my two cents though.

WamWooWam:
At some point it has to become code, I have no issues with this assuming it's correct

foxtacles:
> I have no issues with this assuming it's correct

Yeah, I guess correctness would be difficult to ascertain. Since almost all functions match to 0% it's impossible to determine if the addresses are even right, for one

disinvite:
One benefit to adding all these now is that we won't have nearly as many merge conflicts on the isle.mdp file going forward. Should we take the opportunity to organize things into subdirectories where it makes sense? Separate game objects from the actual engine?

foxtacles:
> One benefit to adding all these now is that we won't have nearly as many merge conflicts on the isle.mdp file going forward.

Unfortunately the `.mdp` file will still have frequent issues since it stores absolute paths to files like this:

`"C:\Users\Christian\Downloads\isle\LEGO1\mxvideoparamflags.cpp"`

so it's going to change all the time regardless. We could just choose to ignore it unless the `.mak` changes as well. I suppose that might be a benefit to this PR; on the other hand, automatically merging the `.mak` works fine most of the time anyway right now.

LEGO1/beachhouseentity.h
```diff
@@ -0,0 +1,19 @@
+#ifndef BEACHHOUSEENTITY_H
+#define BEACHHOUSEENTITY_H
+
+#include "buildingentity.h"
+
+#ifndef undefined4
+#define undefined4 int
+#endif
```
madebr, line 8: Would be useful to put this in a global header `unknown_types.h`, to be used globally.
Grepping for types in this header allows us to find types that are still todo.

Cydra:
I was gone over the weekend, so let me catch up with your comments:

- First off: I guess I should have waited until I filled all vtable calls, but I wanted to get the request out for the weekend. I will be working on this more this week.

- The work done here is thoughed of as a kind of foundation. I started jumping into Ghidra and wrote things down as I go. VTABLE is there, because the information was not there in the Ghidra repo for the first classes I did. I later added the comment to other classes, too, just to make it clearer and as a reference for not Ghidra users.

- SIZE is just a nice metainfo for now that was provided by the MxObjectFactory, I was thinking we could later use C++11 static_assert() to verify class sizes. To see how this could be handled without breaking MSVC4.20, see the last paragraph.

- About the address acurracy: I did my best to copy the addresses, but ofc I could (and have) had wrong adresses in some places. After I am done with all vtables, I will check them all again, but I also would be happy for one of the reviewers to check them, too. Should be easy enough, but this would take some time. Sadly I do not see a way to do this automatically.

- About seperating engine("Mx") and gameplay classes in folders: I like the general idea, but I personally do not think it should be done in this request, just to stay out of complications with existing code. This was also the idea of a similar request, see https://github.com/isledecomp/isle/pull/44. If it is wanted however, I can do this easily.

- I also tried to figure out the IsClass implementation , but sadly failed, because I did not try inlining the calls.  Anyways, thanks to @foxtacles we now have the matching implemenation from https://github.com/isledecomp/isle/pull/48. I will update the classes with this, too. I will also remove the global string reference I used in some places, as this was not 100% accurate.

- @ecumber mentioned my "weird" way to handle undefined types. I was thinking about adding a "decomp_util.h", but I did not like the idea of having a source file that is strictly speaking not part of the "real" sourcecode. Maybe we can have a folder beside ISLE and LEGO1 for this?

Last thing: I would also like to add a preprocessor define for the override specifier introduced in C++11. Basicly this would catch virtual function bugs in non MSVC420 code analyzers/builds and I guess most of us are working with an editor that supports override as a keyword anyways. Basicly it would look like this: 

```
#define CPP_STANDARD __cplusplus
#define CPP_STANDARD_11 199711L

#if CPP_STANDARD >= CPP_STANDARD_11
#define OVERRIDE override
#else
#define OVERRIDE
#endif

// Usage
virtual void VTable0x20() OVERRIDE; // vtable+0x20
```

Any opinions about that? I guess this should be added to the "decomp_util.h" file mentioned for the undefined types.

TL:DR: This pull request was done prematurly and will take some time to be completed, but I will try to make it as good as possible.

foxtacles:
Sounds all good to me. Having all the implementations for `GetClassName`/`IsClass` in the classes that implement them is a fair starting point I think so they don't completely exist in a vacuum without any verification.

madebr:
This also looks good. There are gonna be mistakes, but that's okay.
Personally, I like the `// FIXME`'s instead of adding empty functions: source should document itself.

- About `static_assert`, you can express this also with macro's for C90 code bases.
  Something like
    ```c
    #define STATIC_ASSERT(name, x) typedef int isle_compile_time_assert_ ## name[(x) * 2 - 1]
    ```
    This fails for false inputs because negative array sizes can't exist.
   
- About your `override` suggestion, work for another pr.
- About the `decomp_util.h` header. I think this is needed during this project.
   When/if it is finished, then it can be removed.

tamlin-mike:
Regarding static_assert; something I always do in my rev. projects is:
```
#if defined(_MSC_VER) && (_MSC_VER >= 1900)
#define CT_ASSERT(X) static_assert(X, #X);
#else
#define CT_ASSERT(X) namespace { typedef int foo[(X)?1:-1]; }
#endif
```
This might need some tweaking for MSVC42 in case it lacks `namespace` support.

Then for every type (class/struct):
```
struct foo { ... };
CT_ASSERT(sizeof(foo) == 0x470c); // always
CT_ASSERT(offsetof(foo, m_234) == 0x234); // occasionally
```

Since it can become a bit messy having these compile-time asserts inline, they could probably be placed in a separate source file,  (to only perform the checking once every full build) used only for compile-time checking of this kind.

LEGO1/act2policestation.cpp
```diff
@@ -0,0 +1,9 @@
+#include "act2policestation.h"
+
+// OFFSET: LEGO1 0x100d53ac
```
itsmattkc, line 3: Incorrect address: this is the address of the vtable entry, but not the address of the actual function

LEGO1/act3shark.cpp
```diff
@@ -0,0 +1,4 @@
+#include "act3shark.h"
+
+// 0x100f03a0
+static char* g_act3SharkString = "Act3Shark";
```
itsmattkc, line 4: I'm fairly certain this is not a separate global static var. All `ClassName()` implementations return a string, but the string is stored in the data section of the executable, so the result is it always returns the address of a seemingly static variable.

LEGO1/act3state.h
```diff
@@ -0,0 +1,11 @@
+#ifndef ACT3STATE_H
+#define ACT3STATE_H
+
+#include "legostate.h"
+
+class Act3State : public LegoState
+{
+
```
itsmattkc, line 8:
```suggestion
// VTABLE 0x100d7920
```

LEGO1/beachhouseentity.h
```diff
@@ -0,0 +1,19 @@
...
+
+#include "buildingentity.h"
+
+#ifndef undefined4
+#define undefined4 int
+#endif
+
+class BeachHouseEntity : public BuildingEntity
+{
+  virtual undefined4 VTable0x50(); // vtable+0x50
```
itsmattkc, line 12: Is there a purpose to declaring this function and defining a type (`undefined4`) for it?

LEGO1/bike.h
```diff
@@ -0,0 +1,24 @@
...
+#endif
+
+class Bike : public IslePathActor
+{
+public:
+  Bike();
+
+  virtual undefined4 VTable0xcc(); // vtable+0xcc
+  virtual undefined4 VTable0xd4(undefined4 param); // vtable+0xd4
+  virtual void VTable0xe4(); // vtable+0xe4
```
itsmattkc, line 17: What's the purpose of declaring any of these?

LEGO1/bumpbouy.cpp
```diff
@@ -0,0 +1,18 @@
+#include "bumpbouy.h"
+
+// 0x100f0394
+static char* g_bumpBouyString = "BumpBouy";
```
itsmattkc, line 4: This is almost certainly not a static string. It's probably an inlined string that's been moved to the data section of the PE, as I explained in an earlier comment.

itsmattkc:
There are a few issues throughout. I may go through and clean it up myself later today because if we're going to merge this, it should be done sooner rather than later (a lot of PRs are waiting on it right now)

LEGO1/act3shark.cpp
itsmattkc: Ahh my bad. In my haste, I missed that comment and didn't realize this was still being worked on.

Still, there are a few PRs semi-waiting for this now, and I'm still strongly considering fixing up the remaining issues myself so it can be merged.

Cydra: Sure, I have a bunch of local changes and merged 4db49fe6a6e3480413b37706c41fed51163f82b0 in. 

Do you want me to finish up what I have currently or bring over the version here?

itsmattkc: Up to you, if you have stuff you'd like to finish, I think you may as well

Cydra: Updated! If you want I can merge the current master back in, otherwise feel free to take it over and do it yourself.

LEGO1/legoanimactor.cpp
```diff
@@ -0,0 +1,24 @@
...
+const char* GetClassNameInternal()
+{
+  return g_legoAnimActorClassName;
+}
+
+// OFFSET: LEGO1 0x1000fb90 STUB
+const char *LegoAnimActor::GetClassName() const
+{
+  return GetClassNameInternal();
+}
```
madebr, line 16: These are caused by multiple inheritance, and automatically generated by the compiler. I think.

LEGO1/act3state.cpp
```diff
@@ -0,0 +1,13 @@
+#include "act3state.h"
+
+// OFFSET: LEGO1 0x10072270 STUB
```
madebr, line 3: At this address, there is `Act3::Act3`.
`Act3State` is created in ` LegoObjectFactory::Create` int 10009a90.

LEGO1/act3state.h
```diff
@@ -0,0 +1,26 @@
...
+public:
+  Act3State();
+
+  // OFFSET: LEGO1 0x1000e300
+  inline virtual const char *GetClassName() const // vtable+0x0c
+  { 
+    return "Act3State";
+  }; 
+
+  // OFFSET: LEGO1 0x100d4fd8
```
madebr, line 17: The vtable address was given.
```suggestion
  // OFFSET: LEGO1 0x1000e310
```

LEGO1/bumpbouy.h
```diff
@@ -0,0 +1,13 @@
+#ifndef BUMPBOUY_H
+#define BUMPBOUY_H
+
+#include "mxbool.h"
+
+class BumpBouy
```
madebr, line 6: This class has multiple inheritance.

LEGO1/elevatorbottom.cpp
```diff
@@ -0,0 +1,52 @@
...
+// OFFSET: LEGO1 0x10017f30 STUB
+MxBool ElevatorBottom::IsClass(const char *name) const
+{
+  // TODO
+
+  return MxBool();
+}
+
+// OFFSET: LEGO1 0x10017f10 STUB
+undefined4 ElevatorBottom::VTable0x5c()
```
madebr, line 41: Return type is 1 byte
```suggestion
undefined ElevatorBottom::VTable0x5c()
```

LEGO1/gasstation.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include "legoworld.h"
+
+#ifndef undefined
+#define undefined int
+#endif
+
+#ifndef undefined4
+#define undefined4 int
+#endif
```
madebr, line 12: Move to `decomp_typedefs.h` header please.
Also, undefined should be a char.

LEGO1/helicopterstate.cpp
```diff
@@ -0,0 +1,13 @@
...
+// OFFSET: LEGO1 0x1000e0b0
+MxBool HelicopterState::VTable0x14()
+{
+  return MX_FALSE;
+}
+
+// OFFSET: LEGO1 0x1000e0c0 STUB
+MxBool HelicopterState::VTable0x18()
+{
+  return MX_TRUE;
```
madebr, line 12: This is incorrect.
```suggestion
  return MX_FALSE;
}

// OFFSET: LEGO1 0x1000e0c0 STUB
MxBool HelicopterState::VTable0x18()
{
  // TODO

  return MX_TRUE;
```

madebr, line 12: This is incorrect
```suggestion
  // TODO

  return MX_TRUE;
```

LEGO1/isleactor.cpp
```diff
@@ -0,0 +1,26 @@
+#include "isleactor.h"
+
+// 0x100f07dc
+static char* g_isleActorClassName = "IsleActor";
+
+// OFFSET: LEGO1 0x100d5178 STUB
```
madebr, line 6:
```suggestion
// OFFSET: LEGO1 0x1002c7b0 STUB
```

LEGO1/islepathactor.cpp
```diff
@@ -0,0 +1,21 @@
+#include "islepathactor.h"
+
+// OFFSET: LEGO1 0x10001e60 STUB
+IslePathActor::IslePathActor()
```
madebr, line 4: `Helicopter::Helicopter` is at this address.
You probably mean 0x1001a200
```suggestion
// OFFSET: LEGO1 0x1001a200 STUB
IslePathActor::IslePathActor()
```

LEGO1/legoanimactor.cpp
madebr: The real method is at `0x1000fba0`.

LEGO1/legoanimactor.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+// OFFSET: LEGO1 0x1000fbb0 STUB
+MxBool LegoAnimActor::IsClass(const char *name) const
```
madebr, line 19: This one is also messed up due to multiple inheritance.
It's at 0x1000fbc0.

LEGO1/legoentity.cpp
```diff
@@ -0,0 +1,107 @@
+#include "legoentity.h"
+
+// OFFSET: LEGO1 0x100105f0 STUB
+LegoEntity::LegoEntity()
```
madebr, line 4: Incorrect offset.
`LegoEntity` is created `inLegoObjectFactory::Create` at around `1000a3f3`.

madebr: (It's inlined)

LEGO1/legoentity.cpp
```diff
@@ -0,0 +1,107 @@
...
+void LegoEntity::VTable0x20(char *param)
+{
+  // TODO
+}
+
+// OFFSET: LEGO1 0x100108a0 STUB
+void LegoEntity::VTable0x24(undefined4 param_1, undefined1 param_2, undefined1 param_3)
+{
+  // TODO
+}
```
madebr, line 47: I don't think this one is correct.
It's also present in a vtable at `100d487c`.

LEGO1/legoentity.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+// OFFSET: LEGO1 0x10001090 STUB
+void LegoEntity::VTable0x30(undefined4 param)
+{
+  // TODO
+}
+
+// OFFSET: LEGO1 0x10010f10 STUB
+void LegoEntity::VTable0x34(undefined1 param)
```
madebr, line 68: I've seen `undefined` and `undefined1` (move to some global header)

LEGO1/legoinputmanager.h
```diff
@@ -16,10 +16,16 @@ __declspec(dllexport) enum NotificationId
 class LegoInputManager
```
madebr, line 16: This should subclass `MxPresenter`.

LEGO1/legoworld.cpp
```diff
@@ -0,0 +1,27 @@
...
+// OFFSET: LEGO1 0x10022080 STUB
+long LegoWorld::Tickle()
+{
+  // TODO
+
+  return 0;
+}
+
+// OFFSET: LEGO1 0x1001e9d0 STUB
+void LegoWorld::Destroy()
```
madebr, line 24: This has an argument
```suggestion
void LegoWorld::Destroy(undefined4 u)
```

LEGO1/legoworld.h
```diff
@@ -1,11 +1,16 @@
...
+
+class LegoWorld : public LegoEntity
 {
 public:
   __declspec(dllexport) LegoWorld();
-  __declspec(dllexport) virtual ~LegoWorld();
+  __declspec(dllexport) virtual ~LegoWorld(); // vtable+0x0
+
+  virtual long Tickle(); // vtable+0x8
+  virtual void Destroy(); // vtable+0x1c
```
madebr, line 13:
```suggestion
  virtual void Destroy(undefined4 u); // vtable+0x1c
```

LEGO1/mxcompositemediapresenter.cpp
```diff
@@ -0,0 +1,7 @@
+#include "mxcompositemediapresenter.h"
+
+// OFFSET: LEGO1 0x10073ea0 STUB
+MxCompositeMediaPresenter::MxCompositeMediaPresenter()
```
madebr, line 4: The destructor is at 0x10074020

LEGO1/mxmusicpresenter.cpp
```diff
@@ -0,0 +1,7 @@
+#include "mxmusicpresenter.h"
+
+// OFFSET: LEGO1 0x100c22c0 STUB
```
madebr, line 3: Destructor at 0x100c24e0

LEGO1/mxpresenter.cpp
```diff
@@ -0,0 +1,134 @@
...
+void MxPresenter::ParseExtra()
+{
+  // TODO
+}
+
+// OFFSET: LEGO1 0x1000c070 STUB
+MxPresenter::~MxPresenter()
+{
+  // TODO
+}
```
madebr, line 31: This is the first element of the vtable.
I don't think `MxPresenter` has overridden its destructor.

LEGO1/mxsoundmanager.h
```diff
@@ -1,9 +1,14 @@
 #ifndef MXSOUNDMANAGER_H
 #define MXSOUNDMANAGER_H
 
-class MxSoundManager
+#include "mxcore.h"
+
+class MxSoundManager : public MxCore
```
madebr, line 6: This inheritance is not correct.
There are 2 levels of inheritance between `MxSoundManager` and `MxCore`.

LEGO1/mxstillpresenter.cpp
```diff
@@ -0,0 +1,7 @@
+#include "mxstillpresenter.h"
+
+// OFFSET: LEGO1 0x10043430 STUB
```
madebr, line 3: This is not the constructor of `MxStillPresenter`, but of a subclass of `MxStillPresenter`.
`MxStillPresenter`'s vtable is at 0x100d7a38
At 0x10043430, there is a sconstructor with vtable at 0x100d7ac8

LEGO1/mxtransitionmanager.cpp
```diff
@@ -0,0 +1,29 @@
+#include "mxtransitionmanager.h"
+
+// OFFSET: LEGO1 0x1004ba00 STUB
```
madebr, line 3: This is the address of MxTransitionManager's destructor

madebr:
```suggestion
// OFFSET: LEGO1 0x1004b8d0 STUB
```

LEGO1/mxwavepresenter.h
```diff
@@ -0,0 +1,20 @@
+#ifndef MXWAVEPRESENTER_H
+#define MXWAVEPRESENTER_H
+
+#include "mxsoundpresenter.h"
+
+class MxWavePresenter : public MxSoundPresenter
```
madebr, line 6: Destructor of MxWavePresenter is at 0x1000d640

LEGO1/pizzeriastate.cpp
```diff
@@ -0,0 +1,15 @@
+#include "pizzeriastate.h"
+
+// OFFSET: LEGO1 0x10015aa0 STUB
```
madebr, line 3: LegoRace::LegoRace is at this address.
It should be 10017af0

LEGO1/racestandsentity2.h
```diff
@@ -0,0 +1,11 @@
+#ifndef RACESTANDS2ENTITY_H
+#define RACESTANDS2ENTITY_H
+
+#include "buildingentity.h"
+
+class RaceStandsEntity2 : public BuildingEntity
```
madebr, line 6: I think there are 3 `RaceStandsEntity`s,
with vtables at 100d48a8 100d50c0 100d5200
Looks like they don't have a common ancestor, as `GetClassName` is not shared.

LEGO1/radio.h
```diff
@@ -0,0 +1,14 @@
+#ifndef RADIO_H
+#define RADIO_H
+
+#include "mxcore.h"
+
+class Radio : public MxCore
+{
+public:
+  virtual ~Radio();
```
madebr, line 9: constructor is at 1002c850

LEGO1/racestandsentity2.h
tamlin-mike: Since I also noticed that, after some thought... could they be inner classes of the respective race type they belong to? That's about the only way I can see how it makes sense they can have the same name.

itsmattkc:
Man this is such a massive PR. I spent the last 3+ hours trying to review/verify it and I'm not even half way through. I very much appreciate the effort, but I think I'm going to have to make a rule against large PRs like this. Ideally this should have been broken down into smaller steps. As it is, it's very unwieldy, and it becomes much more likely that someone is going to make a mistake along the way.

LEGO1/racestandsentity2.h
madebr: If it's an inner class, then the og code has duplicated the name string.
Perhaps this is a template class?
Time to find out what code msvc generates in this case..

madebr: I just verified with MSVC 4.2 that the following source will emit 3 identical `name` methods.
The old linker does not have comdat folding, only `/opt:ref` (lucky for us)
https://learn.microsoft.com/en-us/cpp/build/reference/opt-optimizations

```c++
#define DLLEXPORT __declspec(dllexport)

class Core {
public:
  virtual const char *name() const {
    return "Core";
  }
};

template <class T>
class TemplateChild : public Core {
public:
  virtual const char *name() const {
    return "TemplateChild";
  }
  const T* data() const {return &m_data;}
private:
  T m_data;
};

// Create 3 exports to force the compiler to emit code for the class
DLLEXPORT TemplateChild<int> *getIntObj() {  return new TemplateChild<int>(); }
DLLEXPORT TemplateChild<float> *getFloatObj() {  return new TemplateChild<float>(); }
DLLEXPORT TemplateChild<double> *getDoubleObj() {  return new TemplateChild<double>(); }
```
Compile, and link this with:
```
cl /c a.cpp /Foa.obj
link a.obj /out:a.dll /dll /debug
```
(Looks like Ghidra does not understand the old PDB format)

tamlin-mike: > The old linker does not have comdat folding

Oh. Oh! That's good news! That's something one always have to look out for and create rev. cludges for otherwise, not to mention it allows to know with 100% certainty that the (non-std) std collection types I have so far identified are of one single concrete type each. Thanks for the info.

(it's been about three decades since I actually used 4.2, so I'm not surprised I had forgotten these pre-std limitations and quirks)

Cydra:
I agree, I wanted to much for this PR. Doing a be-all-end-all solution was the wrong way to go. Sorry that this is such a hassle now.

Logically this should have been split up in:
- Creation of the class files with inheritance
- Constructors/Destructors
- MxCore overrides
- VTable references
- Unrelated edits (MxOmni)

I can seperate the PR into those parts, but I guess it is to late now, as the merge process already started.

LEGO1/legoanimactor.cpp
madebr: I just verified this by reversing a diamond class hierarchy, using virtual inheritance.

LEGO1/jetskiRace.cpp
```diff
@@ -0,0 +1,15 @@
+#include "jetskiRace.h"
+
+// OFFSET: LEGO1 0x10015aa0 STUB
```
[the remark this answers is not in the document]
tamlin-mike: Looking at the generated code @ 0x1000B74D, we can scratch "may". `JetskiRaceState` has an empty constructor, whether it's compiler-generated or user defined in the header.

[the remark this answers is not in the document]
madebr: I agree. It's probably not specialized. Thanks for checking.

LEGO1/legoentity.cpp
tamlin-mike: The argument is with very high probability MxBool.

LEGO1/legoworld.cpp
tamlin-mike: The argument is with very high probability MxBool. See 1001EE4F.

LEGO1/mxpresenter.cpp
tamlin-mike: I think I agree. 0x1000C070 is the scalar deleting destructor. Judging by the generated code, the destructor itself is empty (apart from the obvious call to destruct the aggregated MxCriticalSection). Whether user-defined in header or just compiler generated is a coin-toss.

LEGO1/mxstillpresenter.cpp
tamlin-mike: 0x100D7AC8 is the vftable for `LegoMeterPresenter`. Evidence:
0x10043430 default ctor of the type. Only called by LegoObjectFactory::Create.
0x1000A15E compares the requested type argument against LegoObjectFactory.m_7C_MxAtomId_LegoMeterPresenter
0x10006FC8 (LegoObjectFactory ctor) grabs the pointer to m_7c and store is in local var `p`.
10007A0A-10007A26 `*p = MxAtomId("LegoMeterPresenter", (LookupMode)0);`
Elementary. :) *phew*

LEGO1/racestandsentity2.h
tamlin-mike: 100D5200 JailEntity
100D50C0 CaveEntity
100D48A8 RaceStandsEntity
See my comment above on finding LegoMeterPresenter.

itsmattkc:
I have finished reviewing this and have fixed and verified it as much as I feel I can. I didn't keep everything, there was just so much to keep track of all at once. There may still be issues I didn't catch, or maybe even a few I introduced myself, but again if we're going to merge this, it's better to do it sooner rather than later. Feel free to submit PRs if you do find anything I missed; I know there are quite a few notes in this thread, many of which I didn't directly address (again, already too much to keep track of).

If you do make PRs for this stuff (or anything else really), **please PR one thing at a time**. I will not be accepting any more "meta PRs" that make changes throughout a huge chunk of the files. They screech the whole project to a halt because they take so long to review, and they cause a high probability of merge conflicts with other PRs since they touch so many files. There is now a rule/guideline in `CONTRIBUTING.md` to limit the size of PRs, preferably to one class at a time where possible.

Thanks for the work, I will now merge this.

LEGO1/mxstillpresenter.cpp
madebr: I completely agree with this :)

LEGO1/racestandsentity2.h
madebr: 100% agree!

PR #44, opened 2023-06-24
Add basic implementations of most classes

ecumber:
I have added most classes that have the GetClassName() function. These are most definitely incorrect because I have not taken into account inheritance, etc. but the main purpose of this PR is to just get these added.

I would also suggest that the source files in the LEGO1 folder be organized into subfolders as the folder has become large, but I do not want to break any other PRs at this time.


ecumber:
Just noticed this is the same as #43. Oops

PR #45, opened 2023-06-24
isle: match WndProc and re-arrange functions in `ISLE.EXE`

foxtacles:
The last remaining funtion in `ISLE.EXE` has been matched. There are two minor instruction/comparison swaps left in the `WM_DISPLAYCHANGE` case (as seen in other functions before). I didn't manage to further influence it, it might be impossible. Could be related to `m_unk74` not being uncovered yet, but I'd consider it done for now.

Matching this function was really hard, it's extremely delicate with where variables must be declared and how. Even very small changes in positioning can cause the entire function to go down from 97% to less than 30%. The compiler algorithms must be moving closely around tipping points with regards to register allocations etc.

Also, **the last remaining (known) functional bug of `ISLE.EXE` is fixed** (the "leave bricks" bug). It no longer crashes when attempting to exit the game through the Information Center door. The game seems to rely on undefined behavior here, it's only working when `type` (`NotificationId`) is _not_ initialized.

I've also taken the liberty of preliminary arranging the functions to put them as close to the original offsets as possible. To make this possible, I've removed the `main.cpp` TU. Interestingly, this also resolved a few minor diffs in some functions, so the state we have arrived at now looks like:

image

The offsets are almost all correct. There is something extra (some dtor related compiler-generated artifact) sitting after `SetupLegoOmni` that isn't present in that way in the decomp (it's located elsewhere). I'm not sure where that difference comes from at this time. Might be related to `MxOmniCreateParam` and its mysterious base class, or `MxOmniCreateFlags`. Could also be we have to arrange things a bit differently, but I think this is a fair starting point.


foxtacles:
Okay, so far I've not been making much progress. The key issue is that currently, the original `WinProc` uses one extra register (`EDI`) whereas the decomp one does not. Apart from that, the control flow and all the logic looks identical to me so I think the layout of the function is correct. 

I've tried moving things around a bit, adding variables here and there etc. but no luck so far. If you have any suggestions, let me know!

foxtacles:
Register allocation matches now, got bumped up to ~50% - many instructions still appear in different order though

theR4K:
Hi, I just figure out that link.exe have undocumented flag "/Brepro", which drops the PE timestamps. This can be helpful to minimize difference between two build. Thanks!

itsmattkc:
Fantastic work, I always knew WndProc would be a pain in the ass, I'm glad you were able to figure it out. That `99%` is very sexy.

PR #46, opened 2023-06-24
lego: Implement LegoBackgroundColor

MishaProductions:
Also added editorconfig from matrix chat


MishaProductions:
I don't really know what's causing for there to be another local variable to be added in lego1 (local_14) by the 3 atoi() calls. The RE'd function uses videomanager??
image

LEGO1/mxbackgroundcolor.cpp
```diff
@@ -0,0 +1,22 @@
...
+// OFFSET: LEGO1 0x1003bea0
+MxString *MxBackgroundColor::GetColorString()
+{
+  return &m_colorString;
+}
+
+// OFFSET: LEGO1 0x1003beb0
+void MxBackgroundColor::SetColorString(const char *colorString)
+{
+  m_colorString.operator=(colorString);
```
madebr, line 21:
```suggestion
  m_colorString = colorString;
```

LEGO1/mxbackgroundcolor.cpp
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// OFFSET: LEGO1 0x1003bea0
+MxString *MxBackgroundColor::GetColorString()
+{
+  return &m_colorString;
+}
+
+// OFFSET: LEGO1 0x1003beb0
+void MxBackgroundColor::SetColorString(const char *colorString)
```
madebr, line 19: I doubt the names of this class.
These behave more like variables with callbacks.

The `GetName` method is used in 3 classes:
```
  100d7498, 100d74a8, 100d74b8, 100d8698
  100d86a8, 100d86b8, 100d86c8, 100da878
```
These vtables have the 1st and 3rd member in common, only the 2nd differs (which accepts a `const char*` and modifies the global state)

LEGO1/mxbackgroundcolor.h
```diff
@@ -0,0 +1,22 @@
...
+  virtual void SetColorString(const char *colorString);
+  virtual ~MxBackgroundColor();
+
+protected:
+  MxString m_name;
+  MxString m_colorString;
+  float h;
+  float s;
+  float v;
+};
```
madebr, line 20: Please document which vtable exactly you think belongs to `MxBackgroundColor`.

.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*.{cpp,h}]
+indent_style = space
+indent_size = 2
```
madebr, line 3:
```suggestion
root = true

[*.{cpp,h}]
indent_style = space
indent_size = 2
insert_final_newline = true
trim_trailing_whitespace = true
```

LEGO1/legobackgroundcolor.cpp
```diff
@@ -0,0 +1,54 @@
+#include "legobackgroundcolor.h"
+#include "legoomni.h"
+#include "legoutil.h"
+#include "legovideomanager.h"
+const char *Delimiter = "\t";
+const char *set = "set";
+const char *reset = "reset";
```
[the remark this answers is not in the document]
MishaProductions: The accuracy drops when I do that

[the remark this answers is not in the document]
madebr: And/or, `/GF` for read-only string pooling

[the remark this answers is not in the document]
MishaProductions: I tried that and it had the same result

itsmattkc:
We've switched build systems, please merge/rebase when you get a chance.

.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
+[*.{cpp,h}]
```
madebr, line 2:
```suggestion
[*.{cpp,h,py,txt,editorconfig}]

```

MishaProductions:
> We've switched build systems, please merge/rebase when you get a chance.

Done

itsmattkc:
I merged this locally and for some reason it didn't register with this PR.

This is now merged in a66dbb8136a718e9217e00acc006ab905441c636

PR #47, opened 2023-06-25
Premake5

theR4K:
Added premake5 config and binary
Added stubs to fix isle.exe build using compiled lego1.lib
Added nmake2 module for premake implementing .mak files generation support


theR4K:
So, nmake2 module still in progress(At least I want to change cl and link arguments arrangement).
And it still possible that I miss some compiler options, but it must be easy to fix using temporally methods (add buildoptions to premake config) or by updating https://github.com/theR4K/nmake4premake

madebr:
This pr is doing 2 things: 1. adding premake and 2. adding a lot of classes

Perhaps you can only add premake, and let ISLE.EXE link with the import library in `ISLE/ext`?
Alternatively, `lib /def:ISLE/ext/lego1.def /machine:ix86 /out:imp.lib` will also create an import library.

theR4K:
yes, you a right. I can do same HACK as original .mak did, but I think it have no sense to do something that pointless.

theR4K:
Hi, I did a quick test, and now I know:
for cl.exe call doesn't matter the order of arguments, only content. And /Gm don't affect final .exe.
for link.exe:
1) relative position of .objs, .libs and arguments doesn't matter. But
2) relative positions of .objs itself matter
3) as well as .libs relative position(between each other) matter
4) flags order does not matter
5) /incremental: yes by default

So, based on this. I can tell: right now nmake2 module doing well.
We can control order of .libs, and almost all flags needed (and all flags using buildoptions/linkoptions config).
But we can't(or at least not very reliable) control .objs order. This can be improved.

Finally. I'm expecting good results from this, (maybe with some config tweaks)

theR4K:
usage:
1) premake5 nmake2 - to create .mak script
2) cd to created "build" directory. Here created project files files appears.
3) nmake /f build.mak [CFG=] all - to build both LEGO1 and ISLE

PS. clean target still need some work. for now it almost work.
Thanks, The_R4K_!

theR4K:
Any comments? Any suggestions?

theR4K:
image
build results

theR4K:
oh I didn't include total accuracy. It's  Total accuracy 99.11% across 16 functions. Same as original .mak file.

theR4K:
@madebr ?

theR4K:
any comments? any suggestions? does anybody have an any issue while build?

itsmattkc:
I think demand for this is fairly low since the switch to CMake. I feel any improvements this may have introduced can probably be ported to that system. As such, I'm going to close this. Thanks anyway for the PR.

theR4K:
ok, but premake much more universal. you don't need to specify flags for compilers, and you can have any tool chain natively. So, premake is preferrable. But as you wish....

theR4K:
as, I mentioned you can generate cmake from premake(as well as .sln and gmake), but can't do this backward....

theR4K:
> I think demand for this is fairly low since the switch to CMake. I feel any improvements this may have introduced can probably be ported to that system. As such, I'm going to close this. Thanks anyway for the PR.

as, I mentioned you can generate cmake from premake(as well as .sln and gmake), but can't do this backward....

PR #49, opened 2023-06-25
Improve reccmp.py

stravant:
* Now only shows the info for a single function when a specific function is specified via -v

* Now colors the output by default

* Percentages are shown as green/yellow/red depending on the percentage completed.

* Diff +/- lines are shown as green/red.

* Includes standard --no-color argument in case we need no color for some tooling which consumes the output.

### DIFF
--- a/tools/reccmp/reccmp.py
+++ b/tools/reccmp/reccmp.py
@@ -7,6 +7,7 @@ import struct
 import subprocess
 import os
 import sys
+import colorama
 
 parser = argparse.ArgumentParser(allow_abbrev=False,
   description='Recompilation Compare: compare an original EXE with a recompiled EXE + PDB.')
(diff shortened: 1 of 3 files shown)


tools/reccmp/reccmp.py
```diff
@@ -7,6 +7,7 @@
 import subprocess
 import os
 import sys
+import colorama
```
madebr, line 10: You need to install colorama.
Perhaps add a `tools/reccmp/requirements.txt` file, and install this on ci as well.

tools/reccmp/reccmp.py
```diff
@@ -305,23 +319,53 @@ def parse_asm(file, addr, size):
...
+                  else:
+                    print(line)
+                  if not plain:
+                    print(colorama.Style.RESET_ALL, end='')
+
+                print("\n%s is only %s similar to the original, diff above" % (recinfo.name, percenttext))
+
+            # If html, record the diffs to an HTML file
+            if html:
+              htmlinsert.append('{address: "%s", name: "%s", matching: %s, diff: "%s"}' % (hex(addr), recinfo.name, str(ratio), '\\n'.join(udiff).replace('"', '\\"').replace('\n', '\\n')))
```
madebr, line 368: Looks like you're creating json with strings.
How about using python dicts and lists to gather all data,
and at the end use [`the json module`](https://docs.python.org/3/library/json.html) to create a json string.

stravant: This is an existing line, just shuffled around, not going to go cleaning up the whole script at the moment.

tools/reccmp/reccmp.py
stravant: What do you think about not importing colorama if --no-color is passed?

That way CI doesn't have to worry about installing colorama when it doesn't need color to begin with.

madebr: Personally, I prefer to test things. This avoids silently breaking code.
ci does `pip install capstone` already.

stravant: Added requirements.txt and updated build.yaml

tools/reccmp/reccmp.py
madebr: Let's hope matt reads this ;)

itsmattkc: I guess I never thought of it as JSON because it's just inserting data into the HTML template, but technically yeah I guess it's a JavaScript object, so it's JSON.

PR #50, opened 2023-06-25
Add a CMakeLists.txt + build it on CI

madebr:
This pr adds a basic CMakeLists.txt that generates binaries that are very similar to the ones generates by the current system.

The CMake script has 2 options:
- `ISLE_ACCURACY` that allows configuring the build options such that they are the same as the existing one.
  It is enabled by default for older MSVC compilers, and disabled by default for others.
- `ISLE_WERROR`: treat warnings as errors

The change to the `mxomni.h` header was needed because the linker complained about a duplicate `MxOmni::MxOmni` symbol.

I left the original build system in place.


stonedDiscord:
i feel like the cmake build and the accuracy report should be seperate jobs and not part of one big one

stonedDiscord:
you can make the accuracy check run after the build with the job.needs argument
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

madebr:
> i feel like the cmake build and the accuracy report should be seperate jobs and not part of a single big one

Nice idea, but please add this in a future pr.
This one does the minimum amount of work to get cmake building, and I don't want to complicate it further at this point.

PR #51, opened 2023-06-26
100% Match of MxDSFile

stravant:
* ...almost, MxDSFile::Open is still not quite matching but all of the other methods are 100% matching.

* Turns out that most of the virtual methods and some of the members are actually on the MxDSSource base class, which I've pulled out as part of this.

* In order to implement the methods I added the MXIOINFO class, which seems to be a thin wrapper around the MMIOINFO windows structure. We can tell this because MMIOINFO::~MMIOINFO was included in the DLL exports, and calls down to a function which deconstructs something looking exactly like MMIOINFO.

* Stubbed out most of the MXIOINFO methods, they can be decomped later.


LEGO1/mxdsfile.h
```diff
@@ -1,20 +1,46 @@
...
-  __declspec(dllexport) virtual long Read(unsigned char *,unsigned long);
-  __declspec(dllexport) virtual long Seek(long,int);
+  __declspec(dllexport) MxDSFile(const char *filename, unsigned long skipReadingChunks);
+  __declspec(dllexport) ~MxDSFile();
+  __declspec(dllexport) long Open(unsigned long);
+  __declspec(dllexport) long Close();
+  __declspec(dllexport) long Read(unsigned char *,unsigned long);
+  __declspec(dllexport) long Seek(long,int);
+  __declspec(dllexport) unsigned long GetBufferSize();
+  __declspec(dllexport) unsigned long GetStreamBuffersNum();
```
itsmattkc, line 18: Why did you remove the `virtual` from these? We know from the DLL exports that these are definitely `virtual`.

LEGO1/mxstring.h
```diff
@@ -16,6 +16,8 @@ class MxString : public MxCore
   void ToLowerCase();
   const MxString &operator=(MxString *);
 
+  inline const char *Data() const { return m_data; }
```
itsmattkc, line 19: `GetData` would be preferable to `Data`

LEGO1/mxdsfile.cpp
```diff
@@ -1,6 +1,139 @@
...
+}
+
+// OFFSET: LEGO1 0x100cc620
+long MxDSFile::ReadChunks()
+{
+  _MMCKINFO topChunk;
+  _MMCKINFO childChunk;
+  char tempBuffer[80];
+  
+  topChunk.fccType = 0x494e4d4f;
```
itsmattkc, line 65: Might be good to add a comment mentioning what this is (it's `OMNI` in a little-endian integer)

LEGO1/mxdsfile.cpp
```diff
@@ -1,6 +1,139 @@
...
+  if (m_io.Descend(&topChunk, NULL, MMIO_FINDRIFF) != 0) {
+    return -1;
+  }
+  childChunk.ckid = 0x6448784d;
```
itsmattkc, line 69: Same here (`MxHd`)

LEGO1/mxdsfile.cpp
```diff
@@ -1,6 +1,139 @@
...
+  if (m_io.Descend(&childChunk, &topChunk, 0) != 0) {
+    return -1;
+  }
+
+  m_io.Read((char*)&m_header, 0xc);
+  if ((m_header.majorVersion == SI_MAJOR_VERSION) && (m_header.minorVersion == SI_MINOR_VERSION))
+  {
+    childChunk.ckid = 0x664f784d;
```
itsmattkc, line 77: And here `MxOf`

LEGO1/mxstring.h
itsmattkc, line 19: `GetData` is probably preferable to `Data`, according to the 96 source style guide

LEGO1/mxdssource.cpp
```diff
@@ -0,0 +1,14 @@
+#include "mxdssource.h"
+
+// OFFSET: LEGO1 0x100bffd0
+void MxDSSource::SomethingWhichCallsRead(void* pUnknownObject)
```
itsmattkc, line 4: In the absence of a better name, it might be better to keep it generic like `FUN_100bffd0`. I feel like vague naming could complicate things in situations where a lot of functions appear to do very similar things.

LEGO1/mxioinfo.h
```diff
@@ -1,10 +1,22 @@
 #ifndef MXIOINFO_H
 #define MXIOINFO_H
 
+#include "windows.h"
```
itsmattkc, line 4: Including `windows.h` is known to cause issues because of their macros. Instead, include `legoinc.h` which handles the macros automatically (there should be a CI test at some point to verify this)

LEGO1/mxdsfile.cpp
```diff
@@ -1,6 +1,126 @@
 #include "mxdsfile.h"
 
+#include "stdio.h"
```
itsmattkc, line 3: Very minor, but as stdio is an external header, it should be `<stdio.h>`

LEGO1/mxdsfile.cpp
```diff
@@ -1,6 +1,126 @@
...
+    unsigned long* pLengthInDWords = &m_lengthInDWords;
+    m_io.Read((char *)pLengthInDWords, 4);
+    m_pBuffer = malloc(*pLengthInDWords * 4);
+    m_io.Read((char*)m_pBuffer, *pLengthInDWords * 4);
+    return 0;
+  }
+  else
+  {
+    sprintf(tempBuffer, "Wrong SI file version. %d.%d expected.", SI_MAJOR_VERSION, SI_MINOR_VERSION);
+    MessageBoxA(NULL, tempBuffer, NULL, 0x10);
```
itsmattkc, line 90: `0x10` resolves to `MB_ICONERROR`, would be good to use that

itsmattkc:
Overall looks pretty good, just a few minor notes I had for it

LEGO1/mxdsfile.h
stravant: I guess I could leave the virtual on here too but they're now virtual by default thanks to the base class having them as virtual methods.

LEGO1/mxdsfile.cpp
stravant: CoPilot knows what's up lol: 
image

LEGO1/mxdssource.cpp
stravant: I agree, except in the case of virtual methods like this one.

IMO having as much info as possible there for virtual methods makes sense because you can't easily see / refine their naming later through XREFs in Ghidra.

LEGO1/mxdsfile.cpp
stravant: No worries, I'm all for good style, don't worry about nitpick comments.

stravant:
Applied all feedback except for the SomethingWhichCallsRead (explained above), let me know if you still think it should be left with a generic name.

LEGO1/mxdsfile.cpp
```diff
@@ -1,6 +1,130 @@
 #include "mxdsfile.h"
 
+#include <stdio.h>
+
+#define SI_MAJOR_VERSION 2
+#define SI_MINOR_VERSION 2
+
+#define FCC_OMNI 0x494e4d4f
+#define FCC_MxHd 0x6448784d
+#define FCC_MxOf 0x664f784d
```
madebr, line 10: For another project, I created `FOURCC` macros for this:
```c
#define FOURCC(a, b, c, d) (((a) << 0) | ((b) << 8) | ((c) << 16) | ((d) << 24))
```

This allows you to redefine these macros as:
```c
#define FCC_OMNI FOURCC('O', 'M', 'N', 'I')
#define FCC_MxHd FOURCC('M', 'x', 'H', 'd')
#define FCC_MxOf FOURCC('M', 'x', 'O', 'f')
```

LEGO1/mxdsfile.cpp
```diff
@@ -1,6 +1,130 @@
...
+long MxDSFile::Open(unsigned long uStyle)
+{
+  // No idea what's stopping this one matching, but I'm pretty
+  // confident it has the correct behavior.
+  memset(&m_io, 0, sizeof(MXIOINFO));
+  if (m_io.Open(m_filename.GetData(), uStyle) != 0) {
+    return -1;
+  }
+
+  m_io.SetBuffer(NULL, 0);
```
madebr, line 35:
```suggestion
  m_io.SetBuffer(m_io, 0);
```

LEGO1/mxdssource.cpp
madebr: Agree with stravant. This is a method in local namespace, not globally visible.
So there's no chance on collision.

LEGO1/mxioinfo.cpp
```diff
@@ -0,0 +1,49 @@
...
+// OFFSET: LEGO1 0x100cc800
+MXIOINFO::MXIOINFO()
+{
+  memset(&m_info, 0, sizeof(MMIOINFO));
+}
+
+// OFFSET: LEGO1 0x100cc820
+MXIOINFO::~MXIOINFO()
+{
+  m_info.cchBuffer = 0;
```
madebr, line 12:
```suggestion
  Close();
```

LEGO1/mxdsfile.cpp
tamlin-mike: Not surprising you see difference here. SetBuffer (0x100ccbc0) takes _three_ args, not two. Probably it's inherited from DOS days, where the third arg was used for... something. See its `ret` @0x100CCC05; `ret 0Ch" is purging three stack entries = 3 args.

madebr: You're right. I forgot about the `this` argument.

foxtacles: Made a PR for that earlier: https://github.com/stravant/isle/pull/1

stravant: Sounds good, merged.

stravant: For reference, I'm not exactly sure what the return types on the MXIOINFO functions are.

I thought they would exactly match the return types from the Windows mmio API, however there's at least one which clearly does not from the generated assembly, so they seem to be wrappers with slightly different types?

I'll figure it out when I actually decomp them but they don't seem like very important functions to tackle until we're closer to loading.

foxtacles (approved):

LEGO1/mxdsfile.cpp
```diff
@@ -1,6 +1,130 @@
...
+    sprintf(tempBuffer, "Wrong SI file version. %d.%d expected.", SI_MAJOR_VERSION, SI_MINOR_VERSION);
+    MessageBoxA(NULL, tempBuffer, NULL, MB_ICONERROR);
+    return -1;
+  }
+}
+
+// OFFSET: LEGO1 0x100cc7b0
+long MxDSFile::Seek(long lOffset, int iOrigin)
+{
+  return (m_position = m_io.Seek(lOffset, iOrigin)) == 0xFFFFFFFF ? -1 : 0;
```
itsmattkc, line 102: Strictly the `0xFFFFFFFF` may just be `-1`, based on the documented error code of `mmioSeek`: https://learn.microsoft.com/en-us/windows/win32/api/mmiscapi/nf-mmiscapi-mmioseek

PR #53, opened 2023-06-27
lego1: Two more operators for MxString

disinvite:
Followed the hint from @madebr in #31 that the next function in MxString was operator+. The one after that is operator+= and both are at 100%.

~~I thought maybe these new changes would magically fix the swapped regs in the `cmp` instruction from operator= at `0x100ae510`, but no luck.~~

Replaced malloc/free with new/delete, which had the side effect of fixing the swapped regs in operator=

### DIFF
--- a/LEGO1/mxstring.cpp
+++ b/LEGO1/mxstring.cpp
@@ -6,7 +6,7 @@
 MxString::MxString()
 {
   // Set string to one char in length and set that char to null terminator
-  this->m_data = (char *)malloc(1);
+  this->m_data = new char[1];
   this->m_data[0] = 0;
   this->m_length = 0;
 }
@@ -15,7 +15,7 @@ MxString::MxString()
 MxString::MxString(const MxString &str)
 {
   this->m_length = str.m_length;
-  this->m_data = (char *)malloc(this->m_length + 1);
+  this->m_data = new char[this->m_length + 1];
   strcpy(this->m_data, str.m_data);
 }
 
@@ -24,10 +24,10 @@ MxString::MxString(const char *str)
 {
   if (str) {
     this->m_length = strlen(str);
-    this->m_data = (char *)malloc(this->m_length + 1);
+    this->m_data = new char[this->m_length + 1];
     strcpy(this->m_data, str);
   } else {
-    this->m_data = (char *)malloc(1);
+    this->m_data = new char[1];
     this->m_data[0] = 0;
     this->m_length = 0;
   }
@@ -36,7 +36,7 @@ MxString::MxString(const char *str)
 // OFFSET: LEGO1 0x100ae420
 MxString::~MxString()
 {
-  free(this->m_data);
+  delete[] this->m_data;
 }
 
 // OFFSET: LEGO1 0x100ae490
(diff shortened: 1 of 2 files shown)


LEGO1/mxstring.cpp
```diff
@@ -79,3 +79,37 @@ const MxString &MxString::operator=(const char *param)
 
   return *this;
 }
+
+// Return type is intentionally just MxString, not MxString&.
+// This forces MSVC to add $ReturnUdt$ to the stack for 100% match.
```
madebr, line 84: You need to return a copy, not a reference.
Otherwise you'd return a reference to an already-destroyed `MxString` on the stack.

LEGO1/mxstring.h
```diff
@@ -14,7 +14,9 @@ class MxString : public MxCore
   MxString(const char *);
   void ToUpperCase();
   void ToLowerCase();
-  const MxString &operator=(MxString *);
+  const MxString&  operator=(MxString *);
+  const MxString   operator+(const char *);
+  const MxString& operator+=(const char *);
```
madebr, line 19: The `const ` is not needed for any of the return types.
It needlessly blocks modification of its return type.

LEGO1/mxstring.cpp
disinvite: Does the final call to the MxString constructor at `0x100ae2a0` make that copy? (And does this actually leak memory for the tmp.m_data buffer used inside the function?)

LEGO1/mxstring.h
disinvite: Fair point. I was kinda blindly following the use of `const` in the DLL exported operator

LEGO1/mxstring.cpp
madebr: The caller allocates memory and passes this as an argument.
At the end, this method constructs a `MxString` in this memory.

A `MxString&` return type would just have returned a `MxString *` pointer, and not needed an extra argument.
Low-level, c++ references (`&` suffix) are the same as pointers.

LEGO1/mxstring.cpp
```diff
@@ -79,3 +79,37 @@ const MxString &MxString::operator=(const char *param)
...
+// OFFSET: LEGO1 0x100ae580
+MxString MxString::operator+(const char *str)
+{
+  // MxString constructor calls malloc(1) for m_data, so free that first
+  MxString tmp;
+  free(tmp.m_data);
+
+  tmp.m_length = strlen(str) + this->m_length;
+  tmp.m_data = (char *)malloc(tmp.m_length + 1);
```
madebr, line 93: I think all allocations and deallocations of `m_data` in `MxString` should use `operator new[]` and `operator delete[]` because this is calling the same allocation function when creating objects.

disinvite: Just changed that and it fixed the swapped registers in the other operator function

LEGO1/mxstring.cpp
```diff
@@ -36,7 +36,7 @@ MxString::MxString(const char *str)
 // OFFSET: LEGO1 0x100ae420
 MxString::~MxString()
 {
-  free(this->m_data);
+  delete this->m_data;
```
madebr, line 39:
```suggestion
  delete[] this->m_data;
```

LEGO1/mxstring.cpp
```diff
@@ -52,30 +52,63 @@ void MxString::ToLowerCase()
...
 }
 
-// TODO: this *mostly* matches, again weird with the comparison
 // OFFSET: LEGO1 0x100ae510
 const MxString &MxString::operator=(const char *param)
 {
   if (this->m_data != param)
   {
-    free(this->m_data);
+    delete this->m_data;
```
madebr, line 73:
```suggestion
    delete[] this->m_data;
```

LEGO1/mxstring.cpp
```diff
@@ -52,30 +52,63 @@ void MxString::ToLowerCase()
...
+// OFFSET: LEGO1 0x100ae690
+MxString& MxString::operator+=(const char *str)
+{
+  int newlen = this->m_length + strlen(str);
+
+  char *tmp = new char[newlen + 1];
+  strcpy(tmp, this->m_data);
+  strcpy(tmp + this->m_length, str);
+  
+  delete this->m_data;
```
madebr, line 109:
```suggestion
  delete[] this->m_data;
```

disinvite: Changed all the deletes just now

madebr (approved):
100% similarity. Can't argue with that :)

foxtacles (approved):

PR #54, opened 2023-06-28
lego1: implement MXIOINFO

disinvite:
Implementation of MXIOINFO. We are close to 100% except for the usual register/comparison-order swapping.

**Notes:**
- The class uses the `MMIOINFO` struct but not the `mmio` API. Instead, we use the WINBASE API for file I/O (i.e. `_hread`, `_llseek` and so on.) This requires us to cast the `m_info.hmmio` member to either `HFILE` or `HMMIO` typedefs depending on the situation.
- The Open function uses a `movzx` instruction on the value of parameter `fdwOpen` before pushing to OpenFile from the kernel. You can force this to appear by casting to `MxU16`, so I've done that, odd as it seems.
- This entire file is honestly pretty ugly. The main reason is all the nested ifs; we are constrained by returning a result value from each function, but only at the very end instead of bailing out with a `return`. By far the worst offender is the do/while loop in the Descend function.
- There is a lot of repeated code here, namely the call to `_llseek` to reset `m_lDiskOffset` based on the current file position. Maybe some of these were inlined helper functions, but it's hard to tell.
- There are some resources on Microsoft's site that help explain some of this `MMIOINFO` stuff.
    - https://learn.microsoft.com/en-us/windows/win32/multimedia/buffered-services
    - https://learn.microsoft.com/en-us/windows/win32/api/mmiscapi/nf-mmiscapi-mmioadvance
    - https://learn.microsoft.com/en-us/windows/win32/api/mmiscapi/nf-mmiscapi-mmiodescend


ItzSwirlz:
I'm not an expert at this, but try to make sure that the functions aren't just calls to other things :)

foxtacles:
> * The Open function uses a `movzx` instruction on the value of parameter `fdwOpen` before pushing to OpenFile from the kernel. You can force this to appear by casting to `unsigned short`, but this disturbs the instructions for the rest of the file. To get the "best" overall match I decided to leave this out.

This seems pretty similar to what I've been seeing in `ISLE.EXE`, especially in `WndProc`, that small unrelated changes elsewhere strongly disturb large functions like that. I do believe casting to `unsigned short` is the correct thing based on the assembly and the rest of the code is also correct, but won't actually line up to 100% until...sometime in the future.

Not sure what the best thing to do would be, I guess we can leave it like that and revisit it later.

LEGO1/mxioinfo.cpp
```diff
@@ -3,7 +3,9 @@
 // OFFSET: LEGO1 0x100cc800
 MXIOINFO::MXIOINFO()
 {
-  memset(&m_info, 0, sizeof(MMIOINFO));
+  // This is not good practice, but shouldn't damage anything
+  // because MXIOINFO has no vtable.
+  memset(this, 0, sizeof(*this));
```
foxtacles, line 8: #59 adds the `DECOMP_STATIC_ASSERT` macro. If we use this to verify the size of `MXIOINFO` equals the `MMIOINFO` struct I think this is acceptable.

stravant:
>  I figured I might as well just pull out the members from the struct so we don't have m_info all over the place.

Do you think they didn't do this in the original MXIOINFO struct in the original source code? 

From the naming (We know it's called exactly MXIOINFO from the exports) and layout it's pretty clear that they did use an MMIOINFO struct in some way not have the members directly there like that.

If you want to avoid the overhead then you could consider making it `class MXIOINFO : private MMIOINFO`, but when I tried that I got a slightly less good match (mixed up register allocation) so I guessed that they did not do that in the original code and had a member variable MMIOINFO.

disinvite:
@foxtacles Good idea, I'll add the assert.

@stravant Is there something in the assembly that would tell for certain that they used a struct member and not all individual members?

I took the name MXIOINFO to mean "this is our version of MMIOINFO". If it were exactly the same as the base struct, they could have (and maybe should have) just used the mmio.* API.

The key difference is that they used the _lread, _lseek, _lwrite functions for all the file I/O. You could cast the file handle member to `HFILE` (int) everywhere or change the type as I did here. My thought was: as long as the types are different, we can just put the members directly in the class instead of using another struct.

disinvite:
@stravant Still playing with this... I switched back to an m_info struct, and, while not 100%, it is more stable even with the `movzx` cast. I may rebase this whole thing to put the struct member back in.

stravant:
> Is there something in the assembly that would tell for certain that they used a struct member and not all individual members?

If I had to guess it started out as them wanting to include RAII, closing the MMIOINFO in the destructor. However they then diverged from the base MMIOINFO functions for some reason from there.

Also I like to trust the programmer to be lazy. Writing `: public MMIOINFO` is easier than coping over the members.

disinvite:
It seems like we have similar match percentage between the following options:

- Individual members copied from `MMIOINFO` but using `HFILE` type for the file handle (i.e. the current version in the PR)
- Same as above but use an anonymous struct to encapsulate the values (similar to `MMIOINFO`)
- Extend `MMIOINFO` directly and cast file handle to `HFILE` where needed

So maybe it comes down to personal preference and code style at this point. If we extend `MMIOINFO` directly, we lose the `m_` prefix so the code gets a little confusing to read.

stravant:
I'm fine with you going ahead with this, though I would add a comment in the header letting people know that the layout comes from the windows class.

itsmattkc:
Hmm I'm not really sure about adding all the members, it feels very unlikely that this is what the original developers would have done and I don't love the idea of copying a Win32 struct rather than just using the original. What was the reason for not doing so, was it just not wanting to use `m_info` everywhere, or did it reduce accuracy?

disinvite:
Yeah, as I've thought about this more, I agree that we should just use `MMIOINFO`. My reason for copying the members in the first place was to drop `m_info` and reduce line length, but also because `hmmio` is used as `HFILE` (int) and so we also reduce the need to cast it everywhere by changing the type. But I suppose it's likely the devs  _did_ just cast it.

I'll fix that up and try to squash the commits. The accuracy should be consistent across the different approaches (see my comment above.)

disinvite:
This is ready for final review

PR #55, opened 2023-06-28
lego: Implement MxDSObject

foxtacles:
Everything matches to 100%, except two functions where different registers are used for the same instructions.

### DIFF
--- a/LEGO1/mxdsobject.cpp
+++ b/LEGO1/mxdsobject.cpp
@@ -7,7 +7,7 @@
 MxDSObject::MxDSObject()
 {
   this->m_unk0c = 0;
-  this->m_unk10 = 0;
+  this->m_unk10 = NULL;
   this->m_unk14 = 0;
   this->m_name = NULL;
   this->m_unk24 = -1;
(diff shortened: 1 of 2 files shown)


LEGO1/mxdsobject.cpp
```diff
@@ -7,30 +7,123 @@
...
+  this->CopyFrom(p_dsObject);
+  return *this;
+}
+
 // OFFSET: LEGO1 0x100bf8e0
 void MxDSObject::SetObjectName(const char *p_name)
 {
-  if (p_name != this->m_name)
+  if (p_name != this->m_objectName)
   {
```
madebr, line 51: Imho, only the curly braces of a method/function should be on a separate line.
```suggestion
  if (p_name != this->m_objectName) {
```

madebr (approved):

itsmattkc:
> Also note how we have an extreme accuracy regression in WndProc even though no code related to the function has changed in this PR, that really goes to show how bad the compiler's "butterfly effect" can be:

Heh ironically the big PR I merged dropped WndProc to 79.53 too, and merging this one actually ended up _restoring_ it to 96.96. Compilers are weird.

PR #58, opened 2023-06-29
lego1: match LegoOmni::GetInstance and deconstructor and add a few methods

MishaProductions:
- The m_unka0 field in LegoOmni is an MxDsAction, which allows the deconstructor to match
 - Implemented LegoOmni::GetInstance and Destroy instance
 - moves a few functions to their correct places


madebr (approved):

MishaProductions:
Ready for review

PR #59, opened 2023-06-29
Implement MxObjectFactory::{MxObjectFactory,Create} + match sizes of Mx.*Presenter objects created by it

madebr:
Matching of `MxObjectFactory::Create` is bad, because none of the objects it creates has actually been implemented yet.
Looks like the script is unable to match `MxObjectFactory::MxObjectFactory`.

This pr also introduces a little `decomp.h` header, which is aimed to contain macro's/functions/typedefs only useful for matching the OG binary.
I used a typedef macro for `DECOMP_ASSERT`, because anonymous namespaces didn't work.
```
Z:\home\maarten\projects\isle\LEGO1\mxobjectfactory.cpp(22) : error C2870: '<Unknown>' : a namespace definition must appear either at file scope or immediately within another namespace definition
```

### DIFF
--- a/LEGO1/decomp.h
+++ b/LEGO1/decomp.h
@@ -2,6 +2,7 @@
 #define DECOMP_H
 
 #define DECOMP_STATIC_ASSERT(V) namespace { typedef int foo[(V)?1:-1]; }
+#define DECOMP_SIZE_ASSERT(T, S) DECOMP_STATIC_ASSERT(sizeof(T) == S)
 
 typedef unsigned char undefined;
 typedef unsigned short undefined2;
--- a/LEGO1/decomp.h
+++ b/LEGO1/decomp.h
@@ -1,0 +1,10 @@
+#ifndef DECOMP_H
+#define DECOMP_H
+
+#define DECOMP_STATIC_ASSERT(V) namespace { typedef int foo[(V)?1:-1]; }
+
+typedef unsigned char undefined;
+typedef unsigned short undefined2;
+typedef unsigned int undefined4;
+
+#endif // DECOMP_H
--- a/LEGO1/legoinputmanager.h
+++ b/LEGO1/legoinputmanager.h
@@ -1,9 +1,8 @@
 #ifndef LEGOINPUTMANAGER_H
 #define LEGOINPUTMANAGER_H
 
-#include "mxpresenter.h"
-
 #include "decomp.h"
+#include "mxpresenter.h"
 
 enum NotificationId
 {
--- a/LEGO1/legoinputmanager.h
+++ b/LEGO1/legoinputmanager.h
@@ -29,7 +29,7 @@ public:
 
   virtual long Tickle() override; // vtable+0x8
 
-  undefined m_pad00[0x15c];
+  undefined m_pad40[0x15c];
   int m_joystickIndex;
   undefined m_pad200[0x194];
   MxBool m_useJoystick;
--- a/LEGO1/legoinputmanager.h
+++ b/LEGO1/legoinputmanager.h
@@ -3,6 +3,8 @@
 
 #include "mxpresenter.h"
 
+#include "decomp.h"
+
 enum NotificationId
 {
   NONE = 0,
@@ -27,9 +29,9 @@ public:
 
   virtual long Tickle() override; // vtable+0x8
 
-  char m_pad00[0x19C];
+  undefined m_pad00[0x15c];
   int m_joystickIndex;
-  char m_pad200[0x194];
+  undefined m_pad200[0x194];
   MxBool m_useJoystick;
 };
 
--- a/LEGO1/legoinputmanager.h
+++ b/LEGO1/legoinputmanager.h
@@ -1,9 +1,9 @@
 #ifndef LEGOINPUTMANAGER_H
 #define LEGOINPUTMANAGER_H
 
-#include "mxcore.h"
+#include "mxpresenter.h"
 
-__declspec(dllexport) enum NotificationId
+enum NotificationId
 {
   NONE = 0,
   KEYDOWN = 7,
@@ -15,7 +15,7 @@ __declspec(dllexport) enum NotificationId
 
 // VTABLE 0x100d8760
 // SIZE 0x338
-class LegoInputManager
+class LegoInputManager : public MxPresenter
 {
 public:
   LegoInputManager();
--- a/LEGO1/mxatomid.h
+++ b/LEGO1/mxatomid.h
@@ -1,8 +1,14 @@
 #ifndef MXATOMID_H
 #define MXATOMID_H
 
+#include "mxtypes.h"
+
 enum LookupMode
 {
+    LookupMode_Exact = 0,
+    LookupMode_LowerCase = 1,
+    LookupMode_UpperCase = 2,
+    LookupMode_LowerCase2 = 3
 };
 
 class MxAtomId
@@ -15,6 +21,11 @@ public:
   MxAtomId()
   {
     this->m_internal = 0;
+  }
+
+  inline MxBool operator ==(const MxAtomId &other) const
+  {
+    return this->m_internal == other.m_internal;
   }
 
 private:
--- a/LEGO1/mxobjectfactory.h
+++ b/LEGO1/mxobjectfactory.h
@@ -2,11 +2,33 @@
 #define MXOBJECTFACTORY_H
 
 #include "mxcore.h"
+#include "mxatomid.h"
+
+#define FOR_MXOBJECTFACTORY_OBJECTS(X)  \
+  X(MxPresenter)                        \
+  X(MxCompositePresenter)               \
+  X(MxVideoPresenter)                   \
+  X(MxFlcPresenter)                     \
+  X(MxSmkPresenter)                     \
+  X(MxStillPresenter)                   \
+  X(MxWavePresenter)                    \
+  X(MxMIDIPresenter)                    \
+  X(MxEventPresenter)                   \
+  X(MxLoopingFlcPresenter)              \
+  X(MxLoopingSmkPresenter)              \
+  X(MxLoopingMIDIPresenter)
 
 // VTABLE 0x100dc220
 class MxObjectFactory : public MxCore
 {
-
+public:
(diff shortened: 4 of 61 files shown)


LEGO1/mxatomid.h
```diff
@@ -17,6 +23,11 @@ class MxAtomId
     this->m_internal = 0;
   }
 
+  inline MxBool operator ==(const MxAtomId &other) const
+  {
+      return this->m_internal == other.m_internal;
```
foxtacles, line 28: The 4-space indents are a bit incosistent. If we decide to use 4-space indents we should probably make it a rule and adjust everything else at the same time (I don't really have an opinion on the indents, all I care about would be consistency)

madebr: I forgot about the 2-space indentation.
https://github.com/isledecomp/isle/pull/46 adds an .editorconfig that should fix this (at least for my ide)

LEGO1/mxatomid.h
```diff
@@ -1,8 +1,14 @@
...
 
+#include "mxtypes.h"
+
 enum LookupMode
 {
+    Exact = 0,
+    LowerCase = 1,
+    UpperCase = 2,
+    LowerCase2 = 3
 };
```
madebr, line 12: MSVC4.2 does not supoprt e.g. `LookupMode::Exact`.
So perhaps I should namespace its member values even further.
e.g. `LookupMode::LookupMode_Exact`

tamlin-mike:
Feel free to add my version of CT_ASSERT (tweaked to work for VC42) to decomp.h.
Since the sizes of all objects this method creates are know (by this method), it could be useful to pad them to correct size and use that macro to verify the size, to allow proper "blah = new Whatever;" in this method (I believe effectively all types this method creates have a default ctor).

So even if the types are stubs, they are properly allocated and should allow for a zero diff of this method without having all the types it creates actually implemented. Well, except for the inlined ctors - those obviously need implementation.

Just a thought.

madebr:
> Feel free to add my version of CT_ASSERT (tweaked to work for VC42) to decomp.h. Since the sizes of all objects this method creates are know (by this method), it could be useful to pad them to correct size and use that macro to verify the size, to allow proper "blah = new Whatever;" in this method (I believe effectively all types this method creates have a default ctor).

Thanks, I thought the anonymous namespace approach did not work, but apparently I was wrong.
I named the macro `DECOMP_STATIC_ASSERT` because the macro must only be valid when building the decompiled binary, not when building for a different arch.

> 
> So even if the types are stubs, they are properly allocated and should allow for a zero diff of this method without having all the types it creates actually implemented. Well, except for the inlined ctors - those obviously need implementation.

I added an initial implementations for every object, allocated by `MxObjectFactory::Create`.
Looks like binary matching is not good. But imo that's another issue.

foxtacles (commented):
Looks good to me. We probably want to use `DECOMP_ASSERT` everywhere we possibly can to avoid accidentally introducing bugs (has happened a few times already)

LEGO1/decomp.h
```diff
@@ -0,0 +1,10 @@
+#ifndef DECOMP_H
+#define DECOMP_H
+
+#define DECOMP_STATIC_ASSERT(V) namespace { typedef int foo[(V)?1:-1]; }
+
+typedef unsigned char undefined;
+typedef unsigned short undefined2;
+typedef unsigned int undefined4;
```
itsmattkc, line 8: I'm not sure these are necessary, couldn't we just use `MxU32`, `MxU16`, etc. from `mxtypes.h`?

madebr: The idea is to use undefinedX where we don't know what type something has. It can be a value, or a pointer.
`MxU32` are used when you are 100% sure it's a 32-bit integer.

madebr: The final goal is of this project is to only have `DECOMP_STATIC_ASSERT`'s at the end.

madebr: You can see it as self-documenting code.

itsmattkc: Sure, but if it's called `m_unk50`, doesn't that already indicate that we don't know what it is?

madebr: Yes, and these members are also used in functions.
A simple grep for `undefined` will show what is currenty unknown. MxUxx is fuzzy.

foxtacles: Generally I think it's a good idea. Just like with other measures too though, I think it's better - in case we decide for this - that we don't gradually introduce it but ideally apply it everywhere in the code base at once and establish its usage as a "style" rule. 

Maybe the same should be done with the types from `mxtypes.h` - currently it's used in *some* places, but not everywhere it probably should be. We could also define a rule where it should be used. I always find consistency, even in those small things, to be extremely helpful.

With regards to the naming of variables: generally there are always two possible "unknown" components I suppose, the type and the "purpose" (designated by its name). It's possible that both the type and the purpose are unknown. Sometimes however we know the type, but not its purpose. So I can imagine 4 cases like this:

```
undefined4 m_unk4;
undefined4 m_length; // probably relatively rare
MxU32 m_unk4;
MxU32 m_length;
```

By introducing something like the `undefined*` types we can document that we don't know which type it is (yet). I think that would be good

LEGO1/legoinputmanager.h
```diff
@@ -27,9 +29,9 @@ class LegoInputManager
 
   virtual long Tickle() override; // vtable+0x8
 
-  char m_pad00[0x19C];
+  undefined m_pad00[0x15c];
```
madebr, line 32:
```suggestion
  undefined m_pad40[0x15c];
```

tamlin-mike: m_20 and m_40 are MxCriticalSection

madebr: m_20 is part of `MxPresenter`. I didn't verify m_40.

LEGO1/decomp.h
itsmattkc: Yeah okay, seeing it in use (e.g. writing `undefined[0x16c]` to pad a struct instead of `char[0x16c]`) has convinced me it probably is helpful.

I agree it should probably be made a rule, along with using the `mxtypes`. The exports have a lot of `unsigned char`s that almost certainly resolve to MxBool but don't yet in our code.

itsmattkc:
I've added a `DECOMP_SIZE_ASSERT` macro to help ease writing (and reading) asserts for class sizes because there will probably be a _lot_ of them throughout the code (presumably one for every known class). I also think the sizes should be listed as hex numbers rather than decimals because hex is used for all of the offsets/addresses, and it'll be cumbersome to convert between them when trying to fill out a class' members. If you could update the remaining code to fit those two criteria, I think this will be ready to merge.

tamlin-mike:
> I also think the sizes should be listed as hex number

Definitely! Nothing but hex makes any sense whatsoever.

LEGO1/mxpresenter.h
```diff
@@ -5,65 +5,64 @@
...
   __declspec(dllexport) virtual void DoneTickle(); // vtable+0x2c
-  __declspec(dllexport) void Init();
   __declspec(dllexport) virtual void ParseExtra(); // vtable+0x30
 public:
+  virtual undefined4 VTable0x34(); // vtable+0x34
+  virtual void InitVirtual(); // vtable+0x38
   __declspec(dllexport) virtual long StartAction(MxStreamController *, MxDSAction *); // vtable+0x3c
   __declspec(dllexport) virtual void EndAction(); // vtable+0x40
+  virtual void VTable0x44(undefined4 param); // vtable+0x44
+  virtual unsigned char MxPresenter::VTable0x48(unsigned char param); // vtable+0x48
```
itsmattkc, line 50: Unnecessary `MxPresenter::`

LEGO1/mxobjectfactory.h
```diff
@@ -2,11 +2,33 @@
...
+  X(MxLoopingFlcPresenter)              \
+  X(MxLoopingSmkPresenter)              \
+  X(MxLoopingMIDIPresenter)
 
 // VTABLE 0x100dc220
 class MxObjectFactory : public MxCore
 {
-
+public:
+  MxObjectFactor();
```
itsmattkc, line 25: Typo in `MxObjectFactor()`

LEGO1/mxobjectfactory.h
```diff
@@ -2,11 +2,33 @@
...
+  virtual MxCore *Create(const char *name); // vtable 0x14
+  virtual void vtable18(void *); // vtable 0x18
+private:
+#define X(V) MxAtomId m_id##V;
+  FOR_MXOBJECTFACTORY_OBJECTS(X)
```
tamlin-mike, line 30: Tip: After the type you might want to add compile-time assert for `offsetof` one or more of these fields. Better safe than sorry.

LEGO1/mxpresenter.cpp
```diff
@@ -46,52 +49,57 @@ long MxPresenter::StartAction(MxStreamController *, MxDSAction *)
...
 }
 
-// OFFSET: LEGO1 0x1000be40 STUB
+// OFFSET: LEGO1 0x1000be40
 void MxPresenter::VTable0x18()
 {
-  // TODO
+  ParseExtra();
+  m_unk0xc |= 1 << (unsigned char)m_unk0x8;
+  m_unk0x8 = 2;
```
tamlin-mike, line 74: From the usage of these two fields it seems they are something like:
(u?)int m_8_tickleState.
DWORD m_C_tickleBits; // bitfield

Also, while I haven't checked with the compiler, does it perhaps do the cast to uchar automatically when given an (u)int? After all, it makes no sense to shift more than 31 bits, and `lsh` is AFAIK also only accepting an 8-bit shift count. Might be worth checking, to possibly get rid of all the manual shiftcount casts.

itsmattkc:
I fixed the two minor issues so we can merge in the size assert. I'd say any further notes can be implemented as future PRs.

LEGO1/mxpresenter.cpp
madebr: It is also possible the 1 should be unsigned. 1u. I haven't investigated deeply.

PR #61, opened 2023-06-29
Yet another CMake implementation

itsmattkc:
Yes, I'm finally officially committing to switching over to CMake. I wasn't originally planning to until the end, but the MDP/MAK formats are proving difficult to work with collaboratively, and I'm now relatively reassured that accuracy can still be attained this way.

This is heavily influenced by #50. There was nothing particularly wrong with that PR, I just felt like I wanted to fully understand what CMake was doing here and what was necessary to make it cooperate, and the best way would be to write something from scratch. But I used that PR heavily as a resource, so thanks @madebr for putting it together.

According to `reccmp`, this produces an identical LEGO1 to the MSVC workflow. Unfortunately the infamous operands in `WinMain` flip around again, so ISLE is slightly less accurate, but there's a decent chance something else will flip them back later on. Otherwise everything seems fine, and as such, I think we can pretty comfortably call this as accurate as the old workflow.

I'm using CMake 2.8 in an attempt to improve backwards compatibility. This version works as far back as 2000/XP. Unfortunately I don't think any version works on 9x, so anyone who wishes to build on those platforms will need to generate an NMake makefile in advance.

I didn't include as much code for making non-MSVC/non-Windows/non-vanilla workflows working. Not that we never will, just that it felt unnecessary for the time being. Later on this stuff may be added anyway.


CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
...
+
+  set(CMAKE_CXX_FLAGS "/W3 /GX /D \"WIN32\" /D \"_WINDOWS\"")
+  set(CMAKE_CXX_FLAGS_DEBUG "/Gm /Zi /Od /D \"_DEBUG\"")
+  set(CMAKE_CXX_FLAGS_RELEASE "/O2 /D \"NDEBUG\"")
+  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "/Zi /O2 /D \"NDEBUG\"")
+  set(CMAKE_CXX_FLAGS_MINSIZEREL "/Os /D \"NDEBUG\"")
+
+  set(CMAKE_EXE_LINKER_FLAGS "/machine:I386")
+  set(CMAKE_EXE_LINKER_FLAGS_DEBUG "/incremental:yes /debug")
+  set(CMAKE_EXE_LINKER_FLAGS_RELEASE "/incremental:no")
```
madebr, line 203: The `/DEBUG` is for a pdb library, which is needed for the `reccmp.py` script.
See https://learn.microsoft.com/en-us/cpp/build/reference/debug-generate-debug-info

CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
...
+  set(CMAKE_EXE_LINKER_FLAGS_RELWITHDEBINFO "/incremental:no /debug")
```
madebr, line 204: Great idea to make `RelWIthDebInfo` equal to `Release` apart from this `/debug` flag.

CMakeLists.txt
```diff
@@ -0,0 +1,212 @@
+cmake_minimum_required(VERSION 2.8 FATAL_ERROR)
```
madebr, line 1: Otherwise, this will emit warnings on current CMake versions.
```suggestion
cmake_minimum_required(VERSION 2.8...3.5 FATAL_ERROR)
```

theR4K:
But, how windows users will build this?

theR4K:
As described on #47, .lib and .obj order influencing resulted binary.
At least I can suggest return kernel.lib dependency.

madebr:
> But, how windows users will build this?

Run `vcvars32.bat`, or set `INCLUDE`, `LIB` and `PATH`.
And then run `cmake -S . -B build -G "NMake Makefiles"`

> As described on #47, .lib and .obj order influencing resulted binary. At least I can suggest return kernel32.lib dependency.

We have still control over this.

itsmattkc:
> But, how windows users will build this?

...By using CMake? It's available on Windows and produces NMAKE files compatible with MSVC420.

theR4K:
> > But, how windows users will build this?
> 
> Run `vcvars32.bat`, or set `INCLUDE`, `LIB` and `PATH`. And then run `cmake -S . -B build -G "NMake Makefiles"`
> 
> > As described on #47, .lib and .obj order influencing resulted binary. At least I can suggest return kernel32.lib dependency.
> 
> We have still control over this.

yeh, by installing cmake

theR4K:
don't get me wrong I just don't see any reasons for me to installing cmake...

itsmattkc:
I mean you don't have to "install" it, you can download the executables in a zip that you can just extract too.

theR4K:
it's also way to "install" things. Almost all installators it's noting but self extracting archive(kind of).
And also it's possible to install cmake using visual studio installer. But again, I don't want to, and don't need to.
Cmake it's linux stuff not windows...

itsmattkc:
I'm not sure how you can say it's "Linux stuff" when it generates NMAKE makefiles and is available from the Visual Studio installer (in fact, I think it's even included by default with the Visual C++ toolkit now). Pretty much any build system would be something you'd have to "install", so I don't really see the concern.

theR4K:
nope, they not included
https://learn.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=msvc-170

itsmattkc:
That page literally says:

> C++ CMake tools for Windows is installed as part of the Desktop development with C++

which is what I was referring to.

theR4K:
it's separate component, not "part of"
image

itsmattkc:
Sure, it's an individual component, but it's included by default with the "Desktop development with C++" workload. The page you linked yourself says that.

The Windows 10 SDK is also an "individual component". That doesn't make it Linux stuff.

theR4K:
SDK is optional because it's specific version of sdk. You will always have lasted SDK installed, but also you can install previous versions for compatibility.

theR4K:
I don't want to argue about that. I'm only mentioned that this change is not friendly for windows users.

itsmattkc:
> Optional means "not default"

Optional means you don't _need_ it, not that they don't _recommend_ it or that it isn't _enabled by default_:

https://github.com/isledecomp/isle/assets/34096995/c2b04126-d5bd-467b-9856-290056612b16

> I don't want to argue about that. I'm only mentioned that this change is not friendly for windows users.

I still haven't seen why this is the case. Like I said, it literally comes included by default when you install Visual C++.

PR #62, opened 2023-06-29
Faster reccmp.py on linux

madebr:
I noticed all paths in the output of `cvdump.exe` started with `.\`.
This pr proposes to do a naive string append + replace instead of executing winepath each time.

For my machine, this reduces the time of `reccmp.py` from ~80s to ~2s.

While looking for this bug, I added logging support which might be useful in the future.

Last, I sneaked in a little fix for those people developing on big-endian machines.

### DIFF
--- a/tools/reccmp/reccmp.py
+++ b/tools/reccmp/reccmp.py
@@ -209,9 +209,12 @@ class SymInfo:
     for fn in self.lines:
       # Sometimes a PDB is compiled with a relative path while we always have
       # an absolute path. Therefore we must
-      if os.path.samefile(fn, filename):
-        filename = fn
-        break
+      try:
+        if os.path.samefile(fn, filename):
+          filename = fn
+          break
+      except FileNotFoundError as e:
+        continue
 
     if filename in self.lines and line in self.lines[fn]:
       addr = self.lines[fn][line]
--- a/tools/reccmp/reccmp.py
+++ b/tools/reccmp/reccmp.py
@@ -6,6 +6,7 @@ from capstone import *
 import difflib
 import struct
 import subprocess
+import logging
 import os
 import sys
 import colorama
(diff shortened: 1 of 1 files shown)


itsmattkc:
> I noticed all paths in the output of cvdump.exe started with .\.
> This pr proposes to do a naive string append + replace instead of executing winepath each time.

I mentioned this in the Matrix, but the reason why it still calls `winedump` for those is because sometimes (notably when building with MSDEV, and seemingly with the new CMake system), absolute paths are baked into the PDB rather than relative paths, which will break this trick.

It's not yet clear why NMAKE produces relative paths or how/if that can be replicated elsewhere. Arguably, absolute paths are actually preferable because it means `reccmp` doesn't have to be run from a specific location (the repository root, which all source files are relative too), but then the absolute paths require `winepath` to transform them each time.

theR4K:
> > I noticed all paths in the output of cvdump.exe started with ..
> > This pr proposes to do a naive string append + replace instead of executing winepath each time.
> 
> I mentioned this in the Matrix, but the reason why it still calls `winedump` for those is because sometimes (notably when building with MSDEV, and seemingly with the new CMake system), absolute paths are baked into the PDB rather than relative paths, which will break this trick.
> 
> It's not yet clear why NMAKE produces relative paths or how/if that can be replicated elsewhere. Arguably, absolute paths are actually preferable because it means `reccmp` doesn't have to be run from a specific location (the repository root, which all source files are relative too), but then the absolute paths require `winepath` to transform them each time.

yeah, msdev uses absolute path for sources, and compile all of them in one batch. Actual command looks like this
```
/nologo /MT /W3 /GX /Zi /O2 /D "WIN32" /D "NDEBUG" /D "_WINDOWS" /Fp"LEGO1\Release/LEGO1.pch" /YX /Fo"LEGO1\Release/" /Fd"LEGO1\Release/" /c 
"C:\repo\isle\LEGO1\mxwavepresenter.cpp"
"C:\repo\isle\LEGO1\mxdsfile.cpp"
...
```
It's easy to implement behavior like this with premake\nmake2 module.  In this case we will always have the absolute path in pdb and as side effect this allows to execute nmake from any folder.

itsmattkc:
Great solution to the issue I mentioned above. Thanks!

PR #63, opened 2023-06-30
Get Isle building on MinGW32

madebr:
This pr provides a non-invasive alternative for https://github.com/isledecomp/isle/commit/d05393f883c4e4466f0856a8186b57d67bd56883

### DIFF
--- a/LEGO1/legoinputmanager.h
+++ b/LEGO1/legoinputmanager.h
@@ -27,7 +27,7 @@ public:
 
   virtual long Tickle() override; // vtable+0x8
 
-  char m_pad00[0x19C];
+  char m_pad00[0x15c];
   int m_joystickIndex;
   char m_pad200[0x194];
   MxBool m_useJoystick;
--- a/LEGO1/legoinputmanager.h
+++ b/LEGO1/legoinputmanager.h
@@ -1,7 +1,7 @@
 #ifndef LEGOINPUTMANAGER_H
 #define LEGOINPUTMANAGER_H
 
-#include "mxcore.h"
+#include "mxpresenter.h"
 
 __declspec(dllexport) enum NotificationId
 {
@@ -15,7 +15,7 @@ __declspec(dllexport) enum NotificationId
 
 // VTABLE 0x100d8760
 // SIZE 0x338
-class LegoInputManager
+class LegoInputManager : public MxPresenter
 {
 public:
   LegoInputManager();
--- a/LEGO1/legoinputmanager.h
+++ b/LEGO1/legoinputmanager.h
@@ -3,7 +3,7 @@
 
 #include "mxpresenter.h"
 
-__declspec(dllexport) enum NotificationId
+enum NotificationId
 {
   NONE = 0,
   KEYDOWN = 7,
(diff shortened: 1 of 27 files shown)


itsmattkc:
I'm not sure I love the idea of adding functions either, particularly not to the vtable. I'm aware they wouldn't appear in the compiled MSVC code, but my concern remains that it makes the code harder to manage when our primary focus is matching MSVC output. Sure these changes are _relatively_ minor, but it's a slippery slope I feel.

That being said, I've made a PR on your branch with an approach that I think is probably acceptable. Feel free to take a look and let me know what you think.

madebr:
I rebased it on current master.

LEGO1/mxstringvariable.cpp
```diff
@@ -1,14 +1,6 @@
...
+#include "mxstringvariable.h"
 #include "mxstring.h"
 
-//FIXME: Figure out what exactly this class is used for. It is used in LegoGameState::LegoGameState when loading the background color, and for loading the "fsmovie" variable
-// OFFSET: LEGO1 0x1003bec0
-MxStringVariable::~MxStringVariable()
-{
-
-  delete &m_string;
-  delete &m_name;
```
itsmattkc, line 10: Ahh I'm surprised I didn't pick up on this, thanks for fixing that

LEGO1/legoinputmanager.h
```diff
@@ -15,7 +15,7 @@ __declspec(dllexport) enum NotificationId
 
 // VTABLE 0x100d8760
 // SIZE 0x338
-class LegoInputManager
+class LegoInputManager : public MxPresenter
```
foxtacles, line 18: Can you make sure the `LegoInputManager` structure is still functional after the addition of this base class? I'm assuming you'd have to adjust the private member padding. `ISLE.EXE` depends on a few members in here.

madebr: Can you give https://github.com/isledecomp/isle/pull/59 another look?
It adds compile time assertions to quit a few classes.

If MxPresenter has size 64, I'll subtract 60 from the top member.

madebr: Oh, it's already merged.

madebr:
Closing this one.

itsmattkc:
Strange, it merged but the PR is still open?

LEGO1/legoinputmanager.h
foxtacles: I think it was an accident (or a glitch?) - but I think as long as we fix the structure of `LegoInputManager` it should be fine

foxtacles:
Looking at it again, I think the commit on `master` is not equivalent to the state of this PR. Maybe we should wipe that commit away, re-open the PR and make a proper merge?

itsmattkc:
Huh... very weird, yeah that sounds like a good idea.

LEGO1/legoinputmanager.h
madebr: I think the current size is of by 4 bytes.
When you add all members, you get `0x19c+4+0x194+4` = 824. Including the vtable, this becomes 828 bytes.
But the OG dll allocates 824 bytes.
I'm going to remove 4 bytes from the second array in #59.

itsmattkc: The vtable would be "hidden" because it's included as part of the structure MxCore, that's probably why it doesn't add up mathematically

madebr: Fixed it. `sizeof`  of #59 agrees.

foxtacles (approved):

PR #64, opened 2023-06-30
[WIP] LegoInputManager

ecumber:
I have started work on LegoInputManager, and boy what a hassle it is. Lego Island uses DirectX 5, and documentation for this version next to nonexistent on Google. Fortunately the SDK for this version isn't difficult to find, and it has some documentation (albeit in binary .HLP form, which cannot be opened in Windows 10 and later). I have made sense of the code but have not begun implementing it.
By the looks of it, modern windows headers do not have definitions required for implementing this class. It might be necessary to use the DirectX 5 SDK to get these definitions, so we might need to make a portable install in the same manner as the MSVC420 repository.

### DIFF
--- a/LEGO1/legogamestate.cpp
+++ b/LEGO1/legogamestate.cpp
@@ -1,4 +1,5 @@
 #include "legogamestate.h"
+#include "legoomni.h"
 
 // OFFSET: LEGO1 0x10039550
 LegoGameState::LegoGameState()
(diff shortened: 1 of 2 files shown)


itsmattkc:
For the most part, yeah DX5 will be necessary (we already use it on the CI and recommend it in the instructions). There are also no D3DRM headers in modern Windows, which the game relies on heavily. Its installer doesn't seem to have the same issues as MSVC420 though, and simply installing it makes it available for use with MSVC. A portable version may be nice for some setups, but it's not necessary in the same way.

ecumber:
I set up the CMake generator to look for the DX5 includes and libraries under the dx5sdk directory. This means you can install DX5 and copy the "sdk" folder into dx5sdk and have it compile. Is this acceptable? As of now it doesn't seem like there's any standardization on where to put DirectX files..

PR #65, opened 2023-06-30
CI steps

stonedDiscord:
break up the CI into 3 seperate jobs:
build, compare and upload

CI now runs on every branch to catch things that break the build early
upload only runs on a push to master
PNG generation is now handled on an ubuntu runner with librsvg instead of inkscape


itsmattkc:
I happened to make changes to the CI just as you submitted this (added a test for the DLL exports), so you may need to merge those in.

I'd also like to eliminate the SVG->PNG step at some point. The only reason it's necessary is because for some reason I can't embed SVGs from GitHub Releases, but it's definitely a poor solution because images (obviously) don't scale as well - particularly since it's mostly text.

stonedDiscord:
image
brilliant

PR #67, opened 2023-07-01
lego1: match MxBitmap::Read and vtable

foxtacles:
Matched and cleaned up `MxBitmap::Read` (and implemented `MxBitmap::vtable28`), also added the stubs for the other functions. Their signatures aren't necessarily 100% correct yet, just based on a cursory glance.

### DIFF
--- a/LEGO1/mxbitmap.cpp
+++ b/LEGO1/mxbitmap.cpp
@@ -14,15 +14,75 @@ MxBitmap::MxBitmap()
 // OFFSET: LEGO1 0x100bca10
 MxBitmap::~MxBitmap()
 {
-  if (this->m_info != NULL) {
+  if (this->m_info != NULL)
     delete m_info;
-  }
-  if (this->m_data != NULL) {
+  if (this->m_data != NULL)
     delete m_data;
-  }
-  if (this->m_palette != NULL) {
+  if (this->m_palette != NULL)
     delete m_palette;  
-  }
+}
+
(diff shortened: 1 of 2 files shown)


ItzSwirlz:
thank you! ill check in on this later today

LEGO1/mxbitmap.cpp
```diff
@@ -14,15 +14,75 @@ MxBitmap::MxBitmap()
 // OFFSET: LEGO1 0x100bca10
 MxBitmap::~MxBitmap()
 {
-  if (this->m_info != NULL) {
+  if (this->m_info != NULL)
```
ItzSwirlz, line 17: yeah we need a clang-format thing, i'll open a pr for it

ItzSwirlz:
this looks good! I'm going to open some PR's to your branch if you're okay with that, Sorry I forgot the vtables..

foxtacles:
> this looks good! I'm going to open some PR's to your branch if you're okay with that, Sorry I forgot the vtables..

Sure thing, and no problem! Maybe wait until this PR is merged into `master` and then make your PRs against that. I think that would be easier

LEGO1/mxbitmap.cpp
foxtacles: Should probably discuss this with @itsmattkc first if we want that and if yes, how it should be configured

itsmattkc:
Looks good to me. And yeah I think further PRs on this can be made to `master` here. Merging now.

PR #68, opened 2023-07-02
MXIOINFO Progress

stravant:
* MXIOINFO: 100%
* ~MXIOINFO: 100%
* Close: 100%
* Open: 100%
* SetBuffer: 80% (matching except register allocation)
* Something: 96% (matching except == comparison args swapped)
* Read: 59% (having trouble with the loop)


itsmattkc:
There's another MXIOINFO PR (#54) that appears to have implemented a few more functions and achieved 100% on `::Read`, though its `::Open` is less accurate. Perhaps you could PR some of your work onto there?

stravant:
Ah, didn't see, I'll just roll anything relevant into that PR.

PR #70, opened 2023-07-02
LEGO1: ConvertHSVToRGB is now effectively 100% matched

ktkaufman03:
`reccmp.py` currently gets confused by jump tables and only reports a ~90% match, but the actual code is matching now.

### DIFF
--- a/LEGO1/legoutil.cpp
+++ b/LEGO1/legoutil.cpp
@@ -9,10 +9,13 @@ void ConvertHSVToRGB(float h, float s, float v, float *r_out, float *b_out, floa
   double v9;
   double v12;
   double v13;
+
+  double s_dbl = s;
+
   if (s > 0.5f)
     calc = (1.0f - v) * s + v;
   else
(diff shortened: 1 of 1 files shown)


foxtacles (approved):

itsmattkc:
Yeah I've tried to solve the jump tables, but there doesn't seem to be any information in the PDB about where they're located in the code - seemingly according to it, it's just considered a part of the same function. There may be some heuristics that could be used to detect it, but that could also potentially introduce false positives in other places. The most reliable solution is probably to have an optional parameter in the `// OFFSET:` line to override the expected size of a function (assuming jump tables are always placed at the end).

Ironically, this seems to cause a regression in `MxDSFile::Open`, but it's probably a swapped operand/register. There seem to be a lot of external influences that semi-randomly break/unbreak things like that, I think we'll just have to learn to ignore them (or find a way to get reccmp to ignore them).

Anyways, looks good to me, will merge.

PR #72, opened 2023-07-02
lego: fix/match SetColorString

foxtacles:
Fixed the logic of and matched `SetColorString`. Also applied some of our style rules


itsmattkc:
Looks good to me, thanks!

PR #73, opened 2023-07-02
lego: add MxDSType enum, add it to all ctors, refactor `MxDSObject`

foxtacles:
`m_unk0c` is a type ID in `MxDSObject`, and it's set in every `MxDS*` ctor that belongs to the `MxDSObject` family. It was renamed to `m_type`. In some classes, this is all the ctor does so they match completely now.

The regressions in some of the ctors are meaningless since the bulk of the code in them is still missing. Setting the type like this is correct though. 

The regressions (operand swaps) in two `MxDSObject` functions appeared after adding `DECOMP_SIZE_ASSERT`, so they should be disregarded.

I've marked the dtors as `STUB`'s that were completely empty (but shouldn't be) so they don't appear in the report anymore. The dtors that indeed don't "do" anything (other than the implicit stuff) had their `// TODO` and `STUB` removed if it was present and match 100% now.

Also refactored the members of `MxDSObject` and their types to reflect our current state of knowledge about them. `m_unk26` didn't exist, it's just implicit padding.

### DIFF
--- a/LEGO1/mxdsobject.cpp
+++ b/LEGO1/mxdsobject.cpp
@@ -85,15 +85,15 @@ void MxDSObject::SetSourceName(const char *p_sourceName)
 }
 
 // OFFSET: LEGO1 0x100bf9c0
-int MxDSObject::unk14()
+undefined4 MxDSObject::unk14()
 {
   return 10;
 }
 
 // OFFSET: LEGO1 0x100bf9d0
-unsigned int MxDSObject::CalculateUnk08()
+MxU32 MxDSObject::CalculateUnk08()
 {
-  unsigned int unk08;
+  MxU32 unk08;
 
   if (this->m_sourceName)
     unk08 = strlen(this->m_sourceName) + 3;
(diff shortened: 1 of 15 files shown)


LEGO1/mxdsobject.cpp
```diff
@@ -113,16 +113,16 @@ unsigned int MxDSObject::CalculateUnk08()
 }
 
 // OFFSET: LEGO1 0x100bfa20
-void MxDSObject::Parse(char **p_source, MxU16 p_unk24)
+void MxDSObject::Parse(char **p_source, MxS16 p_unk24)
```
tamlin-mike, line 116: I think chances are pretty good the source is actually a buffer/stream class/struct, with (all inline) members to extract data of specific types and advance the pointer. I have a difficult time believing anyone would manually do "get data, cast to correct type, advance pointer by data-type size" over and over when it could be placed in a tiny neat wrapper.
Just an idea.

foxtacles: Good idea! I'd leave this for a separate PR though

PR #74, opened 2023-07-02
Simple integrated build process

stravant:
* Introduce "configure.py" and "build.py"

* Running "python configure.py && python build.py" will successfully build a clean clone of the project with no additional effort needed.

* The configure script will download Ninja and MSVC420 for you as part of configuring the project.  You can specify --modern-compiler when configuring to configure with a different toolchain if desired.

* The build script will download the original game EXE and DLL to `/original`  if you ask it for a status report / to compare a specific function.

* The repo also now includes a .vscode folder by default, which contains a tasks.json file including the build / configure tasks.

* This includes "build and compare function by cursor" task which allows a very effective workflow where you just hit Ctrl+Shift+B and get an updated diff of the function you were just editing.


CMakeLists.txt
```diff
@@ -220,4 +226,4 @@ if (MSVC)
   set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "/incremental:no")
   set(CMAKE_SHARED_LINKER_FLAGS_RELWITHDEBINFO "/incremental:no /debug")
   set(CMAKE_SHARED_LINKER_FLAGS_MINSIZEREL "/incremental:no")
-endif()
+endif()
```
madebr, line 229: Please keep final newline, here and in the other files.

CMakeLists.txt
```diff
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 2.8...3.5 FATAL_ERROR)
 
 project(isle CXX)
 
+link_directories("$ENV{LIB}")
+include_directories("$ENV{INCLUDE}")
+
 option(ISLE_BUILD_APP "Build ISLE.EXE application" ON)
```
madebr, line 8: Add this to a cmake toolchain.

And even so, when you rerun cmake without these environment variables set, nothing will be added to the include and library search directories.
```suggestion
project(isle CXX)

option(ISLE_BUILD_APP "Build ISLE.EXE application" ON)
```

configure.py
```diff
@@ -0,0 +1,75 @@
+
```
madebr, line 1:
```suggestion
#!/usr/bin/env python

```

build.py
```diff
@@ -0,0 +1,78 @@
+
```
madebr, line 1:
```suggestion
#!/usr/bin/env python

```

stravant:
Made it also download the original ISLE.EXE and LEGO1.EXE, and fixed some of the issues.

Still looking into toolchain file. I felt like it might be an unnecessarily fancy solution because we don't need the toolchain to do anything very complex in this case.

stravant:
Merged work on this from @madebr, the branch now uses a toolchain file.

itsmattkc:
I'm afraid I have to echo some comments I saw earlier in the Matrix room: I'm not sure this is that useful. I understand the point that it makes the compilation process more approachable and lowers the barrier to entry, but scripts like this are also something we have to maintain longterm, and will serve as a distraction from the primary goal of progressing the decompilation.

Sure it might enable a handful more people to compile it, but I doubt people who are knowledgeable and passionate enough to contribute significantly would be discouraged by a fairly straightforward CMake configuration. In fact, those people would likely prefer the manual process anyway, because they'd rather understand exactly what's happening and how to troubleshoot it in the future (I know this doesn't remove the manual process, but it does mean the extra maintenance required will probably have minimal benefit).

Sorry, I can see there was a lot of effort put into this, but I don't think it's likely that I'm going to merge it.

PR #75, opened 2023-07-03
cmake: fix default ninja + add clang-format target

madebr:
- Fix building isle with ninja without having to pass `-j1` to disable parallelism.
  This pr adds isle and lego1 to 2 separate job pools with 1 job available, effectively disabling parallelism per-target.
- ~add `compare-isle` and `compare-lego1` cmake targets, which run `reccmp.py` and open the report in the browser.
  For this to work, you need to configure cmake with:~
  ```sh
  cmake -S . -B build_ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -GNinja -DORIGINAL_ISLE="Z:/home/maarten/projects/isle/original/ISLE.EXE" -DORIGINAL_LEGO1="Z:/home/maarten/projects/isle/original/LEGO1.DLL"
  ```

- add `clang-format` target that runs clang-format on all sources in the `ISLE` and `LEGO1` directories.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -329,3 +329,27 @@ if (MSVC)
     set_property(TARGET lego1 PROPERTY JOB_POOL_LINK msvc_lego1)
   endif()
 endif()
+
+find_program(CLANGFORMAT_BIN NAMES clang-format)
+if(EXISTS "${CLANGFORMAT_BIN}")
+  execute_process(COMMAND "${CLANGFORMAT_BIN}" --version
+    OUTPUT_VARIABLE "CLANGFORMAT_VERSION_OUTPUT"
+    RESULT_VARIABLE "CLANGFORMAT_RESULT"
+  )
+  if(CLANGFORMAT_RESULT EQUAL 0 AND CLANGFORMAT_VERSION_OUTPUT MATCHES "version ([0-9\\.]+)")
+    set(CLANGFORMAT_VERSION "${CMAKE_MATCH_1}")
+    set(CLANGFORMAT_VERSION_REQUIRED "17.0")
+    message(DEBUG "Found clang-format version ${CLANGFORMAT_VERSION} (needs ${CLANGFORMAT_VERSION_REQUIRED}")
+    if(CLANGFORMAT_VERSION VERSION_GREATER_EQUAL "${CLANGFORMAT_VERSION_REQUIRED}")
+      file(GLOB_RECURSE isle_sources
+        "${PROJECT_SOURCE_DIR}/ISLE/*.cpp"
+        "${PROJECT_SOURCE_DIR}/ISLE/*.h"
+        "${PROJECT_SOURCE_DIR}/LEGO1/*.cpp"
+        "${PROJECT_SOURCE_DIR}/LEGO1/*.h"
+      )
+      string(REPLACE ";" "\n" isle_sources_lines "${isle_sources}")
+      file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/isle_sources.txt" "${isle_sources_lines}\n")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.8...3.5 FATAL_ERROR)
 
-project(isle CXX)
+project(isle CXX RC)
```
itsmattkc, line 3: I can't find any evidence that `RC` is a valid (or necessary) language to declare:

image
(https://cmake.org/cmake/help/latest/command/project.html)

madebr: I added it because no matter how many times I execute `cmake --build .`, `ISLE.EXE` keeps getting rebuilt. But it didn't help and I forgot to remove it.

[Looks like it is an "internal" language](https://github.com/Kitware/CMake/blob/20f07467a740e5eb407689b8c2072dc63763f806/Modules/Platform/Windows-MSVC.cmake#L510)

CMakeLists.txt
```diff
@@ -224,4 +224,38 @@ if (MSVC)
   set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "/incremental:no")
   set(CMAKE_SHARED_LINKER_FLAGS_RELWITHDEBINFO "/incremental:no /debug")
   set(CMAKE_SHARED_LINKER_FLAGS_MINSIZEREL "/incremental:no")
+
+  # Older MSVC versions don't support building in parallel.
+  # Force non-parallel builds of isle and lego1 by putting them in a pool with 1 available job.
+  if(CMAKE_CXX_COMPILER_ID VERSION_LESS 12)
+    set_propertY(GLOBAL PROPERTY JOB_POOLS "msvc_lego=1;msvc_lego1=1")
```
itsmattkc, line 231: Minor typo (capitalized `Y`)

madebr: The capital is for correct pronunciation :sweat_smile:

madebr: This is fixed now.

itsmattkc:
Rather than a separate target, do you think it would be possible to have the comparison as an automatic build step following the compilation? And rather than open the HTML, just print normally so people can compile/compare in the same command?

madebr:
This is possible by doing the comparison [in a `POST_BUILD` event](https://cmake.org/cmake/help/latest/command/add_custom_command.html#build-events).
I didn't go that way because then, the comparison will only be executed when the target needs to be rebuilt.

How about combining these 2?
- add a post build event, printing the comparison to the console, and
- add a target that manually runs the comparison.

About `reccmp.py`, I think it would be useful to have it produce both a html and cli report in the same run.
Then a dev only needs to run e.g. `ninja compare` once and read the cli, refresh his browser.

I'll remove the automatic opening in the browser.

madebr:
As you requested, I added the comparison as a `POST_BUILD` event.
But I'm not a huge fan of it: every rebuild of the exe/dll will trigger a comparison.

Unrelated, this pr also adds a `clang-format` target that runs clang-format on all cpp/h sources in the `ISLE`/`LEGO1` directories.

itsmattkc:
> But I'm not a huge fan of it: every rebuild of the exe/dll will trigger a comparison.

Perhaps I'm not understanding, how does this differ from your original implementation?

My thought process was that a lot of the time people are compiling + comparing a function over and over to match asm, so it would make sense for the compare to happen as an automatic (albeit optional) build step so that people can just run build and get the comparison in one step. Was this PR designed for a different purpose?

madebr:
I removed the addition of the `compare-isle` and `compare-lego1` targets.
I think the developers have found a way that works for them.
If somebody want better integration in cmake, feel free to ask.

PR #76, opened 2023-07-03
[lego] Implement LegoGameState::SetSavepath

MishaProductions:


LEGO1/legogamestate.cpp
```diff
@@ -39,7 +40,19 @@ void LegoGameState::SerializeScoreHistory(MxS16 p)
 }
 
 // OFFSET: LEGO1 0x10039f00
-void LegoGameState::SetSavePath(char *p)
+void LegoGameState::SetSavePath(char *p_SavePath)
```
foxtacles, line 43: for parameters and members the style guide is asking for `camelCase`, so it should be `p_savePath`, `m_savePath` etc.

PR #77, opened 2023-07-04
LegoStream and implementations 100% match

stravant:
* Add LegoStream and it's implementations, LegoFileStream and LegoMemoryStream. The names are ours, we don't know their original naming.

* These expose a Read/Write/Seek/Tell interface for reading and writing data, either from a file on disk or memory buffer.

* 100% match all functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,7 @@ add_library(lego1 SHARED
   LEGO1/mxentity.cpp
   LEGO1/mxeventmanager.cpp
   LEGO1/mxeventpresenter.cpp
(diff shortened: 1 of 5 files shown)


itsmattkc:
Seems good, can you add comments pointing to each class' vtable (`// VTABLE 0x`) and add checks for each class' size (`DECOMP_SIZE_ASSERT`)?

stravant:
Updated with VTABLES / size asserts.

Noted on the size asserts that these aren't entirely certain until we've 100% matched one of the functions with them as a local variable since they're only created on the stack.

PR #78, opened 2023-07-05
MxNotificationManager and MxParam initial work.

BranDougherty:
This is my first time doing reverse engineering so I would appreciate a careful review.

I haven't been able to get the matches all the way up to 100 yet. Some might be impossible, but in any case I think it would take me a long time, and some people here are clearly quite good at matching.

This is the first usage of the STL in the decomp, with the `List` class. I think I have also spotted `List` being used in `MxTickleManager`, and I'm sure the STL is used in plenty of places in this code. The STL functions seem to match just fine to me.

There are a few `TODO`s littered in the code. The construction of `MxParam` objects is a little bit of a mystery, with what may be a struct of function pointers passed into the constructor. Could be static methods, not sure. There are also a few unimplemented `MxTickleManager` methods referenced, which seem like `Register` and `Unregister` methods. And there's one unimplemented `MxNotificationManager` method left which I've taken to calling `FlushPending`, because it delivers all notifications addressed to the parameter object, and is only called in `Unregister` as far as I can tell. The origin of the virtual `Create` method is also not fully understood yet but the `MxOmni::Create` and `LegoOmni::Create` methods probably hold the answer.

There is still some work left for this class but I think this is enough to be useful.

I also had to reinstall CMake into a path without any spaces in it to get the CMake build system to work.


itsmattkc:
Overall, seems good and ready to merge. But yeah I'd say you may as well name those functions.

BranDougherty:
I'm sorry to push changes right after a review. But I was able to get a much higher match% by adding a class `MxList` which inherits from `List`, and I also made some changes to allow this to still be built on modern compilers. I wouldn't mind removing some of the new stuff, or putting it in a new PR if you would prefer that.

itsmattkc:
Compatibilty with newer compilers is something I'll accept, but with conditions. I think it can be done without introducing a CMake option/compile definition, instead using macros to autodetect if an old compiler is being used (e.g. check if `_MSC_VER` is defined and whether its version is less than a certain number). I also think `ifdef`s should be contained to `compat.h` as much as possible so they interfere minimally with the rest of the code (in this case, your code from `stlcompat.h` can probably just be moved into it).

BranDougherty:
Done as suggested, how does that look?

LEGO1/mxnotificationmanager.cpp
```diff
@@ -1,15 +1,145 @@
...
+#include "mxparam.h"
+#include "mxtypes.h"
 
-// OFFSET: LEGO1 0x100ac450 STUB
+#include "compat.h"
+#include "decomp.h"
+
+DECOMP_SIZE_ASSERT(MxNotification, 0x8);
+// Can't use DECOMP_SIZE_ASSERT due to STL type size changes.
+//DECOMP_SIZE_ASSERT(MxNotificationManager, 0x40);
```
itsmattkc, line 13: I think we can keep this assert because `DECOMP_SIZE_ASSERT` will probably get nopped on newer compilers.

BranDougherty: OK. Maybe we can just nop if `_MSC_VER > 1020`? The concept of "accuracy" doesn't really apply on different compilers, and I'm not aware of any instances where we depend on type size for correctness yet.

itsmattkc: Yes I think we can pretty safely assume any compiler that isn't MSVC 4.2 has no expectation of instruction accuracy.

PR #79, opened 2023-07-05
 🍕 

ItzSwirlz:


LEGO1/pizza.cpp
```diff
@@ -1,13 +1,35 @@
...
+  this->m_unk90 = 0x80000000;
 }
 
-// OFFSET: LEGO1 0x10038100 STUB
+// OFFSET: LEGO1 0x1002c7cf STUB
+// MxS32 IsleActor::Notify(MxParam *)
+// {
+//   // TODO
+//   return 0;
+// }
```
itsmattkc, line 20: Shouldn't this stub be moved to `isleactor.cpp`?

ItzSwirlz: Good point, in matrix @ktkaufman03 suggested it, I don't know why I didn't move it over there instead of stubbing it here in the wrong file

LEGO1/pizza.cpp
```diff
@@ -1,13 +1,35 @@
...
+
+// OFFSET: LEGO1 0x10038100
 Pizza::~Pizza()
 {
-  // TODO
+  // FIXME: some vtable call from tickle manager, unimplemented atm
+  delete this;
```
itsmattkc, line 26: You don't need to write `delete this`, that's the compiler's responsibility to handle destruction of the object after calling the destructor.

ItzSwirlz: thanks!! Once that's done I'll update this PR

ItzSwirlz:
ive done my best, something is weird in terms of where the fields are located 🤔

PR #80, opened 2023-07-06
MxSemphore + MxThread + MxThread implementions 100% Match

stravant:
* Add MxSemphore + MxThread and the two implementations I could find of MxThread (consumers extend it and override the Run method).

* Implement a function in MxDiskStreamProvider which uses thread and semaphore to confirm correct layout / size of those classes.

* All 100% match except two functions with a pair of registers swapped.

* Fix a bug in MxTimer which went unnoticed because both globals are just `<OFFSET>` in the diff tool.


itsmattkc:
Seems good to me, thanks!

PR #82, opened 2023-07-07
reccmp.py Disambiguate <OFFSET> + Better 100% match mechanics

stravant:
* Rather than using \<OFFSET\> as a replacement for all offsets in a function, label the unique offsets as \<OFFSET1\>, \<OFFSET2\>, etc. Doing this will avoid false-positive 100% matches resulting from the same function being called two times where a different on should have been called or vice versa. And the same for globals. I already encountered one case of this in the wild.

* When a 100% match initially fails, try to make the functions match by swapping register allocations. This makes it possible to get a 100% match where the generated machine code differs only in register allocation.

* Only apply the above when it is possible to reach a 100% match in that way. Otherwise show the developer the unadulterated diff to avoid causing problems. (This register swapping approach is only robust in the case where it results in a 100% match because it doesn't know which registers do what. It relies on the 100% match to confirm the that the way it replaced the registers is valid calling convention wise)

* In the result listing, show the functions which are "effective matches" in this way as "100%*" instead of "100%".

### DIFF
--- a/tools/reccmp/reccmp.py
+++ b/tools/reccmp/reccmp.py
@@ -10,6 +10,7 @@ import logging
 import os
 import sys
 import colorama
+import re
 
 parser = argparse.ArgumentParser(allow_abbrev=False,
   description='Recompilation Compare: compare an original EXE with a recompiled EXE + PDB.')
@@ -248,9 +249,21 @@ print()
 
 md = Cs(CS_ARCH_X86, CS_MODE_32)
 
-def sanitize(file, mnemonic, op_str):
-  offsetplaceholder = '<OFFSET>'
+class OffsetPlaceholderGenerator:
+  def __init__(self):
+    self.counter = 0
+    self.replacements = {}
 
+  def get(self, addr):
+    if addr in self.replacements:
+      return self.replacements[addr]
+    else:
+      self.counter += 1
+      replacement = '<OFFSET%d>' % self.counter
+      self.replacements[addr] = replacement
+      return replacement
+
+def sanitize(file, placeholderGenerator, mnemonic, op_str):
   op_str_is_number = False
   try:
     int(op_str, 16)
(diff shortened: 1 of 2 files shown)


itsmattkc:
I'd been considering ways to calculate "effective accuracy", but hadn't considered bruteforcing all registers. For the most part I think it's probably a reasonable thing to have; it doesn't seem to address a majority of situations, but it does for a handful which is something.

You can probably exclude `ebp`, `esp`, `bp`, or `sp`. Those are generally reserved by the compiler/architecture and shouldn't ever be used for calculations.

stravant:
Our compiler definitely uses `ebp` as general storage in some cases.

`esp` _could_ be excluded but since the algorithm is blind to what purpose the registers actually serve it felt right to include all of them.

tools/reccmp/reccmp.py
```diff
@@ -383,14 +471,21 @@ def parse_asm(file, addr, size):
...
 
           if recinfo.size:
             udiff = difflib.unified_diff(origasm, recompasm, n=10)
 
             # If verbose, print the diff for that funciton to the output
             if verbose:
-              if ratio == 1.0:
-                print("%s: %s 100%% match.\n\nOK!" % (hex(addr), recinfo.name))
+              if effective_ratio == 1.0:
+                ok_text = "OK!" if plain else (colorama.Fore.GREEN + "✨ OK! ✨" + colorama.Style.RESET_ALL)
```
itsmattkc, line 482: You really want these emoji don't you lol, I saw them in #74 too

stravant: More dopamine = more contributions. Can't argue with that.

stravant:
Are you going to merge this?

itsmattkc:
Yep, done

PR #83, opened 2023-07-07
Add SmartHeap

itsmattkc:
This appears to be a match for the version of SmartHeap used in the real game, and if so, should make it possible to achieve a perfect byte match. Making a PR so we can test before merging into `master`.


ItzSwirlz:
maybe have a separate CI/even a separate build step for SH

itsmattkc:
> maybe have a separate CI/even a separate build step for SH

SmartHeap isn't compiled, it's just linked. There is no additional step.

CMakeLists.txt
```diff
@@ -177,6 +178,13 @@ add_library(lego1 SHARED
   LEGO1/viewmanager.cpp
 )
 
+if (ISLE_USE_SMARTHEAP)
+  target_compile_definitions(lego1 PRIVATE USE_SMARTHEAP)
```
madebr, line 182:
```suggestion
  target_compile_definitions(lego1 PRIVATE USE_SMARTHEAP)
  target_compile_options(lego1 PRIVATE "/FI${CMAKE_SOURCE_DIR}/3rdparty/smartheap/SMRTHEAP.HPP")
```

CMakeLists.txt
```diff
@@ -194,6 +202,13 @@ if (ISLE_BUILD_APP)
   # Include LEGO1 headers in ISLE
   target_include_directories(isle PRIVATE "${CMAKE_SOURCE_DIR}/LEGO1")
 
+  if (ISLE_USE_SMARTHEAP)
+    target_compile_definitions(isle PRIVATE USE_SMARTHEAP)
```
madebr, line 206:
```suggestion
    target_compile_definitions(isle PRIVATE USE_SMARTHEAP)
    target_compile_options(isle PRIVATE "/FI${CMAKE_SOURCE_DIR}/3rdparty/smartheap/SMRTHEAP.HPP")
```

CMakeLists.txt
itsmattkc: Good idea. I was including `SMRTHEAP.HPP` through the `USE_SMARTHEAP` macro (currently just `isleapp.cpp`), but this is probably cleaner/more reliable.

WamWooWam:
This seems... at best questionably legal to do. Maybe a private CI step that validates with SH would be enough?

madebr:
> This seems... at best questionably legal to do. Maybe a private CI step that validates with SH would be enough?

True. Perhaps move the binaries + headers to an external repo and optionally add it using some `find_package` cmake magic.

itsmattkc:
I've thought about this and my gut feeling is that a really old version of SmartHeap designed for MSVC 4.x is likely to be just as much abandonware as LEGO Island itself is (i.e. the same amount of legally questionable as the rest of the repo). They certainly don't still sell this version, and it's hard to imagine someone being able to use it in lieu of a real license in a modern commercial program.

It's hard to even tell if SmartHeap is still being supported at all. Sure there's a website up and it claims to command a high price, but many of the site's pages have been dead/missing for years, and SmartHeap itself doesn't seem to have received an update for nearly a decade. If there is someone who'd want to file a DMCA for it, GitHub [claims we'd get a chance to simply remove it](https://docs.github.com/en/site-policy/content-removal-policies/dmca-takedown-policy#a-how-does-this-actually-work) then, but I think there's a fairly high likelihood no one would bother.

ItzSwirlz:
The point of SmartHeap nowadays is also a joke, if you wanted to make a game as a developer and couldn't handle 16 MB of memory, you might as well just learn rust /s

PR #84, opened 2023-07-08
MxEntity

ItzSwirlz:
just trying to figure out what's happening in the destructor primarily. I did consider the existence of a Destroy() function which got me roughly about to the same matching as where i'm at.

### DIFF
--- a/LEGO1/mxentity.cpp
+++ b/LEGO1/mxentity.cpp
@@ -3,13 +3,10 @@
 // OFFSET: LEGO1 0x1001d190
 MxEntity::MxEntity()
 {
-  // FIXME: This won't match because of m_atom. How would it access m_internal?
-  this->m_atom = NULL;
   this->m_mxEntityId = -1;
 }
 
 // OFFSET: LEGO1 0x1000c110
 MxEntity::~MxEntity()
 {
-  *this->m_atom;
 }
--- a/LEGO1/mxentity.cpp
+++ b/LEGO1/mxentity.cpp
@@ -1,19 +1,15 @@
 #include "mxentity.h"
 
-// OFFSET: LEGO1 0x1001d190 STUB
+// OFFSET: LEGO1 0x1001d190
 MxEntity::MxEntity()
 {
-  // TODO
+  // FIXME: This won't match because of m_atom. How would it access m_internal?
+  this->m_atom = NULL;
+  this->m_mxEntityId = -1;
 }
 
-// OFFSET: LEGO1 0x1000c110 STUB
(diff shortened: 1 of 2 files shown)


LEGO1/mxentity.cpp
```diff
@@ -1,19 +1,15 @@
 #include "mxentity.h"
 
-// OFFSET: LEGO1 0x1001d190 STUB
+// OFFSET: LEGO1 0x1001d190
 MxEntity::MxEntity()
 {
-  // TODO
+  // FIXME: This won't match because of m_atom. How would it access m_internal?
+  this->m_atom = NULL;
```
itsmattkc, line 7: I'm fairly certain MxAtomId is a member of MxEntity (i.e. not a pointer) and `m_internal` being set to NULL is an inlined constructor.

LEGO1/mxentity.cpp
```diff
@@ -1,19 +1,15 @@
...
 {
-  // TODO
-}
-
-// OFFSET: LEGO1 0x10001070 STUB
-void MxEntity::Destroy()
-{
-  // TODO
-}
+  *this->m_atom;
```
itsmattkc, line 14: This destructor is almost certainly empty, the `Unwind`s are the compiler calling the destructors of the member MxAtomId and the base MxCore.

LEGO1/mxentity.cpp
ItzSwirlz: so what could this be then? And how could MxAtomId (a class) be a member? 🤔 haven't done class inheritance in a while

LEGO1/mxentity.cpp
ItzSwirlz: Today and tomorrow I'm busy, so hopefully late tomorrow or Thursday I can figure this out, I'm missing something here. 
image

LEGO1/mxentity.cpp
itsmattkc: An object (class) can be a member of another class in C++. Currently you have the pointer to an MxAtomId object (`MxAtomId *`), but the object itself can be embedded by dropping the `*`. The compiler will then call its constructor and destructor automatically without having to write code for it. Then as I said, the setting `m_internal` to `NULL` is probably MxAtomId's inlined constructor.

LEGO1/mxentity.cpp
itsmattkc: If you have the MxAtomId as a member, as mentioned above, this will work itself out automatically (the compiler will automatically call its destructor).

PR #86, opened 2023-07-10
100% match of many functions in the LegoGameState::Save codepath

stravant:
* Various 100% Matches: LegoGameState::GetState, CreateState, RegisterState, GetFileSavePath. LegoStream::WriteVariable, ReadVariable.

* Implement a few additional parts of ObjectFactory / MxOmni::Create

* Fix up some of the definitions in LegoStream to be more ergonomic to actually use in context. (they remain 100% matches)

@foxtacles: Changes by me detailed in this comment: https://github.com/isledecomp/isle/pull/86#issuecomment-1759928468


itsmattkc:
While a CLI app is a fun idea, I feel similarly to how I feel with #74; it feels like feature creep, an extra thing we have to maintain that distracts us from decompiling, particularly since it requires splitting up the library.

ItzSwirlz:
Maybe we could put it in another repo in the isledecomp github org, just so we have it lying around

stravant:
I don't see how this is a distraction.

Do you not plan to run any code until 99% of the codebase is decompiled? Because the only way to run any of the code before that will be a tool like this. The game will not bootstrap itself without an external tool driving it until almost all of the code is decompiled.

We can actually take advantage of the fact that the codebase is well abstracted to load / save save files and assets using the game classes to verify parts of the functionality as things develop rather than

"Splitting the library" isn't an actual cost, it's just a dozen extra lines of CMake code total.

itsmattkc:
> Do you not plan to run any code until 99% of the codebase is decompiled? Because the only way to run any of the code before that will be a tool like this.

That's not entirely true. Through DLL injection, our recompiled code can be grafted over the original at runtime, which could potentially be even more reliable since it tests the code in context with the rest of the game as it plays. My plan was at some point to write a launcher for the game that would inject our functions automatically (mapping the addresses to the original DLL with the PDB the same way `reccmp` does).

> Would you be more amenable to structuring this as unit tests rather than a CLI tool? My opinion is that, sure, you can focus on only decompiling now when there's plenty of low hanging fruit but you will inevitably need more tooling like this to get through the trickier parts of the decomp.

Sure, I think unit tests would be fine. I think my main concern is the idea of committing to adding extra functionality/features this early in the project. In fact, my original plan for this repo was it would just be the original game, and that enhancements/modernizations would go into a fork, though that may change depending on how things go.

stravant:
Can I at least split the library into the original DLL only exposing specific exported functions + a static library exposing all the functions? That way I can put this code like this in another repo and depend on the codebase as a submodule. As it is there's no good way for me to do that.

itsmattkc:
Yes, I think we can just treat it as building both a shared and static version of LEGO1 (i.e. not denote it as "core", rather just the same library but "static"). That would be perfectly fine.

stravant:
Updated this PR to only have the decomped methods and none of the extra stuff.

stravant:
(It's ready for re-review now)

LEGO1/infocenterstate.h
```diff
@@ -23,6 +23,38 @@ class InfocenterState : public LegoState
   {
     return !strcmp(name, InfocenterState::ClassName()) || LegoState::IsA(name);
   }
+
+  inline MxU32 GetSomething(int p_index) { return m_buffer[p_index]; }
+  inline void SetSomething(int p_index, MxU32 p_value) { m_buffer[p_index] = p_value; }
```
itsmattkc, line 28: Might be good to be a little more descriptive than `GetSomething` as the name. At this point, that could describe accessing _any_ of the members.

Something like `GetElementOfSomeBuffer`/`GetElementOfSomeArray` would be better (assuming we don't know what `m_buffer` actually is yet)

LEGO1/infocenterstate.h
```diff
@@ -23,6 +23,38 @@ class InfocenterState : public LegoState
...
+  void *unk4;
+  MxU16 unk5;
+  MxU16 unk6;
+  MxU16 unk7;
+  MxU16 padding2;
+  void *unk8;
+  MxU16 unk9;
+  MxU16 unk10;
+  MxU16 unk11;
+  MxU16 padding3;
```
itsmattkc, line 54: Unknown vars should preferably use the `undefined` typedefs in `decomp.h`. Also are the `padding` members necessary? I would have thought the compiler would fill those in automatically.

stravant: Not only will it but it has to due to alignment rules, I just prefer to be explicit so I can mess something up while transcribing what I did in Ghidra. I'll avoid it if you don't like the clutter.

LEGO1/infocenterstate.h
stravant: No idea what that array is yet, needed it to make the code run when I was trying to run save.

I'll take it out, it's not actual code that we know was there anyways.

stravant: Named it GetInfocenterBufferElement, and removed the setter.

stravant:
* Removed padding.
* Renamed setter for unknown infocenterstate function which Save uses.
* Changed code to use undefined1/2/4 where applicable.

stravant:
You gonna merge this?

itsmattkc:
Yeah, I'll look at it when I have more time.

LEGO1/infocenterstate.h
```diff
@@ -23,6 +25,36 @@ class InfocenterState : public LegoState
...
+  // Size: 0xC
+  struct SomeStruct
+  {
+    undefined4 unk1;
+    undefined2 unk2;
+    undefined2 unk3;
+    undefined2 unk4;
+  };
+
+  undefined2 unk1;
```
itsmattkc, line 41: `unks` should be numbered by their offset so they can more easily be referred to in asm/pseudocode.

LEGO1/legogamestate.cpp
```diff
@@ -1,39 +1,261 @@
...
+  unsigned char len = strlen(s_endOfVariables);
+  if (p_stream->Write(&len, 1) == SUCCESS)
+    return p_stream->Write(s_endOfVariables, len);
+  return FAILURE;
+}
+
+struct LegoSaveDataEntry3
+{
+  char *m_name;
+  void *m_pSomething1;
```
itsmattkc, line 123: Same here, unknown variables should have their offset, not be numbered sequentially.

PR #87, opened 2023-07-10
lego1: MxVariableTable

disinvite:
We already have a class `MxStringVariable` with key/value pair members that are both `MxString`. This is the base class for `LegoBackgroundColor` and a few others based on the vtable. I submit that this matches `MxVariable` used by the `MxVariableTable` class, so I renamed it.

`MxVariableTable` is a hash table data structure with linked list for overflow. The "available" source code has reference to an `MxHashTable` class with a companion cursor API to access the data, but without the actual implementation. I used the same structure here. The implementation is tricky because some functions are inline in the key/val version of `SetVariable` but not in the other: namely the Resize and "NodeInsert" functions.

MxVariable (formerly MxStringVariable) had a constructor marked as a dll export. I couldn't find a match for this in the Ghidra analysis, so I removed it.

MxHashTable has a parent class that I've left in the main header file as "HashTableParent". You can see its members and virtual functions based on how it's initialized in MxOmni::Create.

I am undecided on what to do with the members of the "hash table node" class. By our convention, the members should be private with accessor methods, but this complicates the pointer access between nodes a little bit. Could this have been just a struct instead?


stravant:
Dunno if you saw the stuff I was going in Ghidra but I agree with this interpretation.

I added an "MxVariables" namespace with some of the implementations of this MxVariable which have different setters too.

disinvite:
Main notes updated above.

@stravant I think it _is_ meant to be a "Destroy" function in MxVariable. We don't re-set the vtable pointer inside like with other destructors. The hash table's destructor calls into there for each node (to clean up whatever kind of value it is) before deleting the node altogether.

itsmattkc:
It's a very bizarre thing for them to do, but yes @disinvite appears to be correct. If changed to a virtual destructor, its accuracy significantly drops, while in its current form it matches 100%.

Maybe one of their interns wrote it...

PR #88, opened 2023-07-10
reccmp: template compare annotations

BranDougherty:
Adds the ability to compare non-inlined instantiations of template functions to reccmp. This could potentially be used to compare STL versions (although the MSVC420 version seems like a match so far), and other template classes like `MxPtrList`.

Usage:
```
// OFFSET: LEGO1 0x100ac320 TEMPLATE
// list<unsigned int,allocator<unsigned int> >::~list<unsigned int,allocator<unsigned int> >
```

I've checked the "Allow edits by maintainers" box please remove the example code before merging.


foxtacles:
I'm going to give this a try soon on real code, I think with `MxList` we have a good test case where we should be able to compare the non-inlined destructors of the various instantiations.

foxtacles (approved):
I've added template compare annotations for instances of `MxList`. Looks good to me. I think this is useful and will be even more so since I'm seeing more templated stuff coming up soon. What do you think @itsmattkc?

itsmattkc:
Yeah I think it's a good idea. If it all works, I'd be happy to merge.

foxtacles:
> If it all works, I'd be happy to merge.

Works correctly as far as I can tell, if you could double check it that would be great since it touches reccmp.

PR #89, opened 2023-07-13
Some MxBitmap vtable functions

ItzSwirlz:
Going for all out on most of the vtable functions. In the comments is the reasoning for the MxBITMAPINFO structure, done by @disinvite .

There are some weird things preventing full matches, hopefully they'll be worked out

ImportColorsToPalette (0x100bd450) is not assigned to MxBitmap, but most of the vtable functions call it. This appears to be importing the contents of a RGBQUAD to a MxPalette. This took me several tries to reverse engineer, here is my latest, successful attempt. https://hackmd.io/@KsNMFSBHTO2gxDyRIPRQ1g/H1LCVQXon


LEGO1/mxbitmap.cpp
```diff
@@ -28,10 +33,34 @@ int MxBitmap::vtable14(int)
...
-int MxBitmap::vtable18(BITMAPINFOHEADER *p_bmiHeader)
+// OFFSET: LEGO1 0x100bcba0
+MxResult MxBitmap::vtable18(BITMAPINFOHEADER *p_bmiHeader)
 {
-  return 0;
+  MxResult result = FAILURE;
+  int width = p_bmiHeader->biWidth;
+  int height = p_bmiHeader->biHeight;
+
+  this->m_info = (BITMAPINFO*) malloc(0x428);
```
itsmattkc, line 43: These hardcoded numbers should probably be avoided. Most likely numbers like `0x428` actually came from somewhere (e.g. `malloc(sizeof(BITMAPINFO))`) so it would be better to figure out their origins . Since this is C++, there's a good chance these might actually be `new`s rather than `malloc`s (e.g. `new BITMAPINFO`), which would figure out the size and malloc automatically.

PR #91, opened 2023-07-16
MxNotificationManager::FlushPending

BranDougherty:
Sitting at 90.51% on my machine, but if you look it really seems to just be the usual MSVC randomness with a bit more mov instruction reordering.

This is the last function left in `MxNotificationManager` as far as I can tell, so I'll call "Tutorial complete!"


itsmattkc:
Looks good, thanks!

PR #92, opened 2023-07-18
lego1: match MxVideoManager::RealizePalette

foxtacles:
Completes/matches the implementation of `MxVideoManager::RealizePalette` and adds the `MxDisplaySurface` class and its virtual functions to do so.


stravant:
`MxUnknown100dc768` is MxDisplaySurface and has most of its members filled out in Ghidra.

foxtacles:
> `MxUnknown100dc768` is MxDisplaySurface and has most of its members filled out in Ghidra, is there some reason you didn't want to use them?

That's good to know! I've not been using the shared Ghidra server so I wasn't aware. I'll see what is there and at the least give the class its proper name.

itsmattkc:
Looks good, thanks!

PR #93, opened 2023-07-19
Start IslePathActor

ItzSwirlz:
A handful of other things rely on it, so let's get it started:

image

### DIFF
--- a/LEGO1/islepathactor.cpp
+++ b/LEGO1/islepathactor.cpp
@@ -1,1 +1,15 @@
 #include "islepathactor.h"
+
+// OFFSET: LEGO1 0x1001a200
+IslePathActor::IslePathActor()
+{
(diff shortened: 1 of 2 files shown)


LEGO1/islepathactor.cpp
```diff
@@ -1 +1,15 @@
 #include "islepathactor.h"
+
+// OFFSET: LEGO1 0x1001a200
+IslePathActor::IslePathActor()
+{
+  this->m_pLegoWorld = NULL;
+  this->m_unk13c = 0x40c00000;
+  this->m_fourcc = 0x3f800000;
```
itsmattkc, line 8: Both of these are floats, not FourCCs. `0x40c00000` is `6.0` and `0x3f800000` is `1.0`.

PR #94, opened 2023-07-20
MxTickleManager mostly done

BranDougherty:
A 100% match of everything in MxTickleManager except the actual Tickle method. SetClientTickleInterval is at a pitiful 24.14% but if you look at 85f4a5521978180c54b46c2b682a857a77ffaa7c you'll see the exact same code at 100%, it just gets obliterated by register allocation (the part that checks the flag). The Tickle method also feels really close to me, but it has issues with register allocation too (the part that checks the flag, and what seems to be unnecessary shuffling of a few vars between the stack and registers). So it's only at 52.17%, but it seems unproductive to keep working on it right now as it may just sort itself out.

### DIFF
--- a/LEGO1/mxticklemanager.cpp
+++ b/LEGO1/mxticklemanager.cpp
@@ -29,16 +29,18 @@ MxTickleManager::~MxTickleManager()
   }
 }
 
+// TODO: Match.
 // OFFSET: LEGO1 0x100bdde0
 MxResult MxTickleManager::Tickle()
 {
   MxTime time = Timer()->GetTime();
 
   MxTickleClientPtrList::iterator it = m_clients.begin();
+
   while (it != m_clients.end()) {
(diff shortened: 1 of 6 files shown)


LEGO1/mxticklemanager.cpp
```diff
@@ -0,0 +1,113 @@
...
+  m_client = p_client;
+  m_interval = p_interval;
+  m_lastUpdateTime = -m_interval;
+}
+
+// OFFSET: LEGO1 0x100bdd30
+MxTickleManager::~MxTickleManager()
+{
+  while (m_clients.size() != 0) {
+    MxTickleClient *client = m_clients.front();
```
ItzSwirlz, line 26: Just commenting, not starting a review yet - interesting, I didn't know these things existed in C++.

PR #96, opened 2023-07-22
Small implementations in MxEntity/MxEventManager

ItzSwirlz:


LEGO1/mxentity.cpp
```diff
@@ -9,4 +9,12 @@ MxEntity::MxEntity()
...
 MxEntity::~MxEntity()
 {
+}
+
+// OFFSET: LEGO1 0x10001070
+MxResult MxEntity::SetEntityId(MxS32 p_id, MxAtomId *p_atom)
+{
+  this->m_mxEntityId = p_id;
+  // FIXME: MxAtomId operator call, probably it will be return whatever the comparison is for comparing this->m_atom to the p_atom
+  return FALSE;
```
itsmattkc, line 19: The MxAtomId `operator=` call is just setting the object with `=`. Most likely that means the param is actually a reference (`&`) rather than a pointer (`*`) - the two are indistinguishable from an asm point of view but this would make more sense code-wise if using `operator=`. Also if it's an MxResult and returns 0, it's almost certainly returning `SUCCESS` not `FALSE`.

```suggestion
MxResult MxEntity::SetEntityId(MxS32 p_id, const MxAtomId &p_atom)
{
  this->m_mxEntityId = p_id;
  this->m_atom = p_atom;
  return SUCCESS;
```

LEGO1/mxentity.h
```diff
@@ -24,6 +24,8 @@ class MxEntity : public MxCore
   {
     return !strcmp(name, MxEntity::ClassName()) || MxCore::IsA(name);
   }
+
+  virtual MxResult MxEntity::SetEntityId(MxS32 p_id, MxAtomId *p_atom);
```
itsmattkc, line 28: Including class name in definition is not standards compliant. Also if the above change is used, exchange `MxAtomId *` for `const MxAtomId &`.

LEGO1/mxeventmanager.cpp
```diff
@@ -1,13 +1,21 @@
...
-  // TODO
+  // TODO: MxMediaManager::TerminateThread call
 }
+
+// OFFSET: LEGO1 0x100c0450
+void MxEventManager::Init()
+{
+  // huh?
+  // imagine if the devs actually put this here to troll us
+  return;
```
itsmattkc, line 20: This probably had something during development but was later removed. The comments and `return` aren't necessary.

LEGO1/mxentity.cpp
```diff
@@ -9,4 +9,12 @@ MxEntity::MxEntity()
...
 MxEntity::~MxEntity()
 {
+}
+
+// OFFSET: LEGO1 0x10001070
+MxResult MxEntity::SetEntityId(MxS32 p_id, const MxAtomId &p_atom)
+{
+  this->m_mxEntityId = p_id;
+  this->m_atom = p_atom;
+  return FALSE;
```
itsmattkc, line 19:
```suggestion
  return SUCCESS;
```

PR #98, opened 2023-07-23
Matching spree

ItzSwirlz:


LEGO1/legonavcontroller.cpp
```diff
@@ -155,7 +155,7 @@ float LegoNavController::CalculateNewTargetSpeed(int p_pos, int p_center, float
   float result;
   int diff = p_pos - p_center;
 
-  if (diff > this->m_mouseDeadzone)
+  if (diff < this->m_mouseDeadzone)
```
foxtacles, line 158: This makes no sense, logically speaking. The function has already been matching 100% before, I suggest double checking this.

ItzSwirlz: The diff was showing jge vs jle so I thought I'd check it out anyways

itsmattkc: Was the diff also showing the registers flipped? Sometimes the compiler will flip the operation around but also the order of the registers, so logically it will be the same conditional

ItzSwirlz: Considering I cant find the original diff ill revert this

LEGO1/mxpresenter.cpp
```diff
@@ -116,8 +117,8 @@ void MxPresenter::InitVirtual()
 // OFFSET: LEGO1 0x1000bf90
 void MxPresenter::VTable0x44(undefined4 param)
 {
-  m_unk0xc |= 1 << (unsigned char)m_unk0x8;
-  m_unk0x8 = param;
+  m_unk0xc |= 1 << (unsigned char)m_someFlag;
+  m_someFlag = param;
```
itsmattkc, line 121: I'm not sure "someFlag" is really any more helpful than "m_unk0x8". But since we've determined some of this behavior, we can probably actually come up with some more useful names either way.

These are interesting because it changes the value at `0x8` but first uses it to store a bit flag in `0xC`. It's almost like `0xC` is holding a cache of previously set values. And then `DoneTickle`, caches `0x8` again and sets it to 0, the name implying that something has ended. Perhaps we could call `0x8` "m_currentTickleFlag" and `0xC` "m_previousTickleFlags".

ItzSwirlz: What values get set with "param"? Is there a time that either var is set to some "empty" value?

itsmattkc: Look at vtable 18 to 28, they all do the same thing, caching the last value of `0x8` to `0xC` and then setting `0x8`. This does that too, but with a custom value instead of a hardcoded one.

ItzSwirlz: Here's what I get with vtable0x28:
image

Basically, a number. Should I add to `mxtypes.h` typedefs for these?
```
#typedef unsigned char TICKLEFLAGS_SIX 6;
```

PR #99, opened 2023-07-29
[wip] lego1: MxAtomId and its tree

disinvite:
Been working on this for a little while; it's not finished but I could use some help. Match percentage is low in some places for reasons I'll explain.

The +0x44 member of `MxOmni` is a red-black binary tree that holds a string value and two `word`s. This is `TreeValue`. There is a connection between this and `MxAtomId`, which itself has only a single `char*` member. ~I'm not sure if there are any other BSTs used in the game, but~ I've named this one `MxBinaryTree` for now. There is a global `TreeNode` pointer at `0x101013f0` that is used to represent a null leaf node. This seems to be standard practice for RB tree implementations rather than just checking for NULL.

The members of the `TreeNode` class are defined in this order: right, parent, left. Yes... backwards. I confirmed that this is correct with a memory dump, unless the intent was to sort the tree in reverse-alphabetical order. The member names right now are `m_child0` (right) and `m_child1` (left) just so it makes sense in ghidra. This should definitely be changed before finalizing the PR.

There is an inlined init function `newTreeNode` that new's the POD object. If we did the obvious thing and used a constructor here, this would add the exception protection (i.e. `mov eax, fs:[0]`) to the start of the function `MxBinaryTree::Insert` at `0x100ad4d0`. But that function doesn't have those instructions. I'm not 100% confident that this is right, but the node's members are always set in the same order after malloc.

Where I'm having trouble matching/reversing this stuff is that there appear to be a lot of inlined function calls, but they are only sometimes inlined. For example: the function at `0x100ad120` compares two `TreeValue` strings (basically a big wrapper for `strcmp`) but it is *not* inlined when called at `0x100ad072`. The same is true for the "search" function at `0x100ad780`, which is called directly at `0x100ad2ae` but inlined in many other places.

I'm not sure how the "Successor" function at `0x100ad480` is meant to be called. It uses `[ECX]` as both input and output, but if I made this an instance method, then that would mean setting `TreeNode->m_child0` which doesn't make a lot of sense.

In all this work, I haven't uncovered what `MxAtomId` is _used for_ exactly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,7 @@ add_library(lego1 SHARED
   LEGO1/mxaudiopresenter.cpp
   LEGO1/mxautolocker.cpp
   LEGO1/mxbackgroundaudiomanager.cpp
+  LEGO1/mxbinarytree.cpp
   LEGO1/mxbitmap.cpp
   LEGO1/mxcompositemediapresenter.cpp
   LEGO1/mxcompositepresenter.cpp
--- a/LEGO1/mxatomid.cpp
+++ b/LEGO1/mxatomid.cpp
@@ -54,13 +54,15 @@ void MxAtomId::Destroy()
 // OFFSET: LEGO1 0x100ad1c0
 MxAtomId &MxAtomId::operator=(const MxAtomId &atomId)
 {
-  // TODO
-  const char *temp = m_internal;
   if (m_internal)
     Destroy();
 
-  if (atomId.m_internal && MxOmni::GetInstance() && AtomIdTree())
-    try_to_open(temp, LookupMode_Exact);
+  if (atomId.m_internal && MxOmni::GetInstance() && AtomIdTree()) {
+    TreeValue *value = try_to_open(atomId.m_internal, LookupMode_Exact);
+    value->RefCountInc();
+  }
+
+  m_internal = atomId.m_internal;
 
   return *this;
 }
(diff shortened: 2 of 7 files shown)


disinvite:
Turns out there are scores of similar functions throughout the game, so the RB-Tree is definitely a generic datatype and not a one-time thing. To illustrate: compare the functions at 10020b20, 10058330, 10083840, and 100a7800 to my Successor function here (100ad480).

Unfortunately, changing it around to a template class makes the match _even worse_ in the short term. It's up to chance whether MSVC will inline something or not. But it's something to go on. I was on the right track with the "not-a-constructor" for tree nodes. See: 10020bb0 and 1002c4a0.

BranDougherty:
I'd like to suggest that this might be an STL map for a few reasons.
- It's a template class, also backed by an RB tree.
- I don't see any reference to MxCore or any MindScape specific classes in the implementation, whereas other MindScape containers like MxPtrList wind up inheriting from MxCore.
- The details seem to line up, like MSVC420's RB tree implementation also makes use of a static null leaf.

I made a mistake when I was working on MxNotificationManager, where I assumed they were using some custom linked list data structure but it was just List. I'm not sure on the exact structure of what's being stored here, you know that better than me, but I would like to avoid us making the same mistake again.

disinvite:
Thanks, you may be right. I'll check it out. The functions in `STL/TREE.H` look remarkably similar to what we have here. (But this is a widely available algorithm.)

Your point about an `MxCore` reference being a sign that something is custom also makes sense. We _did_ have that for `MxHashTable` that I originally suspected was STL.

BranDougherty:
The implementations look similar enough, but we've had success with the MSVC420 STL so far so I'd also try `MSVC420/include/{STL.H,MAP}`, with the implementation stored in `include/XTREE`.

disinvite:
Rewritten using the STL in #109. Closing this.

PR #101, opened 2023-08-05
Implement/match MxDSAction constructor

foxtacles:
`MxDSAction` uses the `MxVector` classes from https://github.com/isledecomp/isle/pull/100 and it looks like we've got a perfect match in this constructor, only a minor change was needed in the `MxVector2` base class (inlined `SetData`)

### DIFF
--- a/LEGO1/mxdsaction.cpp
+++ b/LEGO1/mxdsaction.cpp
@@ -1,7 +1,7 @@
 #include "mxdsaction.h"
 
-#include "float.h"
-#include "limits.h"
+#include <float.h>
+#include <limits.h>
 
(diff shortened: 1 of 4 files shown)


LEGO1/mxdsaction.cpp
```diff
@@ -1,12 +1,39 @@
 #include "mxdsaction.h"
 
+#include "float.h"
+#include "limits.h"
```
itsmattkc, line 4: These should probably be in `<>` since they're external headers. Otherwise this all looks good.

foxtacles: True, not sure how I missed that. Thanks!

PR #102, opened 2023-08-06
Add GetDeviceModeFinder to IsleApp::WndProc

foxtacles:
Resolves a small chain of unknown in `IsleApp::WndProc` (does not appear to impact accuracy).

For some reason the `d3d.h` include (part of the DX5 SDK) cannot be found. Can somebody familiar with the build system assist here and figure out what needs to be done to fix this? Thank you!


itsmattkc:
> For some reason the d3d.h include (part of the DX5 SDK) cannot be found. Can somebody familiar with the build system assist here and figure out what needs to be done to fix this? Thank you!

I think we might need to run `dx5sdk/sdk/samples/dxsetenv.bat` (like vcvars32), which will add the DX5 stuff into the `INCLUDE` and `LIB` vars used by MSVC.

itsmattkc:
Try merging in the latest `master`, should compile with DX5 automatically now

PR #103, opened 2023-08-07
Implement most of MxDSAction

foxtacles:
Implemented and matched most of `MxDSAction` and resolved some unknowns from `MxDSObject`. A few functions aren't quite completed/looked at yet, but I'll leave those for some other time.

There appears to be a major regression in `WndProc` on first glance but as far as I can tell "only" a lot of instructions have been reordered, so I believe it has nothing directly to do with changes in this PR.


itsmattkc:
Looks good to me, will merge. Regression is probably unrelated.

PR #105, opened 2023-08-09
Implement MxDSAction::AppendData

foxtacles:
`MxDSAction::AppendData` matches fully apart from one instruction placed differently. I've given the function and attributes preliminary "unknown" labels since I'm not aware of their purpose (yet).

### DIFF
--- a/LEGO1/mxdsaction.cpp
+++ b/LEGO1/mxdsaction.cpp
@@ -3,8 +3,6 @@
 #include <float.h>
 #include <limits.h>
 
-#define TWOCC(a, b) (((a) << 0) | ((b) << 8))
-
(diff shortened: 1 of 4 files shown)


LEGO1/mxdsaction.cpp
```diff
@@ -3,13 +3,18 @@
 #include <float.h>
 #include <limits.h>
 
+#define TWOCC(a, b) (((a) << 0) | ((b) << 8))
```
itsmattkc, line 6: Minor code maintenance idea: might be worth combining this and `FOURCC` from `mxdsfile.cpp` into one global header, perhaps `mxtypes.h`

PR #108, opened 2023-08-13
Implement MxDSStreamingAction

foxtacles:
Everything matches except for the destructor (which requires implementation of `MxDSBuffer`)

### DIFF
--- a/LEGO1/mxdsstreamingaction.cpp
+++ b/LEGO1/mxdsstreamingaction.cpp
@@ -47,6 +47,14 @@ MxDSStreamingAction *MxDSStreamingAction::CopyFrom(MxDSStreamingAction &p_dsStre
   return this;
 }
 
+// OFFSET: LEGO1 0x100cd090
+MxBool MxDSStreamingAction::HasId(MxU32 p_objectId)
+{
+  if (this->m_internalAction)
+    return this->m_internalAction->HasId(p_objectId);
+  return FALSE;
+}
+
 // OFFSET: LEGO1 0x100cd1e0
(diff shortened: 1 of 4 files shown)


LEGO1/mxdsstreamingaction.cpp
```diff
@@ -0,0 +1,91 @@
+#include "mxdsstreamingaction.h"
+
+DECOMP_SIZE_ASSERT(MxDSStreamingAction, 0xb4)
+
+// OFFSET: LEGO1 0x100cd010
+MxDSStreamingAction::MxDSStreamingAction(MxDSAction &p_dsAction, MxU32 p_offset)
+{
+  Init();
+
+  MxDSAction::operator=(p_dsAction);
```
itsmattkc, line 10: This is interesting, I feel like what this might actually resolve to is `*this = p_dsAction`. I _believe_ that would produce the same result but is more likely to be what was written by the developers.

itsmattkc: Hmm if it doesn't work out of the box than maybe not? It just seems like an odd thing for the developers to have written explicitly, but I guess the asm doesn't lie. Up to you if you want to keep the change or not.

foxtacles: I'd keep the change, it makes the code a little more readable and isn't completely illogical. Welcoming any suggestions for further improvements in the future of course.

itsmattkc: Sounds good, will merge

PR #109, opened 2023-08-13
Implement MxAtomId and reference counter (stl set)

disinvite:
Rewrite of #99. Implement all of MxAtomId. This class interfaces with a reference counter data structure created in MxOmni. I had this partly reversed in the previous PR. Hat tip to @BranDougherty for identifying the red-black tree as a use of the STL.

I used a `<set>` for holding the reference counters. This seems odd because a `<map>` should be the best option: we have MxString as the key and MxU16 as the value. And yet if you try with a map, when we alloc the structure in `MxOmni::Create`, the leaf (or _Nil) node size is much larger than it should be. Also: the Inc/Dec functions at 0x100ad7f0/0x100ad800 refer to offset +0x10, so there's obviously more there than just the MxU16. That offset includes the preceding MxString for a total of 20 bytes on the combined key/value type (rounded to nearest 4 byte offset). I couldn't find a way to define a totally custom key/value type where this would work with a `<map>`, so `<set>` it is.

As a bonus, I built out most of `MxOmni::Destroy` since it touches the set. I may come back to this later in a follow-up PR.

We now get 4786 warnings all over the place. The STL uses templates on top of templates and you quickly exceed the arbitrary 255 character limit for debug symbols. The pragma I added to compat.h removes _most_, but not all of these warnings. If you google for results on disabling this warning, you can find old forum discussions that suggest a bug in (some versions of) MSVC prevented this from being fully ignored. I tried adding a linker flag to do this, but to no effect.

MxAtomId is referenced in MxEntity which is the ancestor for a lot of the game actor types (e.g. Pizzeria, LegoJetskiRaceActor, and many more). So this means we pull in all the symbols for the AtomId set into these files, which is the reason for all the new warning spam. I don't know if there's a great way to avoid this because MxOmni and MxAtomId need to refer back to each other. I guess if we find that this slows down the build dramatically or introduces another problem then we should address it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,7 @@ add_library(lego1 SHARED
   LEGO1/legoworldpresenter.cpp
   LEGO1/motorcycle.cpp
   LEGO1/mxatomid.cpp
+  LEGO1/mxatomidcounter.cpp
   LEGO1/mxaudiopresenter.cpp
   LEGO1/mxautolocker.cpp
   LEGO1/mxbackgroundaudiomanager.cpp
--- a/LEGO1/compat.h
+++ b/LEGO1/compat.h
@@ -14,6 +14,11 @@
 // DIsable "nonstandard extension used : 'bool'" warning spam
 #pragma warning( disable : 4237 )
 
+// Disable "identifier was truncated to '255' characters" warning.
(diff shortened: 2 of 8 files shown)


itsmattkc:
Looks good to me, thanks!

PR #110, opened 2023-08-13
[workflow] Skip upload step for forks

disinvite:
This is a minor thing, but builds on the master branch fail for all the forks. The reason is that only the main repo has the creds required to upload the progress report to legoisland.org. There's no reason to even attempt this for a fork, so just skip this step if we are on master branch but not isledecomp/isle.


ItzSwirlz:
I feel like its helpful for people on mobile who want to see the progress of a PR

PR #113, opened 2023-08-18
The IslePathActor's

ItzSwirlz:
And ideally, as much of a basic implementation for all of then. As of opening this PR, I'm too afraid to go into Helicopter and I think another one that has to do with Matrixes, which I'm a bit scared of.. so I'll come back to that. A few classes are also missing IsA's and ClassName's

### DIFF
--- a/LEGO1/pizza.cpp
+++ b/LEGO1/pizza.cpp
@@ -13,7 +13,7 @@ Pizza::Pizza()
   this->m_unk88 = 0;
   this->m_unk8c = -1;
   this->m_unk98 = 0;
-  this->m_unk90 = 0;
+  this->m_unk90 = 0.0;
 }
 
 // OFFSET: LEGO1 0x10038100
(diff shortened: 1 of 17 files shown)


itsmattkc:
While you're filling out class structures, you may as well add `DECOMP_SIZE_ASSERT`s to make sure everything is coming out right.

ItzSwirlz:
My bad, I'll do that right now.

I wonder if for the things that are built/have more pieces like the Ambulance, or things that may be built in game, if some fields are certain classes for just plain Bricks - so like when building it, the actor/entity/whatever has one brick together done and not another

LEGO1/pizza.cpp
```diff
@@ -13,7 +13,7 @@ Pizza::Pizza()
   this->m_unk88 = 0;
   this->m_unk8c = -1;
   this->m_unk98 = 0;
-  this->m_unk90 = 0x80000000;
+  this->m_unk90 = 0.0;
```
itsmattkc, line 16: I have a feeling this isn't a float. `0.0f` would ordinarily resolve to `0x0` in hex. The `80000000` could be valid for _negative_ `0.0f`, but that would be an extremely unusual thing for a developer to program intentionally. I think leaving this as a U32 is probably more accurate.

foxtacles (approved):
Fixed the `TowTrack` structure, the rest looks good to me!

PR #117, opened 2023-08-27
Some functions in MxPresenter

disinvite:
A few functions from MxPresenter.

### Notes:

* Reordered MxPresenter functions by their offset, hence the diff being what it is. I brought this up on the matrix chat. If we should go about this in a different way, let me know.

* FUN_100b5120 and FUN_100b7050 match but aren't named yet. This is an area where I'd like some insight if anyone has it. Particularly in the case of FUN_100b7050: it's pretty clear what it's doing, but some context on what kind of structure this is parsing (i.e. which file from the game, if it is a file) would help to name it. It looks like it's used in more places than just MxPresenter so I put it outside the class.

* I used global variables for some of the string literals here. I'm not 100% sure this is right but it matches the "pointer to a pointer" way this is structured in the original.

* Intentional switch/case fallthrough on MxPresenter::Tickle.

* MxParam now points at MxCore instead of the temporary reference

* Added some getter/setters for private members to MxDSAction

* Stubbed out more of the vtable in MxOmni because ParseExtra calls one. Named the functions that have a name in the shared ghidra project

### DIFF
--- a/LEGO1/mxdsaction.h
+++ b/LEGO1/mxdsaction.h
@@ -40,6 +40,12 @@ public:
   virtual MxLong GetSomeTimingField(); // vtable+3c;
   virtual MxLong GetCurrentTime(); // vtable+40;
 
+  inline char *GetUnkData() { return m_unkData; }
+  inline MxU16 GetUnkLength() const { return m_unkLength; }
+  inline void SetOmni(MxOmni *p_omni) { m_omni = p_omni; }
+  inline DWORD GetFlags() const { return m_flags; }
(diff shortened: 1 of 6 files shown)


disinvite:
Mostly obsoleted by #119. I'll resubmit the unique stuff in its own PR.

foxtacles:
@disinvite Sorry, I somehow missed this PR, I should have submitted against this one

disinvite:
@foxtacles No worries! Yours looks a little more polished. There's no real way to coordinate this stuff anyway

PR #118, opened 2023-08-29
Match SetUserMaxLOD

MishaProductions:
Match RealtimeView::SetUserMaxLOD

### DIFF
--- a/LEGO1/realtimeview.cpp
+++ b/LEGO1/realtimeview.cpp
@@ -1,10 +1,18 @@
 #include "realtimeview.h"
+#include <math.h>
+
+// 0x10109598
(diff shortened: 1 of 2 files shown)


LEGO1/realtimeview.cpp
```diff
@@ -1,4 +1,11 @@
 #include "realtimeview.h"
+#include <math.h>
+
+// 0x10109598
+float g_userMaxLodPower;
```
itsmattkc, line 5: Is it correct for this to be uninitialized?

MishaProductions: Yep, that's correct. Wonder why they left it uninitialized. 
image

PR #120, opened 2023-09-08
[Waiting/On Hold] Create a tool to preprocess a file for decomp.me

ItzSwirlz:
Small script to automatically call CL.EXE with /EP /P (file). In the future, it might be nice to clean up the output file

----
**Originally:**
Read the comments in the file for details, there is a lot of nuances, especially because of the include calls. The idea at least is that this will give the basic framework of an average class - so I could import a function from say, LegoEntity, and I will already have everything I need prepared. For a function in a class that inherits LegoEntity, copy the context, manually add that class to the context, and then go. As the decomp progresses this could be updated, depending on how crazy things get, but this should be nice for the many kinds of -Presenter's, -Entity's, etc.


ItzSwirlz:
> Is this something that necessarily needs to be included as part of the `isle` repository? It appears to me the Python script is merely a thin wrapper around `cl.exe`. Is there a reason why someone would not be able or willing to run `cl.exe` directly?

Other repositories have scripts that automatically would make context. For our compiler, we're lucky since it has a preprocessor. But it does have a lot of stuff and we still need to figure out how to make it of a good size- as in, windows.h is built in, not to preprocess windows structs and stuff. Its great because of DirectX

> What does this script enable with regards to [decomp.me](https://decomp.me/) precisely? It would be helpful to learn about the motivation and purpose of this for those not familiar with the platform. Is the aim to establish [decomp.me](https://decomp.me/) support for `isledecomp/isle`?

Alright - so you can import the asm into decomp.me but with context, you can actually tell it about structs and stuff. The more context, the more it can help you. Think of it as one big header file - for Ghidra, the context is constantly changing because everything is all managed by it automatically.

When importing into decomp.me you're saying "I know what this function does, I just need to get the assembly code to match". Ghidra can be used to still help- the idea is, as you change the code, it shows you live how the assembly changes which helps you work closer towards a 100% match.

As an example of why Structs are important, see https://decomp.me/scratch/JqHGp - the Gfx* is necessary as it makes it easy to just copy and paste the match into the code file. This is important when coming to things like accessing fields; the context allows the compiler to know what fields exist at what offset, all that stuff

Decomp.me currently does not have support for DirectX 5. There is a PR for DX8 ongoing, but even with DX5, the context needs to be made so that:
- The order of the definitions do not cause redefiniton
- The order of the definitions do not cause 'undefined things' (MxCore being defined after a class already tries to inherit it)
- The overall context is sufficient enough but short so it doesn't bombard the compiler.

You could make it by hand, but one automatically made for that file is what works best. While windows.h structs are there, DirectX structs aren't. So even if DX5 support is added, this script can automatically make that context, but the windows.h and directx headers need to be replaced with `#import <windows.h>` or whatever accordingly, and also get rid of the thousands of blank empty newlines it causes.

So while you could do it manually, the context not only helps you with the asm matching but also keeps everything in the repo up to date- so you're not trying to figure out a struct that already exists and try to find names for stuff.

ItzSwirlz:
I found by passing `/u` to the compiler, it won't generate the extra windows.h crap. It will only generate DX structures that are necessary; however by doing this it won't spit out MxCore and stuff like that. But it is short enough to let the compiler work with it: https://decomp.me/scratch/8ehhi

This is a solution, but you will still need to copy and paste some structures. For now the best thing to do is wait for decamp.me to have DX5 support upstream, and then with the preprocessor, figure out where the end of windows.h/dx5 header files code is, and replace it with single line `#include` calls.

ItzSwirlz:
decomp.me now supports DirectX 5 - the script should be a 'shortcut' for preprocessing a file easily. https://github.com/decompme/decomp.me/pull/877

What will come of this PR is seeing if there is a way to automatically delete the lines pertaining to Win32/DX5 structures, replace them with `#include` macros, and get rid of extra blank lines

PR #121, opened 2023-09-11
Init functions in MxMediaPresenter and LegoPalettePresenter

ItzSwirlz:


LEGO1/legopalettepresenter.h
```diff
@@ -27,6 +27,7 @@ class LegoPalettePresenter : public MxVideoPresenter
 private:
   void Init();
 
+  MxS32 m_unk64;
```
foxtacles, line 30: Are you sure this is a signed integer? If not, preferably use `undefined4` as the type

ItzSwirlz: Ghidra says it's an "int"

foxtacles: Ghidra's decompiler makes a best effort guess (like with everything else) but often doesn't get the types right. Unless you can actually confirm the type by looking at how the variable is used in code/assembly I'd suggest leaving it as `undefined4`

foxtacles (approved):

PR #122, opened 2023-09-12
Add MxList, MxPresenterList, add to MxMediaManager

foxtacles:
The pattern used for `MxList` (and subsequently `MxPresenterList`) is fairly similar to the one used for `MxHashTable`/`MxVariableTable` so I could re-use a fair amount of @disinvite 's code. `MxList` isn't complete yet, it looks like it will also have a `MxListCursor` similarly to `MxHashTableCursor`.

`MxMediaManager` contains the `MxPresenterList` and a function to initialize it, both are implemented and matching.


itsmattkc:
Seems good to me, thanks!

PR #123, opened 2023-09-12
MxPresenter::ParseExtra

disinvite:
Refactor of #117.

MxPresenter::ParseExtra almost matches, except for the length variable getting optimized away into a register. I'm also not sure why the length is clamped to 0xFFFF before the memcpy but this happens in a few other places, too.

I didn't have a name for "KeyValueStringParse" in the previous PR. It is a sort of general purpose function, but one of these purposes is to get values out of strings from MxOb chunks.

Added stubs for the other MxOmni vtable functions to get a better match on ParseExtra. The names (where there are names) came from the ghidra server.

The 0x3c member of MxPresenter seems to point at another MxPresenter (based on the vtable pointers you get in the debugger) so I set its type.

### DIFF
--- a/LEGO1/mxdsaction.h
+++ b/LEGO1/mxdsaction.h
@@ -49,6 +49,9 @@ public:
 
   inline MxU32 GetFlags() { return this->m_flags; } 
   inline void SetFlags(MxU32 m_flags) { this->m_flags = m_flags; }
+  inline char *GetUnkData() { return m_unkData; }
+  inline MxU16 GetUnkLength() const { return m_unkLength; }
+  inline void SetOmni(MxOmni *p_omni) { m_omni = p_omni; }
 
 private:
   MxU32 m_sizeOnDisk;
--- a/LEGO1/mxdsaction.h
+++ b/LEGO1/mxdsaction.h
@@ -13,6 +13,7 @@ public:
   enum
   {
     Flag_Enabled = 0x20,
+    Flag_Parsed = 0x80,
   };
 
   __declspec(dllexport) MxDSAction();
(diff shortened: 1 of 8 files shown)


LEGO1/mxpresenter.cpp
```diff
@@ -1,26 +1,78 @@
...
+      char t_token[256];
+      strcpy(t_token, token);
+
+      token = strtok(NULL, g_parseExtraTokens);
+      int val = token ? atoi(token) : 0;
+
+      int result = MxOmni::GetInstance()->vtable0x30(t_token, val, this);
+      
+      // TODO: magic number for flag
+      m_action->SetFlags(m_action->GetFlags() | 128);
```
foxtacles, line 55: Maybe `128` can be added to the flags enum in `MxDSAction`? It's still unknown but we could call it `Flag_unk128` or something

LEGO1/mxdsaction.h
```diff
@@ -49,6 +49,9 @@ class MxDSAction : public MxDSObject
 
   inline MxU32 GetFlags() { return this->m_flags; } 
   inline void SetFlags(MxU32 m_flags) { this->m_flags = m_flags; }
+  inline char *GetUnkData() { return m_unkData; }
```
foxtacles, line 52: I think we can rename the variables, functions etc. to `m_extraData`, `GetExtraData` etc. - it seems clear to me now this is just the "extra" data component of the SI files, also referenced in @itsmattkc newest YouTube video (which isn't public yet but accessible to patterns)

LEGO1/mxpresenter.cpp
itsmattkc: Hex is probably preferable too for `or` flags like these

disinvite: I took a leap here and named the flag "Parsed" since we are setting it in the ParseExtra function

LEGO1/mxdsaction.h
disinvite: (Spoiler alert!) I saw that name "extra" in the SIEdit repo earlier too. I changed it all throughout MxDSAction.

foxtacles (approved):
Looks good, thank you!

PR #124, opened 2023-09-13
Implement AddPresenter, RemovePresenter, StopPresenters

foxtacles:
Depends on https://github.com/isledecomp/isle/pull/122

Implementation of the remaining virtual functions in `MxMediaManager` (`AddPresenter`, `RemovePresenter`, `StopPresenters`) which require the `MxListCursor` base class.

`AddPresenter`: 100% match
`RemovePresenter`: 100% match
`StopPresenters`: 100% match


itsmattkc:
Merging #122 seems to have caused a merge conflict. Otherwise this seems good to me.

PR #125, opened 2023-09-15
MxAudioManager: Destructor, LockedReinitialize, Reinitialize

ItzSwirlz:


LEGO1/mxaudiomanager.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// OFFSET: LEGO1 0x100b8df0
+void MxAudioManager::Init()
+{
+  this->m_unk2c = 100;
+}
+
+// OFFSET: LEGO1 0x100b8e00
+void MxAudioManager::LockedReinitialize(MxS8 p_skipTeardown)
```
foxtacles, line 22: This looks like it should be a `MxBool`

LEGO1/mxaudiomanager.h
```diff
@@ -0,0 +1,21 @@
...
+#include "mxmediamanager.h"
+
+// VTABLE 0x100dc6e0
+class MxAudioManager : public MxMediaManager
+{
+public:
+  MxAudioManager();
+  virtual ~MxAudioManager() override;
+
+  void LockedReinitialize(MxS8);
```
foxtacles, line 13: Looks like an internally used function to me, so I'd make it `private` unless there is anything that indicates otherwise

LEGO1/mxaudiomanager.h
```diff
@@ -0,0 +1,21 @@
...
+  void Reinitialize();
+protected:
+  void Init();
+
+  int m_unk2c;
```
foxtacles, line 18: While we don't have any code that actually uses it or idea of its purpose it would be good to declare it as `undefined4` type

foxtacles (approved):
Applies fixes and added some missing stuff, thank you!

PR #126, opened 2023-09-16
Implement/match MxPresenter::StartAction

foxtacles:
There is a very minor diff, which might be related to how the location floats are accessed, but otherwise the function matches. The remaining unknown members of `MxPresenter` were coordinates (consisting of `MxPoint32`, introduced with this PR; and a single `Mx32`)

`MxPresenter::Init` also matches now.

### DIFF
--- a/LEGO1/mxdsaction.h
+++ b/LEGO1/mxdsaction.h
@@ -52,6 +52,7 @@ public:
   inline void SetFlags(MxU32 m_flags) { this->m_flags = m_flags; }
   inline char *GetExtraData() { return m_extraData; }
   inline MxU16 GetExtraLength() const { return m_extraLength; }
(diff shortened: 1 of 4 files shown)


itsmattkc:
This is probably an obvious question, but are we sure the Z isn't inside a class/struct like the X and Y are?

foxtacles:
> This is probably an obvious question, but are we sure the Z isn't inside a class/struct like the X and Y are?

Yeah, I'm very certain it's not part of the class. My first approach was to have all 3 coordinates be part of the class, but the code generation was quite off from the original. Then I noticed that assignment/handling of the third value is actually markedly different in the assembly and figured it must be a separate member, which got us the right assembly in both `StartAction` and `Init`.

itsmattkc:
Understood, thanks for clarifying. Looks good, will merge

PR #129, opened 2023-09-19
LegoOmni::CreateStreamObject and related

disinvite:
- 100% match on LegoOmni::CreateStreamObject. 
- Near match on MxDSObject::DeserializeDSObjectDispatch. Instruction order and the jump table are throwing this one off.
- Aesthetic change to MxDSFile::Read to improve clarity. Changed the params to match our style guide.
- Removed some unneeded casts to (char*) when calling MXIOINFO::Read
- Added size asserts and dummy members to children of MxDSAction. This was needed for DeserializeDSObjectDispatch.
- Added MxDSSource::GetBuffer. The buffer type is currently (void*), so there's a slightly ugly cast when it gets called in CreateStreamObject.

### DIFF
--- a/LEGO1/mxdsfile.cpp
+++ b/LEGO1/mxdsfile.cpp
@@ -47,13 +47,13 @@ MxLong MxDSFile::Open(MxULong uStyle)
 }
 
 // OFFSET: LEGO1 0x100cc780
-MxLong MxDSFile::Read(unsigned char *pch, MxULong cch)
+MxResult MxDSFile::Read(void *p_buf, MxULong p_nbytes)
 {
-  if (m_io.Read((char*)pch, cch) != cch)
-    return -1;
+  if (m_io.Read(p_buf, p_nbytes) != p_nbytes)
+    return FAILURE;
 
-  m_position += cch;
-  return 0;
+  m_position += p_nbytes;
+  return SUCCESS;
 }
 
 // OFFSET: LEGO1 0x100cc620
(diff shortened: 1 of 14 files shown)


LEGO1/mxdsfile.cpp
```diff
@@ -47,13 +47,13 @@ MxLong MxDSFile::Open(MxULong uStyle)
...
 {
-  if (m_io.Read((char*)pch, cch) != cch)
-    return -1;
+  if (m_io.Read(p_buf, p_nbytes) != p_nbytes)
+    return FAILURE;
 
-  m_position += cch;
-  return 0;
+  m_position += p_nbytes;
+  return SUCCESS;
```
foxtacles, line 56: Maybe some of the other functions in `MxDSFile` can also be changed to use the result constants and `MxResult` instead of `MxLong`?

foxtacles (approved):

LEGO1/mxdsfile.cpp
disinvite: I thought about doing that before, and I just looked again and remembered why I didn't. It's this line in Open():
https://github.com/isledecomp/isle/blob/830115044d18fd00ed33a4b821cc8b286c8a055e/LEGO1/mxdsfile.cpp#L25
`SUCCESS + 1`?

foxtacles: Yeah, I think that's the only function in the class that seems to use a value that isn't `SUCCESS` or `FAILURE`. `ReadChunk` and `Close` use only those known values. I guess we should leave it as-is for now.

foxtacles:
Looks good, thank you!

PR #131, opened 2023-09-20
MxDiskStreamProvider constructor

ItzSwirlz:
just something random i started doing, i guess. 63% matching


foxtacles:
I've added work-in-progress structures which I believe to be a list and node. There's still a piece missing to it so the constructor doesn't quite match yet, but it's close.

PR #134, opened 2023-09-20
Improve match for MxMatrix::EqualsDataProduct

disinvite:
Quick one in MxMatrix. The for loop variables need to be optimized away for this to match. To do that, I used a pointer to update m_data and then just incremented it directly rather than relying on the `row * 4 + col` calculation each time.

It seems to work, except for the usual business of the wrong registers getting used. Also, I'm not sure why the compiler thinks that the the first param is `esp + 0x18` instead of `+ 0x10`. You can see how this gets called in the neighbor function EqualsMxProduct, and there are only two dword-sized params.

### DIFF
--- a/LEGO1/mxmatrix.cpp
+++ b/LEGO1/mxmatrix.cpp
@@ -111,21 +111,20 @@ void MxMatrix::EqualsMxProduct(const MxMatrix *p_a, const MxMatrix *p_b)
   EqualsDataProduct(p_a->m_data, p_b->m_data);
 }
 
-// Just a placeholder matrix multiply implementation. I think the decomp will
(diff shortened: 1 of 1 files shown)


foxtacles:
> Also, I'm not sure why the compiler thinks that the the first param is `esp + 0x18` instead of `+ 0x10`. You can see how this gets called in the neighbor function EqualsMxProduct, and there are only two dword-sized params.

I believe this is simply due to the fact that the first parameter is loaded in between pushes in the original code, whereas in our decomp both params are loaded afterwards:

image

The code is correct imo, I'm pretty sure we'll randomly see a 100% match in the future.

PR #137, opened 2023-09-22
PresenterNameDispatch

disinvite:
Small function that will either: 
- pass through the source name of an MxDSAction (if present)
- return the name of the MxPresenter subclass to be used for the given action

Some interesting syntax here, but a good match.

### DIFF
--- a/LEGO1/mxpresenter.cpp
+++ b/LEGO1/mxpresenter.cpp
@@ -158,23 +158,23 @@ void MxPresenter::Enable(MxBool p_enable)
 }
 
 // OFFSET: LEGO1 0x100b5310
-char *PresenterNameDispatch(MxDSAction *p_action)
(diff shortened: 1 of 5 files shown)


LEGO1/mxpresenter.cpp
```diff
@@ -155,6 +157,63 @@ void MxPresenter::Enable(MxBool p_enable)
   }
 }
 
+// OFFSET: LEGO1 0x100b5310
+char *PresenterNameDispatch(MxDSAction *p_action)
```
foxtacles, line 161: Maybe we should use a reference instead of a pointer as an argument in this case? We won't really know what the developers used originally since it should compile to the same assembly, so it's up to us to make a decision here. We should probably establish a guideline we want to follow for the decomp.

foxtacles:
Looks good, thank you!

PR #140, opened 2023-09-25
Fix LegoOmni vtable

itsmattkc:
LegoOmni and MxOmni share some vtable functions, but some changes in MxOmni caused shifts in LegoOmni's table.

We should probably have a script that ensures correct vtables and their addresses. I'm not sure how to pull that info from the PDB yet but it's got to be in there.


foxtacles:
Looks like the bug was introduced in #123 - sorry, I missed that in my review. I agree that additional tooling regarding vtables would be great.

PR #141, opened 2023-09-26
LegoEntity parsing extra Action strings

disinvite:
Two new functions and a few changes all related to parsing the Extra string from an MxOb chunk. Specifically: parsing the "Action" section. For example:

```"Action:openram;\lego\scripts\Race\CarRaceR;0"``` (from CARRACE.SI)

Action is the key we are interested in; the rest is the value, with ';' as the delimiter. The first section is the action type, and a new function MatchActionString converts this to an int (enum) that I put in the new file extra.h.

There's no obvious name for the second and third tokens and they don't appear (or are ignored) for some action types. I named these m_actionArgString and m_actionArgNumber for now. The new function LegoEntity::ParseAction sets these variables.

I commented out the size assert on MxEntity (or we could just remove it). It felt more correct to have these new member values defined in LegoEntity instead of MxEntity, because they are set in the PasrseAction function that is part of LegoEntity's vtable only. There isn't an obvious break between which members should go in MxEntity or LegoEntity and we probably can't say for certain until all the MxEntity functions are decomped so we know what things it uses.

Lastly, I put another string literal as a global variable (where I last did so in #123 for MxPresenter). The asm doesn't match unless we push a pointer to the string's address (versus the address where the characters start). I also borrowed the ParseExtra tokens from MxPresenter. Is there a good practice to follow for how and where to define strings like these so we don't have to move them later? Is it not actually required to make these global vars?

**Updates after review**
- Modified size assert is restored to MxEntity
- Moved global string variables (including the ones from #123) to their own define.cpp file
- Minor change to KeyValueStringParse to use const char* args where it makes sense

### DIFF
--- a/LEGO1/legoentity.h
+++ b/LEGO1/legoentity.h
@@ -33,7 +33,7 @@ public:
   }
 
   virtual void vtable18(); // vtable+0x18
-  virtual void Destroy() override; // vtable+0x1c
+  virtual void Destroy(); // vtable+0x1c
   virtual void ParseAction(char *); // vtable+0x20
 
 protected:
--- a/LEGO1/legoentity.h
+++ b/LEGO1/legoentity.h
@@ -2,6 +2,7 @@
 #define LEGOENTITY_H
 
 #include "mxentity.h"
+#include "extra.h"
 
 // VTABLE 0x100d4858
 // SIZE 0x68 (probably)
@@ -31,7 +32,16 @@ public:
     return !strcmp(name, LegoEntity::ClassName()) || MxEntity::IsA(name);
   }
 
+  virtual void vtable18(); // vtable+0x18
   virtual void Destroy() override; // vtable+0x1c
+  virtual void ParseAction(char *); // vtable+0x20
+
+protected:
+  // For tokens from the extra string that look like this:
+  // "Action:openram;\lego\scripts\Race\CarRaceR;0"
+  ExtraActionType m_actionType; // 0x5c
+  char *m_actionArgString; // 0x60
+  MxS32 m_actionArgNumber; // 0x64
 
 };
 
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: LEGO1/legoentity.h [hunk 1]
```diff
@@ -31,7 +32,16 @@ class LegoEntity : public MxEntity
     return !strcmp(name, LegoEntity::ClassName()) || MxEntity::IsA(name);
   }
 
+  virtual void vtable18(); // vtable+0x18
   virtual void Destroy() override; // vtable+0x1c
```
foxtacles, line 36: This is a new virtual function introduced in `LegoEntity`, so we have to remove the `override` keyword.

foxtacles:
Looks good to me, pointed out one minor issue above.

WIth regards to the global strings, in my opinion the cleanest option for now would be to have one compilation unit that aggregates all those externally used strings that are used in multiple files, similar to how it is in `ISLE`: https://github.com/isledecomp/isle/blob/master/ISLE/define.h - otherwise it would just come down to a random decision on where we put them and I find that a bit messy.

LEGO1/mxentity.cpp
```diff
@@ -1,6 +1,6 @@
 #include "mxentity.h"
 
-DECOMP_SIZE_ASSERT(MxEntity, 0x68)
+// DECOMP_SIZE_ASSERT(MxEntity, 0x68)
```
foxtacles, line 3: I think we should leave this in place and just adjust it to specify the currently assumed size (with a comment that it might still change in the future), so we can still prevent any silent unintended changes to the size of `MxEntity` in the future.

disinvite:
Thanks! I like the define.cpp idea, so I added it. Had to make a minor change to the signature for the key-value parse function so we could use const char*.

PR #142, opened 2023-09-26
Bootstrap MxDSSelectAction, generalize MxList

foxtacles:
`MxDSSelectAction` has another list, apparently based on `MxList`. The interesting aspect here is that the list entries hold `T` (`MxString` in this case) instead of `T*`, so I generalized `MxList` to accept `T` so it works with both pointer and value types.

Unsurprisingly, the code generation is quite different when using a non-primitive `T` such as `MxString` (and not very efficient due to needless copying of `T` values in many places), and not all functions end up inlined. Initially I assumed they must have used a different kind of list implementation, but so far it looks like it is the same.

To verify correct code generation, I added some template annotations for the crucial functions. So far it all matches, which makes me relatively confident this list class actually shares the same code with the previous pointer lists.


foxtacles:
Merging since I'm fairly confident this is accurate and it's a prerequisite for more PRs coming up.

PR #143, opened 2023-09-29
implement some of MxTransitionManager

Ramen2X:
Implements some of MxTransitionManager and the global function SetAppCursor.

SetAppCursor and MxTransitionManager::GetDDrawSurfaceFromVideoManager match 100%.

MxTransitionManager::StartTransition is very incomplete and does not match, with only 17.28%.

### DIFF
--- a/LEGO1/legoinputmanager.h
+++ b/LEGO1/legoinputmanager.h
@@ -28,10 +28,26 @@ public:
 
   virtual MxResult Tickle() override; // vtable+0x8
 
-  undefined m_pad40[0x15c];
+  undefined m_pad40[0x48];
+
+  MxBool m_unk88;
+  MxU8 m_unk89;
+  MxU8 m_unk8a;
+  MxU8 m_unk8b;
+
+  undefined m_pad8c[0x110];
+
+  // 0x19C
   int m_joystickIndex;
-  undefined m_pad200[0x194];
+
+  undefined m_pad1a0[0x194];
+
+  // 0x334
   MxBool m_useJoystick;
+
+  MxU8 m_unk335;
+  MxBool m_unk336;
+  MxU8 m_unk337;
 };
 
 #endif // LEGOINPUTMANAGER_H
--- a/LEGO1/legoinputmanager.h
+++ b/LEGO1/legoinputmanager.h
@@ -31,11 +31,7 @@ public:
   undefined m_pad40[0x48];
 
   MxBool m_unk88;
-  MxU8 m_unk89;
-  MxU8 m_unk8a;
-  MxU8 m_unk8b;
-
-  undefined m_pad8c[0x110];
+  undefined m_unk89[0x113];
 
   // 0x19C
   int m_joystickIndex;
@@ -45,9 +41,9 @@ public:
   // 0x334
   MxBool m_useJoystick;
 
-  MxU8 m_unk335;
+  undefined m_unk335;
   MxBool m_unk336;
-  MxU8 m_unk337;
+  undefined m_unk337;
 };
 
 #endif // LEGOINPUTMANAGER_H
--- a/LEGO1/legoomni.h
+++ b/LEGO1/legoomni.h
@@ -21,6 +21,7 @@
 #include "legoplantmanager.h"
 
 class LegoSoundManager;
+class MxTransitionManager;
 
 // VTABLE 0x100d8638
 // SIZE: 0x140
--- a/LEGO1/mxomni.cpp
+++ b/LEGO1/mxomni.cpp
@@ -289,3 +289,9 @@ MxEventManager* EventManager()
 {
   return MxOmni::GetInstance()->GetEventManager();
 }
+
+// OFFSET: LEGO1 0x1003ef40
+void SetAppCursor(WPARAM p_wparam)
+{
+    PostMessageA(MxOmni::GetInstance()->GetWindowHandle(), 0x5400, p_wparam, 0);
+}
--- a/LEGO1/mxtransitionmanager.cpp
+++ b/LEGO1/mxtransitionmanager.cpp
@@ -25,3 +25,47 @@ void MxTransitionManager::SetWaitIndicator(MxVideoPresenter *videoPresenter)
 {
   // TODO
 }
+
+// OFFSET: LEGO1 0x1004baa0
+MxResult MxTransitionManager::GetDDrawSurfaceFromVideoManager() // vtable+0x14
+{
+    LegoVideoManager *videoManager = VideoManager();
+    this->m_ddSurface = videoManager->GetDisplaySurface()->GetDirectDrawSurface2();
+    return SUCCESS;
+}
+
+// OFFSET: LEGO1 0x1004bb70
+MxResult MxTransitionManager::StartTransition(TransitionType p_animationType, MxS32 p_speed,
+                                              MxU8 p_unk, MxBool p_playMusicInAnim)
+{
+    // TODO: Incomplete and far from matching
+
+    if (this->m_transitionType == NOT_TRANSITIONING) {
+        if (!p_playMusicInAnim) {
+            MxBackgroundAudioManager *backgroundAudioManager = BackgroundAudioManager();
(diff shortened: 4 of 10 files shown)


LEGO1/mxtransitionmanager.h
```diff
@@ -2,19 +2,58 @@
...
-  MxTransitionManager();
-  virtual ~MxTransitionManager() override; // vtable+0x0
+    MxTransitionManager();
+    virtual ~MxTransitionManager() override; // vtable+0x0
 
-  __declspec(dllexport) void SetWaitIndicator(MxVideoPresenter *videoPresenter);
+    __declspec(dllexport) void SetWaitIndicator(MxVideoPresenter *videoPresenter);
 
-  virtual MxResult Tickle(); // vtable+0x8
+    virtual MxResult Tickle(); // vtable+0x8
```
foxtacles, line 19: We are currently using 2-space indents in the project as per the Styleguide, could you update the code? We should probably introduce a linter (and a check in the CI) so stuffokes that becomes easier to handle but for now we have to do it manually.

LEGO1/mxtransitionmanager.h
```diff
@@ -2,19 +2,58 @@
...
 
 #include "mxcore.h"
+#include "mxvideopresenter.h"
+#include "legovideomanager.h"
+#include "legoomni.h"
 
-class MxVideoPresenter;
 
-// 0x100d7ea0
+// VTABLE 0x100d7ea0
```
foxtacles, line 10: Could you add the current size of the class and a `DECOMP_SIZE_ASSERT`? Thank you!

LEGO1/mxtransitionmanager.h
```diff
@@ -2,19 +2,58 @@
 #define MXTRANSITIONMANAGER_H
 
 #include "mxcore.h"
+#include "mxvideopresenter.h"
```
foxtacles, line 5: Based on a cursory glance, I think some of those includes can be moved to the `mxtransitionmanager.cpp` unit to cut down on dependencies.

LEGO1/mxtransitionmanager.cpp
```diff
@@ -25,3 +25,47 @@ void MxTransitionManager::SetWaitIndicator(MxVideoPresenter *videoPresenter)
...
+        tickleManager->RegisterClient(this, p_speed);
+
+        LegoInputManager *inputManager = InputManager();
+        inputManager->m_unk88 = TRUE;
+        inputManager->m_unk336 = FALSE;
+
+        MxVideoManager *videoManager = VideoManager();
+        videoManager->SetUnkE4(FALSE);
+
+        SetAppCursor(1);
```
foxtacles, line 67: Maybe we should start introducing shared named constants / enums for those values, since it looka like it interacts with `ISLE`: https://github.com/isledecomp/isle/blob/master/ISLE/isleapp.cpp

LEGO1/mxomni.cpp
```diff
@@ -289,3 +289,9 @@ MxEventManager* EventManager()
 {
   return MxOmni::GetInstance()->GetEventManager();
 }
+
+// OFFSET: LEGO1 0x1003ef40
+void SetAppCursor(WPARAM p_wparam)
+{
+    PostMessageA(MxOmni::GetInstance()->GetWindowHandle(), 0x5400, p_wparam, 0);
```
foxtacles, line 296: Maybe we should start introducing shared named constants / enums for those values, since it looka like it interacts with `ISLE`: https://github.com/isledecomp/isle/blob/master/ISLE/isleapp.cpp

LEGO1/legoinputmanager.h
```diff
@@ -28,10 +28,26 @@ class LegoInputManager : public MxPresenter
 
   virtual MxResult Tickle() override; // vtable+0x8
 
-  undefined m_pad40[0x15c];
+  undefined m_pad40[0x48];
+
+  MxBool m_unk88;
+  MxU8 m_unk89;
```
foxtacles, line 34: Are those `U8` addressed anywhere? If not, they might just be implicit compiler padding and we shouldn't specify them imo

LEGO1/legoinputmanager.h
```diff
@@ -28,10 +28,26 @@ class LegoInputManager : public MxPresenter
...
+  // 0x19C
   int m_joystickIndex;
-  undefined m_pad200[0x194];
+
+  undefined m_pad1a0[0x194];
+
+  // 0x334
   MxBool m_useJoystick;
+
+  MxU8 m_unk335;
```
foxtacles, line 48: We should use `undefined` unless you have a good reason to believe those are `U8` (you might have, in which case it's fine, but there's no code checked in ATM that uses those)

LEGO1/mxtransitionmanager.h
```diff
@@ -2,19 +2,58 @@
...
+        NOT_TRANSITIONING,
+        NO_ANIMATION,
+        DISSOLVE,
+        PIXELATION,
+        SCREEN_WIPE,
+        WINDOWS,
+        BROKEN // Unknown what this is supposed to be, it locks the game up
+    };
+    
+    MxResult StartTransition(TransitionType p_animationType, MxS32 p_speed, MxU8 p_unk, MxBool p_playMusicInAnim);
```
foxtacles, line 45: Similarly here, using `undefined` for `p_unk` would be cleaner (unless you know more than the code shows)

LEGO1/mxvideomanager.h
```diff
@@ -26,13 +26,17 @@ class MxVideoManager : public MxMediaManager
...
+  inline MxDisplaySurface *GetDisplaySurface() { return this->m_displaySurface; }
+  inline void SetUnkE4(MxBool p_value) { this->m_unke4 = p_value; }
 private:
   MxVideoParam m_videoParam;
   LPDIRECTDRAW m_pDirectDraw;
   LPDIRECTDRAWSURFACE m_pDDSurface;
   MxDisplaySurface *m_displaySurface;
   MxRegion *m_region;
   MxBool m_unk60;
+  undefined m_pad61[84];
```
foxtacles, line 38: Are you sure this is part of `MxVideoManager` and not `LegoVideoManager`? Regardless, we have to update the class size and assert.

LEGO1/legoomni.h
```diff
@@ -21,6 +21,7 @@
 #include "legoplantmanager.h"
 
 class LegoSoundManager;
+class MxTransitionManager;
```
foxtacles, line 24: Looks like this class might need a refactor in terms of includes and forward declarations, looks kind of messy right now.

LEGO1/mxtransitionmanager.h
Ramen2X: Sorry about that, I wasn't aware. Maybe this should be added to the style guide in CONTRIBUTING.MD?

LEGO1/mxvideomanager.h
Ramen2X: I think you're right, thanks. My bad, this was just an erroneous assumption on my part.

LEGO1/mxomni.cpp
Ramen2X: I agree, that would probably be best going forward. Is this something you would like to be addressed in this PR?

LEGO1/mxtransitionmanager.h
```diff
@@ -15,6 +16,43 @@ class MxTransitionManager : public MxCore
...
+    SCREEN_WIPE,
+    WINDOWS,
+    BROKEN // Unknown what this is supposed to be, it locks the game up
+  };
+
+  MxResult StartTransition(TransitionType p_animationType, MxS32 p_speed, undefined p_unk, MxBool p_playMusicInAnim);
+
+private:
+  undefined m_pad00[0x20];
+  MxU8 m_unk20;
```
foxtacles, line 48:
```suggestion
  undefined m_unk20;
```

Probably should be `undefined` as well? We could also merge the paddings.

foxtacles (approved):
Looks good!

PR #144, opened 2023-09-29
Relocate SetAppCursor to legoutil

foxtacles:
Something I missed previously, judging by its original address the function `SetAppCursor` is likely part of the "utility" function collection. This PR moves the function to that compilation unit.


Ramen2X (approved):
LGTM, thanks for catching this!

PR #145, opened 2023-09-29
MxMusicPresenter: ClassName, IsA, ctor, Init

ItzSwirlz:
Did this at school, so unable to test locally but this should all be correct


foxtacles:
Looks good

PR #147, opened 2023-09-29
MxVariable subclass for fsmovie setting

disinvite:
New class LegoFullScreenMovie. It is a subclass of MxVariable and used to toggle whether full screen movies (e.g. Lego and Mindscape smacker videos) will play. If they are disabled, the game shows a black screen instead.

Added another EnableFullScreenMovie() method with a single MxBool parameter. It is a wrapper around the other (DLL public) function of the same name with the second parameter set to true. The second parameter corresponds to the "Play movies at native resolution" option from Rebuilder.

Added some size asserts for the other MxVariable classes.

Partial implementation of LegoGameState constructor to demonstrate using the three MxVariable types.

### DIFF
--- a/LEGO1/define.h
+++ b/LEGO1/define.h
@@ -1,8 +1,6 @@
 #ifndef DEFINE_H
 #define DEFINE_H
 
-extern const char *g_str_enable;
-extern const char *g_str_disable;
 extern const char *g_parseExtraTokens;
 extern const char *g_strWORLD;
 extern const char *g_strACTION;
--- a/LEGO1/define.h
+++ b/LEGO1/define.h
@@ -1,6 +1,8 @@
 #ifndef DEFINE_H
 #define DEFINE_H
 
+extern const char *g_str_enable;
+extern const char *g_str_disable;
 extern const char *g_parseExtraTokens;
 extern const char *g_strWORLD;
 extern const char *g_strACTION;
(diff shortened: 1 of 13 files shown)


LEGO1/legovideomanager.cpp
```diff
@@ -27,8 +27,14 @@ int LegoVideoManager::DisableRMDevice()
   return 0;
 }
 
+// OFFSET: LEGO1 0x1007c300
+void LegoVideoManager::EnableFullScreenMovie(MxBool p_enable)
+{
+  EnableFullScreenMovie(p_enable, 1);
```
foxtacles, line 33: `1` should probably be `TRUE`

LEGO1/legogamestate.h
```diff
@@ -16,6 +19,10 @@ class LegoGameState
 
 private:
   char *m_savePath;
+  undefined m_unk[20];
+  LegoBackgroundColor *m_backgroundColor; // 0x18
```
foxtacles, line 23: Could you add a size assert for this class? Thank you!

LEGO1/define.h
```diff
@@ -1,6 +1,8 @@
 #ifndef DEFINE_H
 #define DEFINE_H
 
+extern const char *g_str_enable;
+extern const char *g_str_disable;
```
foxtacles, line 5: I think we should only put strings here that we know are used across multiple compilation units. It seems those are only used in a single function?

disinvite: Good point, I'll switch this around. (The string "disable" _is_ used in LegoGameState too, but it is referenced directly rather than indirectly as in LegoFullScreenMovie, so it's not a global variable.)

foxtacles (approved):
Looks good, thank you!

PR #148, opened 2023-09-30
(Proposal) Refactor MxOmni/LegoOmni headers to use forward declarations

foxtacles:
This PR changes the `mxomni.h` and `legoomni.h` headers to use forward declarations instead of pulling in the full headers of all the classes. The `MxOmni` and `LegoOmni` class declarations (and partially, their implementations) need not be aware of the declarations that come with those headers, since for the most part they only store/return pointers to them.

I've marked this PR as a proposal to see what your opinions are about this change first, since I'm aware that the use of forward declarations is controversial.

Personally I believe this to be a good change since it cuts down massively on (re)-compile time since it eliminates unnecessary dependencies. The 1996 source also indicates that the original developers most likely have employed this practice in the same fashion.

If this is approved, I'd submit further PRs to get rid of other dependencies, especially ones related to `MxList`, since right now changing anything in `mxlist.h` re-builds a vast amount of objects that don't actually need to be rebuilt.


disinvite:
Looks good to me, but I don't know that I fully understand the tradeoff. Is it just that it becomes (maybe only slightly) more difficult to reason about the dependencies if we use forward refs? Or that the compiler might do something unexpected when resolving dependencies that wouldn't be obvious to detect?

If this eliminates the C4786 warning spam then it's fantastic.

foxtacles:
> Looks good to me, but I don't know that I fully understand the tradeoff. Is it just that it becomes (maybe only slightly) more difficult to reason about the dependencies if we use forward refs? Or that the ~compiler~ linker might do something unexpected when resolving dependencies that wouldn't be obvious to detect?

I believe there shouldn't be any major downsides when we use forward declarations the way I used them here (only forward declare the classes).

One I can think of is that calling `delete` on a pointer to an incomplete type results in undefined behavior. The compiler should emit a warning, but if the warning is ignored or goes unnoticed, it will result in incorrect code (although this should be caught by `reccmp`).

I could also imagine a case where given a template `foo`, a template specialization for type `T` is located in `t.h`. When `t.h` is not included but `T` is forward declared, the compiler will instantiate `foo<T>` based on the generic template, and miss any specializations located in `t.h`. This can theoretically go unnoticed.

Since `reccmp` is our main control mechanism of accuracy I'm not too woried about issues like this however and I'd vote to use forward declarations as outlined in this PR.

itsmattkc:
I'm not necessarily against this in principle. The reason I've tended to just include everything is because of old habits writing in [Google Code Style](https://google.github.io/styleguide/cppguide.html#Forward_Declarations). They recommend against forward declarations for the following reasons:

> - Forward declarations can hide a dependency, allowing user code to skip necessary recompilation when headers change.
> - A forward declaration as opposed to an `#include` statement makes it difficult for automatic tooling to discover the module defining the symbol.
> - A forward declaration may be broken by subsequent changes to the library. Forward declarations of functions and templates can prevent the header owners from making otherwise-compatible changes to their APIs, such as widening a parameter type, adding a template parameter with a default value, or migrating to a new namespace.
> - Forward declaring symbols from namespace std:: yields undefined behavior.
> - It can be difficult to determine whether a forward declaration or a full #include is needed. Replacing an #include with a forward declaration can silently change the meaning of code:
> ```
> // b.h:
> struct B {};
> struct D : B {};
> 
> // good_user.cc:
> #include "b.h"
> void f(B*);
> void f(void*);
> void test(D* x) { f(x); }  // Calls f(B*)
> ```
> - Forward declaring multiple symbols from a header can be more verbose than simply #includeing the header.
> - Structuring code to enable forward declarations (e.g., using pointer members instead of object members) can make the code slower and more complex.

That being said, a lot of these shouldn't apply here. We're not forward declaring anything from `std` or from external libraries, and this is only proposing doing so two headers, so these issues likely aren't a big concern (just thought I'd post them for consideration).

I think probably the biggest concern is that changing a header may not propagate when it should, however since they're all pointers in LegoOmni/MxOmni, that's probably not an issue either? As it stands, I'd say I don't oppose this change.

foxtacles:
Yeah, now I remember it was the Google style guide that explicitly recommended against forward declarations, which is why I wanted to propose this to evaluate if the tradeoff would be worth it.

I agree that in principle I'd also prefer straight up including all the headers to avoid the downsides, however during my workflows I started noticing I've spent considerable time waiting for unnecessary re-compilation of 10-20 units every time when making only minor changes in a single header (`mxlist.h` is a good example, but also in many others). The main culprit are the Omni headers and maybe a few others since they include lots of headers and are themselves also included in many places, leading to that issue.

I think as long as we limit the forward declarations to cases like in this PR it should be manageable and like you said, several of the concerns don't apply. Nevertheless we should always remain cautious and aware of the potential pitfalls, and limit the practice to those areas that are major sources of re-compilation slowdowns.

PR #149, opened 2023-09-30
Fix pizza constructor

ItzSwirlz:
Incorrect type and a missing line


foxtacles:
Removed superfluous padding, applied `undefined` types and added `LegoActor` constructor stub for a 100% match.

PR #152, opened 2023-10-01
LegoEntity::Reset

disinvite:
Added a Reset method to LegoEntity. It zeroes out member values or restores to defaults.

In each place this is called, we are inside of a LegoEntity method or have just created an object where the LegoEntity vtable is set. In #141, we debated how many of the members in the 0x68 sized LegoEntity footprint belong to superclass MxEntity. With this Reset method setting members from 0x10 and up, I feel pretty confident that MxEntity owns just the two members referenced in SetEntityId(). I moved everything over to LegoEntity and updated the size asserts.

The Destroy() method has a parameter (probably a bool based on the callers) but Ghidra (and I) can't see where it's used. Adding this brings the destructor to 100%.

The debugger shows m_unk54 set to an object with vtable 0x100dbe38. The shared Ghidra has this as "VideoStuff::ClassA_D::VTABLE_D" right now.

### DIFF
--- a/LEGO1/legoentity.cpp
+++ b/LEGO1/legoentity.cpp
@@ -9,7 +9,7 @@ DECOMP_SIZE_ASSERT(LegoEntity, 0x68)
 // OFFSET: LEGO1 0x1000c290
 LegoEntity::~LegoEntity()
 {
-  Destroy();
+  Destroy(TRUE);
 }
 
 // OFFSET: LEGO1 0x100114f0 STUB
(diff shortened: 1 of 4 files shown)


foxtacles:
Looks good. If you are curious about the meaning of some of the members and parameters, take a look at the class in the 1996 source:

`LEGO/LEGOOMNI/SRC/ENTITY/LENTITY.CPP`

Of course one always has to be very cautious with taking it as too much of a reference since clearly a lot of code has changed, but it gives reasonable hints. (`m_unk54` appears to be `LegoROI`; and the vectors appear to have a helper function called `Fill` to zero them out, that would certainly look better than we have used so far)

PR #153, opened 2023-10-02
Implement MxTransitionManager ctor/dtor/rest of StartTransition

itsmattkc:
Everything matches except for some pesky registers in StartTransition. I believe these are probably influenced by headers outside of the functions control, so I'm submitting as-is.


foxtacles (approved):
Looks good!

Ramen2X:
Thanks for figuring out the rest of StartTransition!

PR #154, opened 2023-10-02
Implement MxDSSelectAction Clone/operator=/CopyFrom

foxtacles:
Matches:

* `Clone/operator=/CopyFrom`: 100%
* `MxList<MxString>::OtherAppend`: 100%
* `MxListEntry<MxString>::MxListEntry<MxString>`: 100%
* `MxList<MxString>::_InsertEntry`: ~75% (apparently one extra local variable is used, haven't figured it out yet, but functionally should be accurate)

This PR highlights the importance of the template annotations, since several functions of `MxList` are *not* inlined (due to the value copies of `MxString`) and thus not covered by the accuracy comparison of the caller functions.

I'm not sure why they have used a different kind of "append" function here. Initially I wanted to specialize `Append(MxString T)` but apparently explicit template specialization is not supported in this version of VC++. Instead, I've created a distinct function called `OtherAppend` (same interface). That most likely had a different name / use case in the original code base, but I haven't figured it out yet so it will probably change in the future.

I've had to change the assignment operator interface of `MxString` to take a reference instead of a pointer (the match stays the same obviously, I'm not sure why it was a pointer in the first place):

`MxString &MxString::operator=(MxString *param)` is now
`MxString &MxString::operator=(MxString &param)`

I've also added the `override` keyword to the virtual functions in the `MxDS*` class hierarchy where needed.


itsmattkc:
All looks good to me. The only thing I can think of regarding the OtherAppend function is perhaps it's an "Insert" function that they just happen to be using the last entry to insert after in this instance? It would be unusual in the context it's used in but that's my best guess.

Otherwise I think this is good to merge!

foxtacles:
> The only thing I can think of regarding the OtherAppend function is perhaps it's an "Insert" function that they just happen to be using the last entry to insert after in this instance? It would be unusual in the context it's used in but that's my best guess.

Yeah, that's what I'm thinking as well. 

Something else I'm still contemplating is that perhaps they *did* use a different list implementation for storing value types like `MxString` in this case. The 1996 source explicitly refers to `MxPtrList` many times (implying a list containing pointers) so maybe they just duplicated the list implementation (which would explain why they share almost all code - the reason I'm currently sharing the template) but made a few adjustements.

EIther way, hopefully this will become clearer over time as we keep using more of `MxList`.

PR #155, opened 2023-10-02
MxTransitionManager::Dissolve

disinvite:
Dissolve transition, as selected by the Tickle() method. The code comments give my reasoning for naming the two array members from the class. I tried to figure out `FUN_1004c4d0` and `FUN_1004c580` (called by each of the transition functions) to at least give a name, but they both depend on pieces we don't have yet. `FUN_1004c4d0` is doing a memcpy on something. Maybe part of page flipping?

The pixel plot for the "two bytes per pixel" case doesn't match exactly. There is either some subtlety in the casting that I'm missing, or it's our old pal compiler weirdness.

Do we have a preference for using the `ddraw.h` macros for calling virtual methods? e.g. `IDirectDrawSurface_BltFast` versus `surf->BltFast()`.

### DIFF
--- a/LEGO1/mxtransitionmanager.cpp
+++ b/LEGO1/mxtransitionmanager.cpp
@@ -90,7 +90,7 @@ void MxTransitionManager::Transition_Dissolve()
   }
 
   if (res == DD_OK) {
-    FUN_1004c4d0(&ddsd);
+    FUN_1004c4d0(ddsd);
 
     for (int i = 0; i < 640; i++) {
       // Select 16 columns on each tick
@@ -115,7 +115,7 @@ void MxTransitionManager::Transition_Dissolve()
       }
     }
 
-    FUN_1004c580(&ddsd);
+    FUN_1004c580(ddsd);
     m_ddSurface->Unlock(ddsd.lpSurface);
 
     if (VideoManager()->GetVideoParam().flags().GetFlipSurfaces()) {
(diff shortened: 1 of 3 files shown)


LEGO1/mxtransitionmanager.cpp
```diff
@@ -38,12 +41,110 @@ MxResult MxTransitionManager::Tickle()
...
+  // If we are starting the animation
+  if (m_animationTimer == 0) {
+    // Generate the list of columns in order...
+    for (int i = 0; i < 640; i++) {
+      m_columnOrder[i] = i;
+    }
+
+    // ...then shuffle the list (to ensure that we hit each column once)
+    for (i = 0; i < 640; i++) {
+      int swap = rand() % 640;
```
foxtacles, line 69: We should probably use `MxS32` for all the loop bars in here

disinvite: For `swap`? Or loop counters `i` and `j`?

foxtacles: For all of the variables I'd say - or is there a good reason not to?

disinvite: No reason other than the habit of using int for the loop counter. I changed the types over.

foxtacles (approved):
Looks good, thanks!

> Do we have a preference for using the ddraw.h macros for calling virtual methods? e.g. IDirectDrawSurface_BltFast versus surf->BltFast().

I'd say the latter, this is what we have been using in all (known to me) places so far.

PR #159, opened 2023-10-04
Implement MxTransitionManager::EndTransition and GetCurrentWorld

itsmattkc:
EndTransition is *almost* 100%. It's off by a single instruction (`mov dword ptr [ebp - 4] , esi`), which alters some surrounding jmps:

Screenshot_20231003_213548

If anyone has any suggestions to improve this, I'd be happy to hear them, though it could also be random compiler entropy and is arguably close enough to be merged at this state (the missing instruction is literally pointless since it immediately gets replaced in the next instruction anyway)

### DIFF
--- a/LEGO1/legoomni.cpp
+++ b/LEGO1/legoomni.cpp
@@ -312,3 +312,9 @@ MxBool LegoOmni::vtable40()
   // FIXME: Stub
   return 0;
 }
+
+// OFFSET: LEGO1 0x100157a0
+LegoWorld *GetCurrentWorld()
+{
+  return LegoOmni::GetInstance()->GetCurrentWorld();
+}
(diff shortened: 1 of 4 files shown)


foxtacles (approved):
Matched the remaining diff. Passing the `MxParam` directly as a parameter was the first thing that occured to me since they've been doing that in other places. Not entirely sure why this generates slightly different code though.

This practice isn't allowed by the C++ Standard (afaik) but MSVC has always skirted the rules in this case a bit I think. Might not compile with g++.

itsmattkc:
Nice! Somehow that hadn't occurred to me, probably because, as you say, it wouldn't be permitted in standard C++. But I do remember this being a thing at one point earlier on. The `COMPAT_CONST` macro in `compat.h` was how I solved it last time - it's defined to `const` on GCC and nothing on MSVC so matching asm is preserved here while standards compliance is maintained there. However, it's a bit of a pain in this situation because it's on `Notify`, which is derived a lot throughout the code.

Perhaps not worth worrying about right now, but yeah it will be something we'll need to keep in mind in the future.

PR #160, opened 2023-10-04
Implement MxTransitionManager::SubmitCopyRect

cuckydev:
This PR implements and matches `MxTransitionManager::SubmitCopyRect`

I'm still working on SetupCopyRect, and that touches into various parts of the codebase, so I haven't included that here yet.

### DIFF
--- a/LEGO1/mxdsaction.h
+++ b/LEGO1/mxdsaction.h
@@ -57,6 +57,8 @@ public:
   inline char *GetExtraData() { return m_extraData; }
   inline MxU16 GetExtraLength() const { return m_extraLength; }
   inline MxLong GetStartTime() const { return m_startTime; }
+  inline MxS32 GetLoopCount() { return m_loopCount; }
+  inline void SetLoopCount(MxS32 m_loopCount) { this->m_loopCount = m_loopCount; }
   inline const MxVector3Data &GetLocation() const { return m_location; }
   inline void SetOmni(MxOmni *p_omni) { m_omni = p_omni; }
 
--- a/LEGO1/mxdsaction.h
+++ b/LEGO1/mxdsaction.h
@@ -59,7 +59,7 @@ public:
   inline MxU16 GetExtraLength() const { return m_extraLength; }
   inline MxLong GetStartTime() const { return m_startTime; }
   inline MxS32 GetLoopCount() { return m_loopCount; }
-  inline void SetLoopCount(MxS32 p_loopCount) { m_loopCount = m_loopCount; }
+  inline void SetLoopCount(MxS32 p_loopCount) { m_loopCount = p_loopCount; }
   inline const MxVector3Data &GetLocation() const { return m_location; }
   inline void SetOmni(MxOmni *p_omni) { m_omni = p_omni; }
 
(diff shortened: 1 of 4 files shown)


LEGO1/mxdsaction.h
```diff
@@ -57,6 +57,8 @@ class MxDSAction : public MxDSObject
   inline char *GetExtraData() { return m_extraData; }
   inline MxU16 GetExtraLength() const { return m_extraLength; }
   inline MxLong GetStartTime() const { return m_startTime; }
+  inline MxS32 GetLoopCount() { return m_loopCount; }
+  inline void SetLoopCount(MxS32 m_loopCount) { this->m_loopCount = m_loopCount; }
```
foxtacles, line 61: Minor issue, but the argument should be called `p_loopCount`

LEGO1/mxtransitionmanager.cpp
```diff
@@ -165,14 +147,14 @@ MxResult MxTransitionManager::StartTransition(TransitionType p_animationType, Mx
...
-      m_unk08->vtable54(1);
+    if (m_copyFlags.bit0 && m_waitIndicator != NULL) {
+      m_waitIndicator->Enable(TRUE);
 
-      MxTransitionManagerUnknownSubclass2 *iVar2 = m_unk08->m_unk1c;
-      iVar2->m_unk3c = 10000;
-      iVar2->m_unk30 |= 0x200;
+      MxDSAction *action = m_waitIndicator->GetAction();
+      action->SetLoopCount(10000);
+      action->SetFlags(action->GetFlags() | 0x200);
```
foxtacles, line 157: Can you add an "unknown" flag to the enum in `MxDSAction` and use that instead of `0x200`? Thank you!

LEGO1/mxdsaction.h
cuckydev: I was just following the precedent set here:
https://github.com/isledecomp/isle/blob/bb1782af46509c5cfe7d88e56411b95f7c8f42fc/LEGO1/mxdsaction.h#L56C1-L56C1

cuckydev:
I've noticed the function doesn't match anymore, which it did when I made the PR. Unfortunately this seems to depend on outside factors such as surrounding functions.

LEGO1/mxdsaction.h
```diff
@@ -52,16 +53,18 @@ class MxDSAction : public MxDSObject
...
 
-  inline MxU32 GetFlags() { return this->m_flags; } 
-  inline void SetFlags(MxU32 m_flags) { this->m_flags = m_flags; }
+  inline MxU32 GetFlags() { return m_flags; } 
+  inline void SetFlags(MxU32 p_flags) { m_flags = p_flags; }
   inline char *GetExtraData() { return m_extraData; }
   inline MxU16 GetExtraLength() const { return m_extraLength; }
   inline MxLong GetStartTime() const { return m_startTime; }
+  inline MxS32 GetLoopCount() { return m_loopCount; }
+  inline void SetLoopCount(MxS32 p_loopCount) { m_loopCount = m_loopCount; }
```
foxtacles, line 62: Another minor issue, but it should be `m_loopCount = p_loopCount`

foxtacles (approved):
Looks good, thank you!

PR #163, opened 2023-10-04
Implement MxTransitionManager::SetWaitIndicator

DimaLeon2000:


foxtacles:
There doesn't seem to be a match on `reccmp`; could you check? Thank you!

DimaLeon2000:
I've checked it, and it came out empty, as well. I think it's either a bug with the reccmp itself or it was my rookie mistake.

cuckydev:
image

The decompiled function here doesn't appear to be based off the original code.

PR #164, opened 2023-10-04
MxSoundManager: dtor, destroy, update types

ItzSwirlz:
Also make MxThread's destructor public.

ckdev also found the type of m_unk34, which is very helpful. I placed the import for DirectSound at the header of mxmediamanager so the other classes have it


foxtacles:
Looks good. `mxmediamanger.h` should not include `dsound.h` since it doesn't make use of it and shouldn't have any knowledge about requirements of potential subclasses. I moved it to `mxsoundmanger.h`.

PR #165, opened 2023-10-05
Implement MxTransitionManager SetWaitIndicator and SetupCopyRect

cuckydev:
This PR implements and matches `MxTransition::SetWaitIndicator`
This PR implements `MxTransition::SetupCopyRect`
`SetupCopyRect` does not currently match.

This PR touches into various other classes out of necessity.

### DIFF
--- a/LEGO1/mxdisplaysurface.cpp
+++ b/LEGO1/mxdisplaysurface.cpp
@@ -174,7 +174,7 @@ void MxDisplaySurface::SetPalette(MxPalette *p_palette)
 }
 
 // OFFSET: LEGO1 0x100bc200 STUB
-void MxDisplaySurface::vtable24(undefined4, undefined4, undefined4, undefined4, undefined4, undefined4, undefined4, undefined4)
+void MxDisplaySurface::vtable24(DDSURFACEDESC&, undefined4, undefined4, undefined4, undefined4, undefined4, undefined4, undefined4)
 {
 
 }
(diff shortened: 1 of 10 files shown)


LEGO1/mxdisplaysurface.cpp
```diff
@@ -174,7 +174,7 @@ void MxDisplaySurface::SetPalette(MxPalette *p_palette)
 }
 
 // OFFSET: LEGO1 0x100bc200 STUB
-void MxDisplaySurface::vtable24(undefined4, undefined4, undefined4, undefined4, undefined4, undefined4, undefined4, undefined4)
+void MxDisplaySurface::vtable24(DDSURFACEDESC&, undefined4, undefined4, undefined4, undefined4, undefined4, undefined4, undefined4)
```
foxtacles, line 177: This is something I missed previously, but could you change the interface of this and all the other functions (I believe it's `vtable2c` and the two Setup functions in the transition manager) to take the COM pointer version, `LPDDSURFACEDESC` instead of a reference to `DDSURFACEDESC`? This doesn't change anything functionally but would be more consistent with how they have used the DirectX APIs in other places. Thanks!

LEGO1/mxvideopresenter.cpp
```diff
@@ -2,8 +2,88 @@
...
+  // TODO
+}
+
+// OFFSET: LEGO1 0x1000c730 STUB
+void MxVideoPresenter::VTable0x70()
+{
+  // TODO
+}
+
+// OFFSET: LEGO1 0x1000c740
```
foxtacles, line 29: Can you mark this as `STUB`? Alternatively, maybe add a stub for the `Destroy` function at `100b27b0` and call it from the destructor, looks like it's the only thing it does anyway.

foxtacles (approved):
Looks good!

PR #166, opened 2023-10-05
(Proposal) Add injection hooking to test decompiled functions in-game

cuckydev:
This proposal adds a new compile option and a few macros that allows the decompilation to be built as to patch decompiled functions into the original LEGO1.DLL.

It's just a draft right now, there's a few things that still need to be worked out:
- SmartHeap is static linked into the patch DLL, which breaks memory allocation. There is probably a simple solution to this.
- Global variables do not have hooks yet. MSVC 4.2 makes this pretty hard so I'm not sure where to go with this.

I'm hoping some other people can provide insight on how to move this forward, as it'll help to be able to test non-matching functions in-game.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
 project(isle CXX)
 
 option(ISLE_BUILD_APP "Build ISLE.EXE application" ON)
+option(ISLE_BUILD_PATCH "Build patching into ISLE.EXE and LEGO1.DLL (requires ISLE_BUILD_APP)" OFF)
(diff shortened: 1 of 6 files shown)


foxtacles:
Interesting approach. Personally I think using a launcher (script) instead of baking the patching into the decomp directly might be easier/cleaner. Using the information from the `.pdb` file and annotations we already have should make this possible (like `reccmp.py` makes use of it; might even be possible to pull out some code from there and share it) - this would then also allow using any version of `ISLE.EXE` (let's say, the original) if so desired.

Just my two cents though, curious what everyone else thinks about it.

itsmattkc:
Yeah I agree that a separate launcher that automatically injects feels like a better approach. It wouldn't require any modifications to the codebase as we can just use the PDB mechanisms to link up original addresses to the recompiled addresses.

PR #168, opened 2023-10-05
MxTransitionManager::Tickle() & FUN_1004bcf0

Ramen2X:
Implements MxTransitionManager::Tickle() and MxTransitionManager::FUN_1004bcf0().

FUN_1004bcf0 is 100% matching. Tickle() was matching before merge.

This PR also additionally stubs all of the unimplemented transition methods.

### DIFF
--- a/LEGO1/mxdisplaysurface.cpp
+++ b/LEGO1/mxdisplaysurface.cpp
@@ -232,7 +232,7 @@ undefined4 MxDisplaySurface::vtable44(undefined4, undefined4*, undefined4, undef
 }
 
 // OFFSET: LEGO1 0x100ba640 STUB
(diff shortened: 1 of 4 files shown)


Ramen2X:
Quick additional note - I was unsure whether or not I should rename FUN_1004bcf0 adjacent to the other transition methods called in Tickle because it isn't really a transition in of itself, but I'd like to hear some more thoughts on this.

foxtacles:
> Quick additional note - I was unsure whether or not I should rename FUN_1004bcf0 adjacent to the other transition methods called in Tickle because it isn't really a transition in of itself, but I'd like to hear some more thoughts on this.

Yeah, I think we could call it something like `Transition_None`? Seems appropriate to me

foxtacles (approved):
Looks good, feel free to merge

PR #170, opened 2023-10-06
Refactor deserialization of objects

foxtacles:
This PR adds a few template helpers to clean up the source code of most of the `Deserialize` functions. It's not quite as elegant as I'd like it to be due to limited template support in this version of VC++ but it's still an improvement I think.

Matches stay the same.


itsmattkc:
Definitely cleaner. Looks good to me!

PR #176, opened 2023-10-07
Implement Start()

MishaProductions:
This PR implements the start function. 100% matching.

### DIFF
--- a/LEGO1/mxomni.cpp
+++ b/LEGO1/mxomni.cpp
@@ -42,10 +42,16 @@ void MxOmni::Init()
   m_unk64 = NULL;
 }
 
-// OFFSET: LEGO1 0x100b0090 STUB
-void MxOmni::vtable0x20()
+// OFFSET: LEGO1 0x100b0090
+long MxOmni::Start(MxDSAction* ds)
 {
-  // TODO
+  MxResult result = -1;
+  if(ds->GetAtomId().GetInternal() != NULL && ds->GetObjectId() != -1 && m_streamer != NULL)
+  {
+    result = m_streamer->Unknown100b99b0(ds);
+  }
+
+  return result;
 }
 
 // OFFSET: LEGO1 0x100b00c0 STUB
@@ -258,7 +264,7 @@ MxAtomIdCounterSet *AtomIdCounterSet()
 MxStreamer* Streamer()
 {
   return MxOmni::GetInstance()->GetStreamer();
-} 
+}
 
 // OFFSET: LEGO1 0x100acf00
 MxSoundManager* MSoundManager()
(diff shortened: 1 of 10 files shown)


LEGO1/mxomni.cpp
```diff
@@ -42,10 +42,16 @@ void MxOmni::Init()
   m_unk64 = NULL;
 }
 
-// OFFSET: LEGO1 0x100b0090 STUB
-void MxOmni::vtable0x20()
+// OFFSET: LEGO1 0x100b0090
+long MxOmni::Start(MxDSAction* ds)
```
foxtacles, line 46: The return type should probably be `MxResult`. Parameters should be prefixed with `p_`. We should also pass a reference instead of a pointer here, so I'd suggest `MxDSAction &p_dsAction` as a signature

LEGO1/mxomni.cpp
```diff
@@ -42,10 +42,16 @@ void MxOmni::Init()
...
 {
-  // TODO
+  MxResult result = -1;
```
foxtacles, line 48: Please use the appropriate constants for `MxResult` values (`SUCCESS` and `FAILURE`) (several places in this PR)

LEGO1/mxstreamcontroller.cpp
```diff
@@ -30,3 +30,22 @@ MxResult MxStreamController::Open(const char *p_filename)
...
+}
+
+// OFFSET: LEGO1 0x100b9410
+undefined4 MxStreamController::vtable0x1C(undefined4 p_unknown, undefined4 p_unknown2)
+{
+  return -1;
+}
+
+// OFFSET: LEGO1 0x100c1690 STUB
+long MxStreamController::vtable0x20(MxDSAction* action)
```
foxtacles, line 47: Signature adjustments necessary as described in the other comment

LEGO1/mxomni.cpp
MishaProductions: If I use a reference instead of a pointer, it wouldn't match anymore because MxOmni::Start and MxStreammer::Unknown100b99b0 checks if it's null.

LEGO1/mxstreamcontroller.cpp
```diff
@@ -30,3 +30,22 @@ MxResult MxStreamController::Open(const char *p_filename)
...
+}
+
+// OFFSET: LEGO1 0x100b9410
+MxResult MxStreamController::vtable0x1C(undefined4 p_unknown, undefined4 p_unknown2)
+{
+  return FAILURE;
+}
+
+// OFFSET: LEGO1 0x100c1690 STUB
+long MxStreamController::vtable0x20(MxDSAction* action)
```
foxtacles, line 47: This still needs to be `MxResult`

foxtacles (approved):
Looks good, thank you!

PR #177, opened 2023-10-07
Supporting items for SetupCopyRect()

disinvite:
I was looking at SetupCopyRect() in MxTransitionManager (#165) to see if I could improve the match percentage. No such luck but I have some minor improvements to offer for the related code.

GetWidth() and GetHeight() from MxVideoPresenter point at a reference to MxBitmap. Confirmed with debugger. The 0x60 vtable method from subclass MxSmkPresenter sets this up, so I implemented that to give an example. The 0x60 method from MxVideoPresenter is a no-op. The width and height methods should be correct, but the compiler keeps picking the wrong order for the if/else. I went with the most concise syntax in hope that it will resolve itself later.

The one direct improvement to SetupCopyRect has to do with the m_location member of m_waitIndicator. Grabbing both the x and y coordinates of the MxPoint32 struct pulls in a pointer to the object. I added getter methods to grab the values directly and this eliminates the reference to better match the code. (I thought maybe this meant that MxPresenter does not use MxPoint32 at all, but if you break it out into two MxS32 members, StartAction() no longer matches.)

Plus: A few other very minor improvements while I had the files open.

### DIFF
--- a/LEGO1/mxpresenter.h
+++ b/LEGO1/mxpresenter.h
@@ -70,8 +70,8 @@ public:
 
   inline MxS32 GetCurrentTickleState() const { return this->m_currentTickleState; }
   inline MxPoint32 GetLocation() const { return this->m_location; }
-  inline MxS32 GetDisplayX() const { return this->m_location.m_x; }
-  inline MxS32 GetDisplayY() const { return this->m_location.m_y; }
+  inline MxS32 GetLocationX() const { return this->m_location.m_x; }
+  inline MxS32 GetLocationY() const { return this->m_location.m_y; }
   inline MxS32 GetDisplayZ() const { return this->m_displayZ; }
   inline MxDSAction *GetAction() const { return this->m_action; }
 
--- a/LEGO1/mxpresenter.h
+++ b/LEGO1/mxpresenter.h
@@ -68,17 +68,19 @@ public:
 
   MxBool IsEnabled();
 
-  inline MxS32 GetCurrentTickleState() { return this->m_currentTickleState; }
-  inline MxPoint32 GetLocation() { return this->m_location; }
-  inline MxS32 GetDisplayZ() { return this->m_displayZ; }
-  inline MxDSAction *GetAction() { return this->m_action; }
+  inline MxS32 GetCurrentTickleState() const { return this->m_currentTickleState; }
+  inline MxPoint32 GetLocation() const { return this->m_location; }
+  inline MxS32 GetDisplayX() const { return this->m_location.m_x; }
+  inline MxS32 GetDisplayY() const { return this->m_location.m_y; }
+  inline MxS32 GetDisplayZ() const { return this->m_displayZ; }
+  inline MxDSAction *GetAction() const { return this->m_action; }
 
 protected:
   __declspec(dllexport) void Init();
   void SendTo_unkPresenter(MxOmni *);
 
 private:
-  MxS32 m_currentTickleState; // 0x8
+  TickleState m_currentTickleState; // 0x8
   MxU32 m_previousTickleStates;
   MxPoint32 m_location;
   MxS32 m_displayZ;
--- a/LEGO1/mxtransitionmanager.cpp
+++ b/LEGO1/mxtransitionmanager.cpp
@@ -23,7 +23,7 @@ MxTransitionManager::MxTransitionManager()
 // OFFSET: LEGO1 0x1004ba00
 MxTransitionManager::~MxTransitionManager()
 {
-  free(m_copyBuffer);
+  delete[] m_copyBuffer;
 
   if (m_waitIndicator != NULL) {
     delete m_waitIndicator->GetAction();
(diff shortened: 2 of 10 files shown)


LEGO1/mxpresenter.h
```diff
@@ -68,17 +68,19 @@ class MxPresenter : public MxCore
 
   MxBool IsEnabled();
 
-  inline MxS32 GetCurrentTickleState() { return this->m_currentTickleState; }
-  inline MxPoint32 GetLocation() { return this->m_location; }
-  inline MxS32 GetDisplayZ() { return this->m_displayZ; }
-  inline MxDSAction *GetAction() { return this->m_action; }
+  inline MxS32 GetCurrentTickleState() const { return this->m_currentTickleState; }
+  inline MxPoint32 GetLocation() const { return this->m_location; }
+  inline MxS32 GetDisplayX() const { return this->m_location.m_x; }
```
foxtacles, line 73: I'd call this `GetLocationX` (same with Y) since it appears to be somewhat distinct from the Z value (also confirmed by the 1996 source).

Did you try adding getters like `GetX` (and Y) to `MxPoint32` such that the call would be `GetLocation().GetX()` or does this also lower accuracy? If not, I think that might be the cleanest. I've seen those getters in the 1996 source on that class.

disinvite: I just tried with GetX() and GetY() and it's the same as using `m_location.m_x` directly. With that you get an `LEA` on the address of the MxPoint32 member. To illustrate what I mean:

```assembly
; ECX points to m_waitIndicator:
  00044 8d 71 10    lea esi, DWORD PTR [ecx+16]

; [...]
; 361  :     m_copyRect.left = m_waitIndicator->GetLocation().GetX();

  0004b 8b 06       mov eax, DWORD PTR [esi]
  0004d 89 43 0c    mov DWORD PTR [ebx+12], eax

; 362  :     m_copyRect.top = m_waitIndicator->GetLocation().GetY();

  00050 8b 46 04    mov eax, DWORD PTR [esi+4]
  00053 89 43 10    mov DWORD PTR [ebx+16], eax
```

Versus the original code that just grabs the values as if there were no struct at all:

```assembly
; ECX points to m_waitIndicator:
1004c5ba 8b 79 10   MOV     EDI,dword ptr [ECX + 0x10] ; m_location.X
1004c5bd 8b 46 54   MOV     EAX,dword ptr [ESI + 0x54]
1004c5c0 45         INC     EBP
1004c5c1 c1 e8 03   SHR     EAX,0x3
1004c5c4 8b 51 14   MOV     EDX,dword ptr [ECX + 0x14] ; m_location.Y
```

So I'll change it to GetLocationX() and Y. Is the Z value in this context meant to be the z-depth (or z-index in HTML land)? As in: we have screen x and y and an order to draw things, not a full 3d coodinate? (Maybe that's just splitting hairs, don't mind me)

Last thing: I changed the malloc/free in TransitionManager to new/delete to match what we're doing everywhere else.

foxtacles: > Is the Z value in this context meant to be the z-depth (or z-index in HTML land)? As in: we have screen x and y and an order to draw things, not a full 3d coodinate? (Maybe that's just splitting hairs, don't mind me)

Something like that I'd say, yes. I don't know for sure yet but considering they chose to separate the Z-value into its own member separate from the X/Y point with a distinct name (`displayZ` - that's how they call it in the 1996 source). The presenters are also sorted according to their Z-value here: https://github.com/isledecomp/isle/blob/master/LEGO1/mxvideomanager.cpp#L76

LEGO1/mxvideopresenter.cpp
```diff
@@ -44,24 +44,24 @@ void MxVideoPresenter::VTable0x78()
...
+// OFFSET: LEGO1 0x1000c7e0
 MxS32 MxVideoPresenter::GetWidth()
 {
-  // TODO
-  return 0;
+  return m_unk54 ? m_unk54->width
+                 : m_bitmap->GetBmiHeader()->biWidth;
 }
 
 // OFFSET: LEGO1 0x1000c800 STUB
```
foxtacles, line 60: Probably shouldn't be `STUB`?

LEGO1/mxtransitionmanager.cpp
```diff
@@ -370,7 +370,7 @@ void MxTransitionManager::SetupCopyRect(LPDDSURFACEDESC ddsc)
     // Allocate the copy buffer
     const char *src = (const char*)ddsc->lpSurface + m_copyRect.top * ddsc->lPitch + bytesPerPixel * m_copyRect.left;
 
-    m_copyBuffer = malloc(bytesPerPixel * width * height);
+    m_copyBuffer = new char[bytesPerPixel * width * height];
```
foxtacles, line 373: We should probably change the type of `m_copyBuffer` from `void*` to `char*`

disinvite: MxU8* is even better

foxtacles (approved):
Looks good, thank you!

PR #178, opened 2023-10-07
Move/Merge MxMIDIManager into MxMusicManager

DimaLeon2000:


foxtacles:
Seems correct to me since the Omni getter `MusicManager` does indeed return an instance of a class with this vtable. Will merge.

PR #179, opened 2023-10-07
Implement a few mxomni functions

MishaProductions:
- Implemented MxOmni::Create (doesn't match probably due to inline constructors)
 - Implemented MxOmni::StartTimer and StopTimer
 - Improved accuracy of MxOmni::Notify
 As a result of implementing MxOmni::Create, the game no longer crashes with our lego1 when you run it.

### DIFF
--- a/LEGO1/mxeventmanager.cpp
+++ b/LEGO1/mxeventmanager.cpp
@@ -1,5 +1,4 @@
 #include "mxeventmanager.h"
-#include "decomp.h"
 
 // OFFSET: LEGO1 0x100c0360
 MxEventManager::MxEventManager()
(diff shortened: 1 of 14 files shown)


LEGO1/mxeventmanager.cpp
```diff
@@ -15,4 +16,11 @@ MxEventManager::~MxEventManager()
 // OFFSET: LEGO1 0x100c0450
 void MxEventManager::Init()
 {
-}
\ No newline at end of file
+}
+
+// OFFSET: LEGO1 0x100c04a0 STUB
+MxResult MxEventManager::vtable0x28(undefined4 p_unknown1, MxU8 p_unknown2)
```
foxtacles, line 22: `MxU8` should be `undefined`

LEGO1/mxomni.cpp
```diff
@@ -51,7 +51,7 @@ void MxOmni::Init()
   m_timer = NULL;
   m_streamer = NULL;
   m_atomIdCounterSet = NULL;
-  m_unk64 = NULL;
+  m_TimerRunning = FALSE;
```
foxtacles, line 54: Should be called `m_timerRunning` based on our style guide

foxtacles:
Could you split this into smaller PRs that each focus on one thing at a time? This way it would be much easier to review. Thank you!

MishaProductions:
Yeah I probably should have done that.

PR #180, opened 2023-10-07
Silence all C4786 warnings

foxtacles:
I've found a rather obscure workaround that eliminates all of the `C4786` warning spam. The fact that the `#pragma` doesn't work has been a bug that survived all the way into VC6:

https://web.archive.org/web/20140128153313/http://support.microsoft.com/kb/167355

The workaround has been figured out by someone about 20 years ago, credit is in the comments.


itsmattkc:
Classic MSVC moment. Looks good to me!

(sidenote, it's wild how even something as inconsequential as this messes with function accuracy, but I guess that's just how MSVC is)

PR #181, opened 2023-10-07
Implement MxOmni Start/stop timer

MishaProductions:


LEGO1/mxsoundmanager.cpp
```diff
@@ -46,4 +46,23 @@ void MxSoundManager::Destroy(MxBool p_param)
   if (!p_param) {
     MxAudioManager::Destroy();
   }
-}
\ No newline at end of file
+}
+
+// OFFSET: LEGO1 0x100ae8b0
```
foxtacles, line 51: Should be marked as `STUB`?

foxtacles:
Please remove the two junk files, I think then it's good to go!

foxtacles (approved):

PR #182, opened 2023-10-07
Implement MxOmni::Create

MishaProductions:


LEGO1/mxomni.cpp
```diff
@@ -248,8 +344,24 @@ void MxOmni::Destroy()
 // OFFSET: LEGO1 0x100b07f0
 MxLong MxOmni::Notify(MxParam &p)
 {
-  // FIXME: Stub
-  return 0;
+  MxAutoLocker lock(&this->m_criticalsection);
+  MxLong result;
+  if (p.GetType() == 2)
```
foxtacles, line 349: I think it would be a good moment to start introducing an `enum` in `MxParam` to have named constants for the known (and unknown) notification types, along the lines of: https://github.com/isledecomp/isle/blob/master/LEGO1/mxpresenter.h#L18

and also use it as the type for `m_type` in `MxParam`.

LEGO1/mxomni.h
```diff
@@ -61,6 +61,7 @@ class MxOmni : public MxCore
   MxMusicManager* GetMusicManager() const { return this->m_musicManager; }
   MxEventManager* GetEventManager() const { return this->m_eventManager; }
   MxAtomIdCounterSet* GetAtomIdCounterSet() const { return this->m_atomIdCounterSet; }
+  MxLong HandleNotificationType2(MxParam& p_param);
```
foxtacles, line 64: Return type should be `MxResult`

foxtacles:
Although the accuracy of `MxOmni::Create` went down, it looks like progress in terms of functional implementation. A proper match might take some effort since it's a very large function. Merging.

PR #183, opened 2023-10-07
Implement remaining MxDSSelectAction functions

foxtacles:
Matches:

* `GetSizeOnDisk`: 100%
* `Deserialize`: ~97% (remaining % is compiler randomness)

`Deserialize` is a fairly involved function, but gives a hint on what the "select" part of the class name is referring to - it only picks out a single action from the data stream based on an identifier and adds it to the list.

I believe this marks the completion in terms of both functional implementation and accuracy of all the subclasses of `MxDSObject` (unless I missed something), except for this one `TODO`: https://github.com/isledecomp/isle/blob/master/LEGO1/mxdsaction.cpp#L165


itsmattkc:
Damn that's a big function indeed. Congrats on getting it as close as you could (minus the MSVC weirdness). This looks good to me, feel free to merge.

PR #184, opened 2023-10-07
MxMusicPresenter vtable

ItzSwirlz:
vtable+0x34 adds the current presenter to the music manager vtable+0x38 calls destroy with the param as "FALSE"

---

I also added DECOMP_SIZE_ASSERT's, although for this class its probably unnecessary

### DIFF
--- a/LEGO1/mxmusicpresenter.cpp
+++ b/LEGO1/mxmusicpresenter.cpp
@@ -1,5 +1,11 @@
 #include "mxmusicpresenter.h"
 
+#include "decomp.h"
+#include "mxmusicmanager.h"
+#include "mxomni.h"
+
+DECOMP_SIZE_ASSERT(MxMusicPresenter, 0x54);
+
 // OFFSET: LEGO1 0x100c22c0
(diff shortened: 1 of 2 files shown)


LEGO1/mxmusicpresenter.cpp
```diff
@@ -21,4 +27,22 @@ void MxMusicPresenter::Init()
...
+    MusicManager()->AddPresenter(*this);
+  }
+  return ret;
+} 
+
+// OFFSET: LEGO1 0x100c25d0
+void MxMusicPresenter::vtable38()
+{
+  // TODO: Name this function when we know what the argument to Destroy does
+  Destroy(FALSE);
```
MishaProductions, line 47: Looking at 0x100b55fe, if that argument is false, then MxPresenter::Init is called

ItzSwirlz: Cool! I'd implement that now but I'd rather make another PR for Destroy for MxMusicPresenter since it requires making a stub and also making a member public in MxPresenter

PR #185, opened 2023-10-07
Finish MxMediaManager and MxEventManager

MishaProductions:
All functions are 100% matching, except MxEventManager::CreateEventThread (which is almost identical to MxMusicManager::StartMIDIThread).

### DIFF
--- a/LEGO1/mxeventmanager.cpp
+++ b/LEGO1/mxeventmanager.cpp
@@ -23,7 +23,7 @@ void MxEventManager::Init()
 }
 
 // OFFSET: LEGO1 0x100c04a0
-MxResult MxEventManager::CreateEventThread(int p_frequencyMS, MxBool p_noRegister)
+MxResult MxEventManager::CreateEventThread(MxU32 p_frequencyMS, MxBool p_noRegister)
 {
   MxResult status = FAILURE;
   MxBool locked = FALSE;
--- a/LEGO1/mxeventmanager.cpp
+++ b/LEGO1/mxeventmanager.cpp
@@ -1,4 +1,8 @@
 #include "mxeventmanager.h"
+#include "mxcriticalsection.h"
+#include "mxthread.h"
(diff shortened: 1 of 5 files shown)


LEGO1/mxeventmanager.h
```diff
@@ -11,7 +11,7 @@ class MxEventManager : public MxMediaManager
 public:
   MxEventManager();
   virtual ~MxEventManager() override;
-  virtual MxResult vtable0x28(undefined4 p_unknown1, MxU8 p_unknown2); // vtable+28
+  virtual MxResult MxEventManager::CreateEventThread(int p_frequencyMS, MxBool p_noRegister); // vtable+28
```
foxtacles, line 14: The class name shouldn't be prefixed

LEGO1/mxmediamanager.h
```diff
@@ -23,6 +23,7 @@ class MxMediaManager : public MxCore
   virtual void StopPresenters(); // vtable+24
 
   MxResult Init();
+  void TerminateThread(MxBool p_dontDestroy);
```
foxtacles, line 26: The argument name is not consistent with the one used in the implementation

LEGO1/mxeventmanager.cpp
```diff
@@ -9,17 +13,54 @@ MxEventManager::MxEventManager()
...
 // OFFSET: LEGO1 0x100c0450
 void MxEventManager::Init()
 {
+  // This is intentionally left blank
 }
 
-// OFFSET: LEGO1 0x100c04a0 STUB
-MxResult MxEventManager::vtable0x28(undefined4 p_unknown1, undefined p_unknown2)
+// OFFSET: LEGO1 0x100c04a0
+MxResult MxEventManager::CreateEventThread(int p_frequencyMS, MxBool p_noRegister)
```
foxtacles, line 26: Should use `MxU32` instead of `int`

PR #186, opened 2023-10-07
Start MxCompositePresenter ctor/dtor

ItzSwirlz:
really weird register swaps


foxtacles:
`MxCompositePresenter` has the same list-like class that is being used in `MxDiskStreamProvider` (and several other classes). I refactored the work-in-progress class chunks into `mxunklist.h`.

PR #187, opened 2023-10-07
MxVideoPresenter functions

MishaProductions:
TODO: match MxVideoPresenter::Destroy


LEGO1/mxrect32.h
```diff
@@ -1,6 +1,9 @@
 #ifndef MXRECT32_H
 #define MXRECT32_H
 
+#define MxRect32GetWidth(rect) (rect.m_right - rect.m_left) + 1
+#define MxRect32GetHeight(rect) (rect.m_bottom - rect.m_top) + 1
+
```
disinvite, line 6: Does it affect the match percentage if these are inline getter methods instead?

MishaProductions: It shouldn't, but that's how microsoft does the rectangle width/height macros, so I assumed that's how mindscape did

MishaProductions: Nope, it doesn't affect the percentage, but it's cleaner.

MishaProductions:
I don't really know how to make it match

foxtacles:
Got the match up to ~88%. The rest *might* be compiler randomness but I think there's also a chance a minor piece needs to be changed, however I can't think of any.

Changes that were made:

* `m_unk58` is `LPDIRECTDRAWSURFACE`
* `UnkStruct` (previously `unknown_meta_struct`) seems to have a virtual table according to this function, so I added a virtual destructor for now. This structure still needs more investigation.
* Reordered some instructions

PR #188, opened 2023-10-08
Remove goto from isleapp.cpp

ChrisMiuchiz:
WinMain stays at 99.52% but this is much more likely to be how the code was actually written.

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -250,11 +250,8 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
         break;
       }
 
-      if (g_mousedown == 0) {
-        // Nothing
-      }
-      else if (g_mousemoved && g_isle) {
-          g_isle->Tick(0);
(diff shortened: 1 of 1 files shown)


DimaLeon2000 (changes requested):
Shouldn't it be `if (g_mousedown && g_mousemoved && g_isle)`?

foxtacles:
Nice! We've tried eliminating that `goto` before but I guess we missed this obvious solution. The suggestion of @DimaLeon2000 seems to match as well and would be the most concise.

ISLE/isleapp.cpp
```diff
@@ -251,15 +251,14 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
...
-LAB_00401bc7:
-        if (g_mousemoved) {
-          g_mousemoved = FALSE;
-        }
-      } else if (g_mousemoved) {
-        if (g_isle) {
+        // Nothing
+      }
+      else if (g_mousemoved && g_isle) {
           g_isle->Tick(0);
```
foxtacles, line 257: Minor issue, but please use 2-space indents. Thanks!

PR #189, opened 2023-10-08
MxMediaPresenter vtable

MishaProductions:


LEGO1/mxpresenter.h
```diff
@@ -79,7 +79,6 @@ class MxPresenter : public MxCore
   __declspec(dllexport) void Init();
   void SendTo_unkPresenter(MxOmni *);
 
-private:
```
foxtacles, line 82: By removing the `private` keyword you make the members `public`, which I don't think is a good idea in this case. `protected` seems more appropriate.

MishaProductions: `protected` is already used above

foxtacles: Ah, right!

PR #190, opened 2023-10-08
Implement LegoVideoManager::MoveCursor and improve SetSkyColor

MishaProductions:
MoveCursor doesn't match because the registers are swapped for some reason.

### DIFF
--- a/LEGO1/legovideomanager.cpp
+++ b/LEGO1/legovideomanager.cpp
@@ -1,5 +1,7 @@
 #include "legovideomanager.h"
 
+DECOMP_SIZE_ASSERT(LegoVideoManager, 0x590);
+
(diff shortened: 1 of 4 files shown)


foxtacles:
Looks good, I've just made some changes:

* I removed the members that aren't actually being used anywhere yet. The reason for that is that their declaration implies that their type and location is generally assumed to be correct, but unless this has been validated by actual code there is no way to really know, and it causes needless confusion down the road when they turn out wrong. My recommendation is to wait with their declaration unless you have at least 1 piece of code that makes use of them and thus validates their correctness (to some degree). Until then, keep that information as comments somwhere (in Ghidra, for instance)
* Used `MxS32` type instead of `int`
* Added size assert for `LegoVideoManager`
* `#include <ddraw.h>` instead of `#include "ddraw.h"`

PR #191, opened 2023-10-08
MxStream provider and controller vtables

MishaProductions:
Everything matches except MxRAMStreamProvider::MxRAMStreamProvider because the vtable gets written after `m_bufferSize = 0;` for some reason.


foxtacles:
Looks good, my changes:

* Matched `MxRAMStreamProvider::MxRAMStreamProvider` (see changes to parent class initialization)
* `GetAtom` broke the match for `Isle::Tick`; changed it to return a reference
* Replaced unused member with `undefined4`
* Minor naming changes

PR #192, opened 2023-10-08
Implemented MxVector4::UnknownQuaternionOp

AngelTomkins:
Implemented MxVector4::UnknownQuaternionOp,
Currently builds to binary with 65% similarity to original
Removed the function from being a stub, although I have not tested the functionality of this function, visually it seems to me to have exactly the same functionality.


AngelTomkins:
Apparently I don't understand how git works and if I update my own branch it will update this pull request, I was planning on making a pull request for bringing MxVector4::DotImpl to match 100%, but I guess it is here now. Let me know if I should do anything to fix this right now or just let it be as it is?

foxtacles:
Looks good! Yes, generally try to create distinct branches for your developments. A pull request is always based on the current state of a branch in your fork.

foxtacles (approved):

PR #194, opened 2023-10-09
Fix MxDSMediaAction structure/matches

foxtacles:
This completes the remaining, open inaccuracies in the class. Two members are part of a structure (which remains unknown for now; could be `MxPoint32` or similar).


itsmattkc:
Looks good to me, merging

PR #195, opened 2023-10-09
MxBitmap code clarity

disinvite:
Some minor touch-ups to MxBitmap. Should be no change to match percentage.

- The BMP header magic string "BM" was backwards. It is `42 4d` in the file, but is compared as a little-endian unsigned short, so the value in C should be `0x4d42`.

- We discussed this on the Matrix chat a while back, but after some more thought, I think the correct thing is for the m_bitDepth member (renamed here to m_isHighColor) to just be an `MxBool`. It's strange that they use a switch statement to check the values, but not invalid. We know the devs used enums in other places; if this was intended to have more values, it would have been an enum and would be 4 bytes wide. What other values could there be? Grayscale mode?

- Added a helper function to replace the `(value + 3) & -4` expressions. This is a nice little trick to round the bitmap width up to the next multiple of four so when we alloc it will be `DWORD` aligned. This padding is part of the BMP standard and called "image stride."

- Added another helper function to replace (what amounted to) `height = height > 0 ? height : -height` expressions. Image height can be negative to represent top-down images; again part of the BMP standard.

  This is of course just the `Abs()` function, but if we use the one from `legoutil.h`, the terms are reversed and it doesn't match. Furthermore, if you try to put this new `_Abs()` function directly into the `width * height` calculation, it also doesn't match. There has to be a "height" variable, which is sort of what we already had. Maybe something to come back to at the very end.

  You could argue that both the dword-align and absolute value steps should go together as we will always use both to determine how many bytes are needed to store the pixel data. However, they are not always both used. See: ImportBitmapInfo.

- Added some newlines and spaces, and moved variable declarations around to help overall clarity.

### DIFF
--- a/LEGO1/mxbitmap.cpp
+++ b/LEGO1/mxbitmap.cpp
@@ -7,7 +7,7 @@ DECOMP_SIZE_ASSERT(MxBITMAPINFO, 0x428);
 // Bitmap header magic string "BM" (42 4d)
 // Sources: https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapfileheader, DirectX Complete (1998)
 // GLOBAL: LEGO1 0x10102184
-MxU16 g_bitmapSignature = 0x4d42;
+MxU16 g_bitmapSignature = TWOCC('B', 'M');
 
 // OFFSET: LEGO1 0x100bc980
 MxBitmap::MxBitmap()
@@ -132,7 +132,7 @@ MxResult MxBitmap::ImportColorsToPalette(RGBQUAD* p_rgbquad, MxPalette* p_palett
       return ret;
   }
 
-  for (int i = 0; i < 256; i++) {
+  for (MxS32 i = 0; i < 256; i++) {
     p_rgbquad[i].rgbRed      = entries[i].peRed;
     p_rgbquad[i].rgbGreen    = entries[i].peGreen;
     p_rgbquad[i].rgbBlue     = entries[i].peBlue;
@@ -144,7 +144,7 @@ MxResult MxBitmap::ImportColorsToPalette(RGBQUAD* p_rgbquad, MxPalette* p_palett
 }
 
 // OFFSET: LEGO1 0x100bcaa0
-MxResult MxBitmap::SetSize(int p_width, int p_height, MxPalette *p_palette, int p_isHighColor)
+MxResult MxBitmap::SetSize(MxS32 p_width, MxS32 p_height, MxPalette *p_palette, MxBool p_isHighColor)
 {
   MxResult ret = FAILURE;
   MxLong size = AlignToFourByte(p_width) * p_height;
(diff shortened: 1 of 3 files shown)


LEGO1/mxbitmap.cpp
```diff
@@ -1,12 +1,13 @@
...
-DECOMP_SIZE_ASSERT(MxBITMAPINFO, 1064);
+DECOMP_SIZE_ASSERT(MxBitmap, 0x20);
+DECOMP_SIZE_ASSERT(MxBITMAPINFO, 0x428);
 
-// The way that the BITMAPFILEHEADER structure ensures the file type is by ensuring it is "BM", which is literally just 0x424d.
+// Bitmap header magic string "BM" (42 4d)
 // Sources: https://learn.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapfileheader, DirectX Complete (1998)
 // GLOBAL: LEGO1 0x10102184
-undefined2 g_bitmapSignature = 0x424d;
+MxU16 g_bitmapSignature = 0x4d42;
```
foxtacles, line 10: I think we have a `TWOCC` macro we can use for this

LEGO1/mxbitmap.cpp
```diff
@@ -130,14 +144,14 @@ MxResult MxBitmap::ImportColorsToPalette(RGBQUAD* p_rgbquad, MxPalette* p_palett
 }
 
 // OFFSET: LEGO1 0x100bcaa0
-MxResult MxBitmap::SetSize(int p_width, int p_height, MxPalette *p_palette, int p_bitDepth)
+MxResult MxBitmap::SetSize(int p_width, int p_height, MxPalette *p_palette, int p_isHighColor)
```
foxtacles, line 147: The params should be one of our Mx types ideally?

LEGO1/mxbitmap.cpp
```diff
@@ -30,43 +31,54 @@ MxBitmap::~MxBitmap()
...
+// https://learn.microsoft.com/en-us/windows/win32/medfound/image-stride
+inline MxLong AlignToFourByte(MxLong p_value)
+{
+  return (p_value + 3) & -4;
+}
+
+// Same as the one from legoutil.h, but flipped the other way
+// TODO: While it's not outside the realm of possibility that they
+// reimplemented Abs for only this file, that seems odd, right?
+inline MxLong _Abs(MxLong p_value)
```
foxtacles, line 45: There's plenty of redundancy I believe, wouldn't surprise me they had multiple implementations of stuff like that. Could be they just had this logic inline and not in a separate function

LEGO1/mxbitmap.cpp
disinvite: Ah good call

foxtacles:
Looks good! Accuracy went down on paper but looks like compiler randomness to me, merging.

foxtacles (approved):

disinvite:
Thanks!

PR #196, opened 2023-10-09
implement EndAction

vMidz:
implement EndAction
+offsets commenets

@foxtacles: also contains changes from this PR: https://github.com/vMidz/isle/pull/1 - see description over there


foxtacles:
Added an attempt to refactor `MxParam` in this PR: https://github.com/vMidz/isle/pull/1 which I think might be useful to incorporate into this PR before merging.

itsmattkc:
> Added an attempt to refactor `MxParam` in this PR: [vMidz#1](https://github.com/vMidz/isle/pull/1) which I think might be useful to incorporate into this PR before merging, but would like feedback before doing so.

That feels right to me. I remember thinking when I was doing #130 that it would make a lot of sense for `MxOmniCreateParam` to inherit from `MxParam`, but second guessed myself because `MxParam` had already been implemented as a derivative of `MxOmniCreateParamBase`. I hadn't considered that it could actually be an `MxParam` specifically for `Notify` signals, but it makes total sense the way we see it used, especially since it's a member of `MxNotification`. I think this is correct and I'd say it's good to merge.

PR #198, opened 2023-10-11
Implemented  MxVector4::SetMatrixProductWrapper

AngelTomkins:
Implemented SetMatrixProductWrapper with 100% match. Notably this function is a wrapper for SetMatrixProduct with different parameters. The name is not the same with the function overloading paradigm since it makes the function no longer have 100% binary match.


foxtacles:
Looks good, renamed the function to `SetMatrixProductImpl` (to be consistent with the other functions in the class with the same pattern)

PR #202, opened 2023-10-12
Rename MxMusicPresenter function vtable38

DimaLeon2000:


foxtacles:
Maybe I can shine some light on the `Destroy` pattern we have found in the codebase: based on the 1996 source, the argument to the non-virtual function `Destroy` is called `p_fromDestructor`, meaning that the switch is meant to be used from two places:

* From the destructor, where it is called with `TRUE` as an argument
* From the *virtual* function called `Destroy` (which takes no argument) with the argument `FALSE`

My suggestion would be to just use the same naming pattern for the sake of authenticity and consistency.

PR #203, opened 2023-10-13
rename all uncertain labels to unk

Ramen2X:
This PR renames all uncertain member/variable/method names to `unk` equivalents. The reason for this is that labeling something that is unknown does not change the fact that, ultimately, it is still unknown and just serves to further complicate the code style of the project. These names still end up providing little useful information to the developer. This naming style also creates further issues as the existence of these labels in the code makes it impossible to get an accurate number of unlabeled members/variables/methods if we were to, for instance, `grep` or otherwise search for all instances of `unk` or `unknown` in the repository.

### DIFF
--- a/LEGO1/infocenterstate.h
+++ b/LEGO1/infocenterstate.h
@@ -31,7 +31,7 @@ public:
 private:
   // Members should be renamed with their offsets before use
   /*
-    struct SomeStruct
+    struct UnkStruct
     {
       undefined4 unk1;
       undefined2 unk2;
@@ -53,7 +53,7 @@ private:
     undefined2 unk10;
     undefined2 unk11;
     undefined2 padding3;
-    SomeStruct unk12[6];
+    UnkStruct unk12[6];
     undefined4 unk13;
   */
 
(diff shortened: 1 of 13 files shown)


foxtacles (approved):

foxtacles:
Tangentially related, I think we should also introduce a naming rule that we format the offset suffix as `0x{offset}` instead of appending just the offset, I think it would be more readable.

Ramen2X:
> Tangentially related, I think we should also introduce a naming rule that we format the offset suffix as `0x{offset}` instead of appending just the offset, I think it would be more readable.

Agreed, I have been noticing that everyone names their unk members a bit differently. It would clean up the codebase nicely if it were to be uniform.

foxtacles:
> > Tangentially related, I think we should also introduce a naming rule that we format the offset suffix as `0x{offset}` instead of appending just the offset, I think it would be more readable.
> 
> Agreed, I have also been noticing that everyone names their unk members a bit differently. It would clean up the codebase nicely if it were to be uniform.

This along with the other naming/styling rules can probably be enforced by a tool/linter to a large degree. I'm not very familiar with the tooling out there in this regard but considering the recent influx of new contributors (which is great) additional tooling like that could save quite some time, so I'd say we should consider looking into introducing something

PR #204, opened 2023-10-14
MxMIDIPresenter functions

ItzSwirlz:


LEGO1/mxmidipresenter.cpp
```diff
@@ -9,8 +9,20 @@ MxMIDIPresenter::MxMIDIPresenter() {
...
+
 // OFFSET: LEGO1 0x100c2820
 void MxMIDIPresenter::Init()
 {
   m_unk54 = 0;
 }
+
+// OFFSET: LEGO1 0x100c2830 STUB
+void MxMIDIPresenter::Destroy(MxBool p_param)
+{
```
MishaProductions, line 26: p_param should be called p_fromDestructor according to https://github.com/isledecomp/isle/pull/202#issuecomment-1761171826

PR #208, opened 2023-10-14
Implement MxVideoManager::vtable0x28 and Create

foxtacles:
Depends on https://github.com/isledecomp/isle/pull/207

Implements `MxVideoManager::vtable0x28` and `0x2c`, 100% matches.

They are both very similar. `0x28` uses `MxDisplaySurface::Init` and `0x2c` uses `MxDisplaySurface::Create` instead. `m_unk60` indicates which one of the methods has been used to initialize the class. I can't really think of good names right now so I'll leave them as-is. 

Note: the usage of the `goto` pattern is intentional, just like in `MxDisplaySurface`. I've also added this pattern to `MxMusicManager::StartMIDIThread` in order to achieve a 100% match. There's no other way to get the 100% match I believe (also the usage of this pattern is confirmed by the 1996 source).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,6 +159,7 @@ add_library(lego1 SHARED
   LEGO1/mxpresenterlist.cpp
   LEGO1/mxramstreamcontroller.cpp
   LEGO1/mxramstreamprovider.cpp
+  LEGO1/mxregion.cpp
   LEGO1/mxscheduler.cpp
   LEGO1/mxsemaphore.cpp
   LEGO1/mxsmkpresenter.cpp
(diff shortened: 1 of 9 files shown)


itsmattkc:
> They are both very similar. 0x28 uses MxDisplaySurface::Init and 0x2c uses MxDisplaySurface::Create instead. m_unk60 indicates which one of the methods has been used to initialize the class.

How strange... they really are the same function, just with that very slight difference.

I have a strong feeling `0x2c` is called `Create` because it gets called just after construction in both `LegoOmni::Create` and `MxOmni::Create` (and was a fairly common naming scheme throughout the "Manager"s). I couldn't say why there's a second one at `0x28` though...

Otherwise, this seems good to me.

PR #209, opened 2023-10-14
MxTransitionManager Pixelation transition

disinvite:
Some work towards the pixelation transition function. aka: the default transition in the game if you don't change it with Rebuilder.

The main thing that keeps this from matching is the pixel plot section that begins at `0x1004c07f`. If we are using 8-bit color, we need to replace 10 bytes with the pixelated color, and 20 bytes for 16-bit. The problem is the way that the color value `DWORD` is prepared before we set it in the `lpSurface` pointer. The code (for 8-bit color) looks like this:

```assembly
; DL has the color value we want to set
MOV     AL,DL
MOV     AH,AL       ; AX = DL:DL
MOV     AL,DL       ; unnecessary

MOV     EBX,EAX
SHL     EBX,0x10    ; EBX = DLDL:0
MOV     BX,AX       ; EBX = DLDL:DLDL

; [...]

; EAX is the offset in lpSurface
MOV     dword ptr [EAX],EBX
MOV     dword ptr [EAX + 0x4],EBX
MOV     word ptr [EAX + 0x8],BX
```

The question is: how to do this efficiently? I tried using a bitfield struct to set the MSB and LSB individually, but this didn't work. `windef.h` has the macro `MAKEWORD` that does what you expect, but this will put down the actual `SHL` and `OR` instructions instead of using the registers to do it. The macros are there now to at least illustrate what we're trying to do.

I can't imagine this is inline assembly, but even if it were, why set `AL` again when you don't need to?

Other small stuff:
- Changed the variable names from the previous transition functions to be more expressive. i.e. `i, j` to `col, row`.
- Used macros from `ddraw.h` to replace magic numbers sent in as args for the DirectDraw functions

### DIFF
--- a/LEGO1/mxtransitionmanager.cpp
+++ b/LEGO1/mxtransitionmanager.cpp
@@ -318,7 +318,10 @@ void MxTransitionManager::Transition_Pixelation()
           } else {
             MxU32 new_color = MAKELONG(sample, sample);
 
-            MxU8 *pos = surface + k * ddsd.lPitch + x_shift;
+            // You might expect a cast to MxU16* instead, but lPitch is
+            // bytes/scanline, not pixels/scanline. Therefore, we just
+            // need to double the x_shift to get to the right spot.
+            MxU8 *pos = surface + k * ddsd.lPitch + 2*x_shift;
             MxU32 *dest = (MxU32*)pos;
             // Sets 10 pixels (20 bytes)
             dest[0] = new_color;
--- a/LEGO1/mxtransitionmanager.cpp
+++ b/LEGO1/mxtransitionmanager.cpp
@@ -127,34 +127,34 @@ void MxTransitionManager::Transition_Dissolve()
   memset(&ddsd, 0, sizeof(ddsd));
   ddsd.dwSize = sizeof(ddsd);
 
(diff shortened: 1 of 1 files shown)


LEGO1/mxtransitionmanager.cpp
```diff
@@ -230,10 +230,118 @@ void MxTransitionManager::Transition_None()
...
+            
+            // Sets 10 pixels (10 bytes)
+            dest[0] = new_color;
+            dest[1] = new_color;
+            MxU16 *half = (MxU16*)(dest+2);
+            *half = new_color;
+          } else {
+            MxU32 new_color = MAKELONG(sample, sample);
+
+            MxU8 *pos = surface + k * ddsd.lPitch + x_shift;
```
foxtacles, line 321: I think this should be `x_shift * 2` instead of `x_shift`

foxtacles:
Looks good. I tried a few things but didn't get anything that matches, but functionally it seems correct!

LEGO1/mxtransitionmanager.cpp
disinvite: Good catch, thanks!

PR #211, opened 2023-10-14
MxOmni::Create 100% match, make Create calls consistent

foxtacles:
Mostly refactoring (renaming the "create" functions to `Create` with proper argument names) and several fixes, getting the match of `MxOmni::Create` to 100%.

### DIFF
--- a/LEGO1/mxomnicreateparam.h
+++ b/LEGO1/mxomnicreateparam.h
@@ -14,7 +14,7 @@ public:
   __declspec(dllexport) MxOmniCreateParam(const char *mediaPath, struct HWND__ *windowHandle, MxVideoParam &vparam, MxOmniCreateFlags flags);
 
   const MxOmniCreateFlags& CreateFlags() const { return this->m_createFlags; }
-  const const MxString& GetMediaPath() const { return m_mediaPath; }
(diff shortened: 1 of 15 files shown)


LEGO1/mxomnicreateparam.h
```diff
@@ -14,7 +14,7 @@ class MxOmniCreateParam : public MxParam
   __declspec(dllexport) MxOmniCreateParam(const char *mediaPath, struct HWND__ *windowHandle, MxVideoParam &vparam, MxOmniCreateFlags flags);
 
   const MxOmniCreateFlags& CreateFlags() const { return this->m_createFlags; }
-  const MxString GetMediaPath() const { return m_mediaPath; }
+  const const MxString& GetMediaPath() const { return m_mediaPath; }
```
itsmattkc, line 17: Too many consts?

foxtacles: Oh yeah

itsmattkc:
Excellent work, the `*Omni::Create` functions were the ones I got caught up on in 2019 when I was trying to decomp from raw assembly, so it's great to see these matched. I think it also makes sense to propagate the `Create` name to all of the other managers, so this looks good to merge to me.

PR #213, opened 2023-10-16
Add LegoVehicleBuildState

Andoryuuta:
This is a bit of an odd class which reflects it's class name through a member field set on construction (rather then through proper inheritance). I've named this class `LegoVehicleBuildState` as it doesn't have a single static string associated with it in the binary. Instead it's created for all of the following "class" strings in the `LegoObjectFactory::Create` method:

* `LegoRaceCarBuildState`
* `LegoCopterBuildState`
* `LegoDuneCarBuildState`
* `LegoJetskiBuildState`

New functions/method match rates:
```
LegoVehicleBuildState::LegoVehicleBuildState (0x10025f30) is 100.00% similar to the original
LegoVehicleBuildState::UnkStruct::UnkStruct (0x10017c00) is 100.00% similar to the original
LegoVehicleBuildState::ClassName (0x10025ff0) is 100.00% similar to the original
LegoVehicleBuildState::IsA (0x10026000) is 100.00% similar to the original
```

### DIFF
--- a/LEGO1/legovehiclebuildstate.h
+++ b/LEGO1/legovehiclebuildstate.h
@@ -1,0 +1,61 @@
+#ifndef LEGOVEHICLEBUILDSTATE_H
+#define LEGOVEHICLEBUILDSTATE_H
+
+#include "legostate.h"
+#include "mxstring.h"
+#include "decomp.h"
+
+#include "decomp.h"
+
+// VTABLE 0x100d66e0
+// SIZE 0x50 (from 1000acd7)
+class LegoVehicleBuildState : public LegoState
+{
+public:
+  LegoVehicleBuildState(char* p_classType);
+
+  // OFFSET: LEGO1 0x10025ff0
+  inline virtual const char *ClassName() const override // vtable+0x0c
+  {
+    return this->m_className.GetData();
+  }
+
+  // OFFSET: LEGO1 0x10026000
+  inline virtual MxBool IsA(const char *p_name) const override // vtable+0x10
+  {
+    return !strcmp(p_name, this->m_className.GetData()) || LegoState::IsA(p_name);
+  }
+
+public:
+  struct UnkStruct
+  {
+    undefined4 m_unk0;
+    undefined2 m_unk1;
+    undefined2 m_unk2;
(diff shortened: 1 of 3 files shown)


LEGO1/legovehiclebuildstate.h
```diff
@@ -0,0 +1,61 @@
...
+  // OFFSET: LEGO1 0x10026000
+  inline virtual MxBool IsA(const char *p_name) const override // vtable+0x10
+  {
+    return !strcmp(p_name, this->m_className.GetData()) || LegoState::IsA(p_name);
+  }
+
+public:
+  struct UnkStruct
+  {
+    undefined4 m_unk0;
```
foxtacles, line 32: Could you specify the members' offsets instead of numbering them sequentially? Thank you!

LEGO1/legovehiclebuildstate.h
```diff
@@ -0,0 +1,61 @@
...
+    undefined2 m_unk1;
+    undefined2 m_unk2;
+    undefined2 m_unk3;
+    undefined2 _padding;
```
foxtacles, line 36: No need to specify padding that should be added by the compiler implicitly

LEGO1/legovehiclebuildstate.h
```diff
@@ -0,0 +1,61 @@
...
+  // * LegoDuneCarBuildState
+  // * LegoJetskiBuildState
+  MxString m_className; // 0x38
+
+  // Known States:
+  // * 1 == enter(ing) build screen
+  // * 3 == cutscene/dialogue
+  // * 6 == exit(ing) build screen
+  MxU32 m_animationState; // 0x48
+  undefined m_unk2; // 0x4c
```
foxtacles, line 55: Same here, if you can put that offset into the variable name (formatted `m_unk0x4c` etc.) that would be great

Andoryuuta: Updated the to use include the offsets like `m_unk4c` (without `0x` to match the rest of the code).

foxtacles:
Interesting, looks good. Thank you!

PR #215, opened 2023-10-16
implement GasStationState::GasStationState()

Ramen2X:
Implements GasStationState::GasStationState(). This is not matching due to register swapping, and I wasn't able to remedy it successfully.

This PR also adds a size assert for LegoState as we were able to confirm it is a small class at only `0x08`. All classes that derive from it are expected to create their own members.


foxtacles:
Looks good! I matched it to 100%. `m_unk0x08` almost certainly is a structure but it remains unknown for now. The last member was also `undefined2` instead of `undefined`. Also added a size assert for the class.

Ramen2X:
Nice catch! Thanks!

PR #217, opened 2023-10-17
fix LegoState::VTable0x1C fake-match & implement LegoFileStream::FUN_10006030

Ramen2X:
This PR corrects LegoState::VTable0x1C to take a `LegoFileStream *` parameter instead of the initially assumed `LegoState *` parameter that happened to coincidentally work out while still retaining its' match. I've confirmed this is the correct class through dynamic analysis. It also implements LegoFileStream::FUN_10006030 (moved from `LegoState`) which is currently a 75% match. This function is tricky and I wasn't able to improve it further from around an hour's worth of testing.

### DIFF
--- a/LEGO1/legostate.cpp
+++ b/LEGO1/legostate.cpp
@@ -25,3 +25,16 @@ MxResult LegoState::VTable0x1C(LegoFileStream *p_legoFileStream)
   }
   return SUCCESS;
 }
+
+// OFFSET: LEGO1 0x10006030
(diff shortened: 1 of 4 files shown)


foxtacles:
I've matched `FUN_10006030` to 100% and also moved it back to its original position. Even though it's counterintuitive at first since it's a member function of `LegoFileStream`, the original location of the function is literally next to `VTable0x1C` and thus suggests that they defined the implementation in the `legostate.cpp` unit, so I think we should leave it there as well.

PR #218, opened 2023-10-17
MxRegion: initial mplementation

foxtacles:
Ongoing process of implementing `MxRegion`. I've figured out the list structure, it is indeed a "list of lists" as can be observed in this WIP. 

Unfortunately my graphics programming experience is essentially non-existent so the entire purpose of this class is unclear to me as of now. Names I've assigned to the structures should most likely be changed into something that comes closer to the original intention.

I'm currently in the process of implementing `0x100c3750` which is quite massive and makes extensive use of `MxList` facilities while traversing the `MxRegionList`.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,23 @@
+{
+    "files.associations": {
+        "functional": "cpp",
+        "deque": "cpp",
+        "algorithm": "cpp",
+        "iterator": "cpp",
+        "memory": "cpp",
+        "utility": "cpp",
+        "iosfwd": "cpp",
+        "cstdio": "cpp",
+        "cstring": "cpp",
+        "cstddef": "cpp",
+        "cstdlib": "cpp",
+        "cwchar": "cpp",
+        "map": "cpp",
+        "set": "cpp",
+        "xtree": "cpp",
+        "new": "cpp",
+        "xiosbase": "cpp",
+        "queue": "cpp",
+        "stack": "cpp"
+    }
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,6 +161,7 @@ add_library(lego1 SHARED
   LEGO1/mxramstreamcontroller.cpp
   LEGO1/mxramstreamprovider.cpp
   LEGO1/mxregion.cpp
+  LEGO1/mxregionlist.cpp
   LEGO1/mxscheduler.cpp
   LEGO1/mxsemaphore.cpp
   LEGO1/mxsmkpresenter.cpp
--- a/LEGO1/mxregion.cpp
+++ b/LEGO1/mxregion.cpp
@@ -1,11 +1,24 @@
 #include "mxregion.h"
 
+#include <limits.h>
+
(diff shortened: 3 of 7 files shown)


foxtacles:
The main challenge right now is to achieve the same inlining patterns as in the original. Right now, more functions are inlined than desired. Getting this right will probably take a good amount of experimentation similarly to the large functions in `ISLE.EXE`

Ramen2X:
Great work! I managed to match `MxRegion::vtable1c`. It's a very annoying swap, but this seems to force the register loads into the correct order.

foxtacles:
> I managed to match `MxRegion::vtable1c`. It's a very annoying swap, but this seems to force the register loads into the correct order.

Thank you! However, I believe the prior inaccuracy was purely due to compiler randomness. If you revert the function to how it was before and then move it just below `FUN_100c5280`, you will get a 100% match.

If you take your version of the function and relocate it below `Clone`, you will get the same inaccuracy as before again.

Almost certainly the function is going to keep meandering between 100% and < 100% until we have literally everything in place and nothing changes anymore. We aren't there yet, major parts of `vtable18` are still inaccurate and the involved header files are probably also going to keep changing for a while.

I guess what I'm trying to say is that in these cases, imo it makes no sense to apply "band aids" that may fix the function to 100% temporarily while it is in a certain constellation, unless it can be proven that it will be fixed permanently, which I think is not the case here.

Ramen2X:
I see. That's interesting, I didn't realize that this was the case. Thanks for letting me know! Is it best to revert this then?

foxtacles:
> I see. That's interesting, I didn't realize that this was the case. Thanks for letting me know! Is it best to revert this then?

Yeah, for the time being I'd revert it. My general rule of thumb is: if there is an inaccuracy that I can completely resolve purely by moving the function somewhere else, it's probably correct and shouldn't need further changes. There may be exceptions to this rule but I've found it generally holds true.

Ramen2X:
All done, sorry for the trouble!

foxtacles:
> All done, sorry for the trouble!

No problem! I should probably write some of this stuff into the `CONTRIBUTING.md` file since all of this rather arcane information is not obvious but important to know when working on the project.

Ramen2X:
> No problem! I should probably write some of this stuff into the `CONTRIBUTING.md` file since all of this rather arcane information is not obvious but important to know when working on the project.

That would probably be for the best. While we're on the subject, I wonder if it would be worth introducing a standardized way of marking the hacks that we do actually employ to force the compiler to generate specific code to work around randomness, so that when the entire code-base is complete, we can go through and remove all of them in hopes that it will all resolve itself and so that we're not left with unnecessarily messy logic riddled in the final product.

Ramen2X:
In any case, in an effort to not derail the conversation in this PR, I feel this is probably good for review. I think further attempts at improving match% can be made post-merge.

Ramen2X (approved):
I initially didn't intend to do this review, but I didn't want this PR to potentially end up in limbo forever. I had quite a thorough look and I believe this is good to merge as-is, however it's still possible I may have missed something.

PR #221, opened 2023-10-19
label ScoreState::VTable0x18 and associated member

Ramen2X:
This PR gives the name `SetScoreCubeTutorialFlag` to `VTable0x18` in ScoreState and labels its associated member. This bool `m_playCubeTutorial` is responsible for determining whether or not the explanation audio for the first time a player visits the Score Cube room plays.


Ramen2X:
Side note: It's probable that this function is actually returning an `MxResult`, but I haven't been able to confirm. I'm unsure what the point of the return value is because the setting of the member isn't nested whatsoever.

PR #222, opened 2023-10-19
Rename virtual function to fix override

foxtacles:
Noticed after meging that the override isn't working. I've given the function a more generic name since it's based in `LegoState`. Maybe its purpose is something like "Activate", but we'd have to figure that out first.


Ramen2X (approved):
Huh, that's odd. But LGTM, thanks!

PR #226, opened 2023-10-21
MxVideoPresenter AlphaMask (formerly UnkStruct)

disinvite:
Implementing the constructor (and destructor) for what I am calling the `AlphaMask` struct inside of `MxVideoPresenter`. My reasoning is: we are walking the bitmap rows in order (regardless of top-down or bottom-up orientation) and setting a bit in the allocated array if the given pixel has a non-zero value. I'm not sure yet how this gets used, but that's outside the scope of this PR.

There is a similar constructor in the `MxVideoPresenter` subclass `MxStillPresenter`. It's a lot shorter, but still should be a separate PR.

The `BITMAPINFOHEADER` struct has a field `biCompression` that can have a few different values. A value of zero corresponds to `BI_RGB`, which means uncompressed. By default, the rows of the bitmap are ordered bottom-up (that is, the lowest row of pixels on the screen comes first in the file). If the `biHeight` field is negative, this means that the bitmap is top-down (pixels appear in screen order).

The devs added a non-standard value of `0x10` for `biCompression`; if it is found, we assume the bitmap has top-down order regardless of the sign of `biHeight`. I added a macro `BI_RGB_TOPDOWN` for this case. I didn't check the SI files to see if any actual `MxBitmap` data (or video, I guess) uses this alternate value, but maybe somebody can speak to it.

The code in this constructor that checks `biCompression` and `biHeight` is a little bit of a mess. I used the dreaded `goto` but even that doesn't completely match. (Hopefully the intent is clear.) There is probably some optimization here that I'm missing, but I've spent enough time on this and I want to throw it out to the group.

I added some inlines to `MxBitmap` to simplify certain actions, including the `Abs` re-implementation from back in #195. I haven't dropped them in everywhere yet.

### DIFF
--- a/LEGO1/mxbitmap.cpp
+++ b/LEGO1/mxbitmap.cpp
@@ -369,7 +369,7 @@ MxResult MxBitmap::SetBitDepth(MxBool p_isHighColor)
 MxResult MxBitmap::StretchBits(HDC p_hdc, MxS32 p_xSrc, MxS32 p_ySrc, MxS32 p_xDest, MxS32 p_yDest, MxS32 p_destWidth, MxS32 p_destHeight)
 {
   // Compression fix?
-  if ((this->m_bmiHeader->biCompression != 16) && (0 < this->m_bmiHeader->biHeight)) {
+  if ((this->m_bmiHeader->biCompression != BI_RGB_TOPDOWN) && (0 < this->m_bmiHeader->biHeight)) {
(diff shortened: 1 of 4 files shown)


foxtacles:
Nice work, looks good to me. I'm pretty confident the `goto` part can be eliminated, the line after the label `seek_to_last_row` probably needs to be duplicated as well, and then it's "only" a matter of trying out all the possible permutations of the branching logic until the right one comes out.

That being said, I gave it a try for like 10 minutes and didn't get the right one, but I think this is good to merge.

PR #228, opened 2023-10-21
Checkorder tool to keep functions in original binary order

disinvite:
A while back on Matrix chat I posed the question of whether MSVC would (or even could) reorder functions within a translation unit. To put it another way: is the order of the code in the file the only deciding factor on the order of functions in the binary?

Maybe that was a no-brainer question; seems like the consensus is yes, code order decides binary order.

We have some work to do with the code we already have, so I created this python script to check each file and tell if the `// OFFSET: LEGO1 ____` comments are in order. If the `--enforce` option is used, this will exit with an error status (to break the build) like the `verexp` tool. We probably don't want to enforce until we get the files to a good place.

It's easy to report that a file is in order or not, but what else can we do? Right now, the `--verbose` option just tells when each run of offsets in order gets broken, along with number of lines and expected position of each function.

I debated whether the tool should provide some assistance to the user about how to fix the order, but what information would be helpful? Would we want to do this in incremental commits so the diffs don't turn into soup? Do we care about git blame on each person's work?


foxtacles:
Looks really useful to me. A few remarks:

* I think we should start enforcing this as soon as possible. One idea I have is that the GitHub build could run this script for the translation units that have been changed/touched in the respective branch (compared to `master`) only. This way, it is the contributor's responsibility to make sure that the syntax and order of our "decomp" declarations is correct in the changeset they submitted.
* The script currently has `LEGO1` hardcoded. I know that `ISLE` is a minor part of the project but maybe it can be extended/changed to be "module"-agnostic.
* `is_exact_offset_comment` is currently unused, right? I think it would be useful to have this enabled as well and have the information be reported to the user, maybe even as an error. I do believe we should start being more strict with the style, especially if we have a tool for automatic verification.

disinvite:
Thanks for the feedback! We now match any alphanumeric for the module name. This change has revealed that `legoentitypresenter.cpp` has `LEG01` with a zero for the O. 😅

Offset comments that fail the exact match are displayed in verbose mode.

disinvite:
Do we want any more features for this? Or pull out the file parsing bit into a python module to load in a few smaller scripts? I can do both, but I wasn't sure if it was good idea to jam it all into this PR. Some ideas:

- Should be easy to tell if a function marked STUB has been built out. Probably a non-fatal warning like the "bad offset comment" message
- Tracking of `// TODO` and other similar comments so we can turn those into `// TODONE`
- Identify functions that do _not_ have the `// OFFSET` comment. Would need a better regular expression or heuristic to match a C++ function signature. Might be more trouble than it's worth since I doubt there are many cases of this
- Tracking global variable offsets can be done but we would have to decide on a format for the comments
- What order to use for the _linker_? (i.e. not just tracking the offsets in a single file.) Would suggest an order for the CMakeLists.txt file. If the devs used an MSVC project like we started with, can you adjust the order of the modules in that? Does it just use alphabetical order?
- Lastly, the obvious thing: re-order the pieces of the file so the offsets are in order. Trivial to do if the code blocks are all contiguous. Maybe that's good enough?

And a question: should the exact match on offset comment just be `ISLE` and `LEGO1` for module name, instead of `[A-Z0-9]+`? Would we ever build any other modules in this repo? `OMNI` ?

foxtacles:
I'd say we should limit this PR to providing the "order detection" feature. The ideas are great and I'm sure we can implement some of them later.

However, I do think it would be great indeed if we can refactor the parsing of the decomp declarations into a module that can then be used by both `reccmp` and `checkorder` and other scripts in the future. That module should probably parse all `.h` and `.cpp` files (just like `reccmp`) and provide all the decomp comments and associated metadata for each file, which can then be further consumed by all scripts that need it. I think it's worth doing it in this PR.

disinvite:
The refactor into the `isledecomp` module is done. No new features yet, and the `checkorder` script stands to get a little simpler. I invite any python experts to check my work here.

foxtacles:
> The refactor into the `isledecomp` module is done. No new features yet, and the `checkorder` script stands to get a little simpler. I invite any python experts to check my work here.

Looks good! Do you think you could adjust `reccmp` to make use of the module? Thanks!

disinvite:
Getting this ready to work with `reccmp` has revealed some flaws in my design, but this is good. We are in much better shape now. I have the `reccmp` code stashed in case we want to merge #256 first.

There are two minor differences in the `reccmp` output but it will be simpler to just fix the source files in question rather than build out more edge cases. It would affect the match percentage slightly because the current implementation reports these cases as failures to match anything. My code does match something (but does so incorrectly).

And those cases are:

- legonavcontroller.cpp, where a function with `// OFFSET` marker is commented out entirely
- mxdirectdraw.cpp, where a static variable inside the `Error()` function has the `// OFFSET` marker

itsmattkc:
I apologize it's taken me so long to look at this PR. These past few weeks have been a little busy for me and I haven't had a decent moment to look at all the changes.

I hadn't realized when I made #256 that this had evolved into a general purpose library for parsing our codebase. I think this is a good idea, and from what I can see, it looks fairly comprehensive and well put-together. I might have to see how `reccmp` would use it to really get an idea of how it all works though, and I'd be happy to do that first and rework #256 to fit later on. I'd actually originally planned to do #256 as a separate script, but folded it into `reccmp` so I could take advantage of the existing code/PDB parsers. If we spin that stuff off into libraries, I may go back to my original plan of making a separate script.

My only real note on the `checkorder.py` script is that perhaps the term `target` is too ambiguous, something like `target-dir` or `source-dir` might be more immediately intuitive. Otherwise it seems to work as expected. I agree with @foxtacles that we may as well enforce this early so we don't have to worry about it later on.

disinvite:
I just added the integration with reccmp, but there's a slight problem with running it on github as you see. What do I need to have in requirements.txt to point at the local `isledecomp` directory? I tried just `-e ../isledecomp` on another branch but it didn't like that.

I used `target` as the parameter name for `checkorder` since you can provide either a file or directory. I can change it to `target-dir` if it's more clear; scanning everything will be the major use case anyway.

foxtacles:
```
@@ -18 +17,0 @@
-  MxDSObject::SetAtomId (0x401c40) is 100.00% similar to the original
@@ -20 +19 @@
```

Looks like there is a small regression, maybe that's because there are two offset comments for the same function? I think this might be the only case in the entire codebase

tools/reccmp/reccmp.py
```diff
@@ -414,143 +416,119 @@ def can_resolve_register_differences(original_asm, new_asm):
 basename = os.path.basename(os.path.splitext(original)[0])
 pattern = '// OFFSET:'
```
foxtacles, line 417: This can probably be removed?

disinvite: pattern yes. basename is used inside the loop to match the binary filename passed in against the module name (i.e. `// OFFSET: <this-part> 0x1234`) from the offset markers.

This could be related to the previous regression; reccmp was pulling in every .cpp and .h file in the entire repo, which included my unit test files added here. I changed it around to pull from the individual ISLE and LEGO1 directories instead, but for ISLE we _do_ want to compare the common functions from LEGO1. I switched it back and just modified all my test files instead, but maybe something is still wrong.

I am also a few commits behind the master so we are probably comparing the saved match percentage against stale code here.

disinvite: Here's why we have a regression:
```cpp
// OFFSET: ISLE 0x401c40
// OFFSET: LEGO1 0x10005530
inline virtual void SetAtomId(MxAtomId p_atomId) { this->m_atomId = p_atomId; } // vtable+20;
```
Two offset markers, and my parser isn't set up for that. I'll have to think about how to best handle this.

disinvite:
We should be good on `mxdsobject.h` or any future files that use multiple `// OFFSET` markers as far as `reccmp` is concerned.

New problem: Because the parser now returns any offsets it sees, `checkorder` will have to isolate on module name to give accurate results.

I also wonder how useful it is to check whether the `TEMPLATE` markers are in order between any surrounding movable code blocks. It's nice if they're in order for style points. Maybe only contiguous blocks of template functions should be checked for order?

foxtacles:
Looks good!

> I also wonder how useful it is to check whether the `TEMPLATE` markers are in order between any surrounding movable code blocks. It's nice if they're in order for style points. Maybe only contiguous blocks of template functions should be checked for order?

I think that would be useful mainly for documentation purposes. Maybe it would also be a good idea to move the `TEMPLATE` declaration blocks to `.cpp` files and order them together with the rest of the functions.

For example, the inlined list functions are placed by the compiler into the `mxregion.cpp` address space in between the functions: 

https://github.com/isledecomp/isle/blob/master/LEGO1/mxregionlist.h#L47
https://github.com/isledecomp/isle/blob/master/LEGO1/mxregion.cpp 

It would make sense to me to also see those declarations in the right places there, the idea being to have a "full" 1:1 contiguous accounting of all the functions in the right order as they appear in the assembly/module.

.github/workflows/build.yml
```diff
@@ -70,6 +70,11 @@ jobs:
         path: legobin
         key: legobin
 
+    - name: Build isledecomp library
```
tntexplosivesltd, line 73: Build or install?

tools/checkorder/checkorder.py
```diff
@@ -0,0 +1,124 @@
+import os
+import sys
+import argparse
+from typing import TextIO
```
tntexplosivesltd, line 4: Not used in this file?

tools/isledecomp/isledecomp/parser/util.py
```diff
@@ -0,0 +1,96 @@
...
+# flexibility in the formatting seems OK
+templateCommentRegex = re.compile(r'\s*//\s+(.*)')
+
+
+# To remove any comment (//) or block comment (/*) and its leading spaces
+# from the end of a code line
+trailingCommentRegex = re.compile(r'(\s*(?://|/\*).*)$')
+
+
+def template_function_name(line: str) -> str:
```
tntexplosivesltd, line 37:
```suggestion
def get_template_function_name(line: str) -> str:
```

tools/isledecomp/isledecomp/parser/util.py
```diff
@@ -0,0 +1,96 @@
...
+    """Parse function signature for special TEMPLATE functions"""
+    template_match = templateCommentRegex.match(line)
+
+    # If we don't match, you get whatever is on the line as the signature
+    if template_match is not None:
+        return template_match.group(1)
+    else:
```
tntexplosivesltd, line 44: Unnecessary `else` after `return`

.github/workflows/build.yml
disinvite: Well... both.

foxtacles:
Added all the fixes and a GH action to enforce `checkorder` here: https://github.com/disinvite/isle/pull/2

I think this should be good to merge afterwards.

disinvite:
One last thing: `reccmp` now depends on the `isledecomp` library so everyone will have to install it with pip.
```
pip install -e tools/isledecomp
```

PR #234, opened 2023-10-22
cleanup: fix all improper uses of MxResult

Ramen2X:
This PR fixes all uses of MxResult to use its' defined names rather than hardcoded values.

### DIFF
--- a/LEGO1/mxioinfo.cpp
+++ b/LEGO1/mxioinfo.cpp
@@ -124,9 +124,9 @@ MxLong MXIOINFO::Read(void *p_buf, MxLong p_len)
 }
 
 // OFFSET: LEGO1 0x100cca00
-MxResult MXIOINFO::Seek(MxLong p_offset, int p_origin)
+MxLong MXIOINFO::Seek(MxLong p_offset, int p_origin)
 {
-  MxResult result = FAILURE;
+  MxLong result = -1;
 
   // If buffered I/O
   if (m_info.pchBuffer) {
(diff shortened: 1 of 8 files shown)


Ramen2X:
Also noticed `MXIOINFO::Seek()` was returning an MxLong despite sticking to the MxResult standard, so I've corrected this too. Match of the function remains the same.

LEGO1/mxioinfo.cpp
```diff
@@ -124,9 +124,9 @@ MxLong MXIOINFO::Read(void *p_buf, MxLong p_len)
 }
 
 // OFFSET: LEGO1 0x100cca00
-MxLong MXIOINFO::Seek(MxLong p_offset, int p_origin)
+MxResult MXIOINFO::Seek(MxLong p_offset, int p_origin)
```
foxtacles, line 127: This should remain `MxLong`. If you look at the function you will see it assigns various values to `result` which don't necessarily have to be `SUCCESS` or `FAILURE`, so using `MxResult` as the return type would be misleading

Ramen2X: Sorry! I didn't notice this. I think I was tripped up by the `result` name.

foxtacles: I'd just leave it `-1` and only use the constants in the context of the `MxResult` type (think C++11 `enum class` behavior)

foxtacles (approved):

PR #240, opened 2023-10-23
(Proposal) Introduce clang-format to enforce consistent code style

foxtacles:
Since we've gained several contributors over the last few weeks (which is great), the issue regarding code style is coming up more and more often - so far we've been manually verifying that all contributions are more or less in line with the (implicit) style rules we've been adhering to so far.

However, this becomes increasingly unfeasible since the current manual process requires spending a considerable amount of extra time, not to mention that the lack of explicit guidelines adds to the workload of every contributor.

Personally, I believe we have to introduce the following scripts/checks to enhance the collaboration process:

* Verify formatting of the code - addressed by this PR using `clang-format`
* Verify our decomp declarations - (partially) addressed by https://github.com/isledecomp/isle/pull/228
* Verify naming conventions (i.e. `m_unk0x00` and https://github.com/isledecomp/isle/blob/master/CONTRIBUTING.md#code-style )

This PR introduces the `clang-format` tool, a `.clang-format` file that specifies the rules and a GitHub action that verifies that the code is formatted according to the rules (and fails if there are any issues).

`clang-format` has many integrations with popular editors and IDEs, so adoption should be relatively straightforward for everyone.

The set of rules is based on the `Microsoft` preset, since it comes closest to the style we've been using so far. On top of that, `.clang-format` specifies rules that are custom to this project. These are based on how we have written code ~80-90% of the time so far, since the goal is to solidify the existing, implicitly used style instead of introducing a new one.

So far the output looks pretty good to me. Let me know what you think. Here's a list of all options: https://clang.llvm.org/docs/ClangFormatStyleOptions.html - feel free to suggest changes.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,13 @@
 root = true
-[*.{cpp,h,py,txt,editorconfig}]
 
+[*.{py,txt,editorconfig}]
 indent_style = space
 indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{cpp,h}]
+indent_style = tab
+tab_width = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,25 @@
+BasedOnStyle: Microsoft
+
+AccessModifierOffset: -4
+AlignAfterOpenBracket: BlockIndent
+AllowAllArgumentsOnNextLine: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortFunctionsOnASingleLine: InlineOnly
+AlwaysBreakTemplateDeclarations: Yes
+BinPackArguments: false
+BinPackParameters: false
+BracedInitializerIndentWidth: 4
+BraceWrapping:
+  AfterClass: false
+  AfterControlStatement: Never
+  AfterEnum: false
+  AfterStruct: false
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+IncludeBlocks: Regroup
+IndentAccessModifiers: false
+IndentWidth: 4
+PointerAlignment: Left
+SpaceAfterCStyleCast: true
+TabWidth: 4
+UseTab: ForContinuationAndIndentation
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,20 @@
+name: Format
+
+on: [push, pull_request]
+
+jobs:
+  clang-format:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Run clang-format
+      run: |
+        pipx run "clang-format>=17,<18" \
+          --Werror \
+          --dry-run \
+          --style=file \
+          ISLE/*.cpp ISLE/*.h \
+          LEGO1/*.cpp LEGO1/*.h \
+          LEGO1/realtime/*.cpp LEGO1/realtime/*.h
--- a/ISLE/define.cpp
+++ b/ISLE/define.cpp
@@ -1,7 +1,7 @@
 #include "define.h"
 
 // 0x410030
-IsleApp *g_isle = 0;
+IsleApp* g_isle = 0;
 
 // 0x410034
 unsigned char g_mousedown = 0;
--- a/ISLE/define.h
+++ b/ISLE/define.h
@@ -6,7 +6,7 @@
 
 class IsleApp;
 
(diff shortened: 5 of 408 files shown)


tahg:
Seems like Microsoft is all about the line count. In all seriousness I think it looks fairly good. I would set `AllowShortCaseLabelsOnASingleLine` back to true. (Only MS enforces a new line, like it does for most other places.)  Aside from that, I personally prefer `SpaceAfterCStyleCast` to false, so that it's adjacent to the variable, and aligns with the style of a * or & on a variable, but I don't have a really strong opinion on it one way or the other.

disinvite:
Never seen a `case` _not_ get indented after a `switch` until reading the 96 source. My preference would be `IndentCaseLabels: true` though it goes against the available evidence.

tahg:
> Never seen a `case` _not_ get indented after a `switch` until reading the 96 source. My preference would be `IndentCaseLabels: true` though it goes against the available evidence.

I've just about gotten used to it. You can't convince me that it should be different from public:, protected:, private:; whichever way it goes.

itsmattkc:
I think this is a good idea. My biggest suggestion is that this is probably a good opportunity to introduce the original 1996 source's style of indentation (namely 4-width tabs instead of 2-width spaces).

The 1996 source in fact includes a style document (`STYLE.DOC`) that might be good to double check with. Even though it doesn't affect anything accuracy-wise, if we know what the original code style was, we may as well use it ourselves.

foxtacles:
> I think this is a good idea. My biggest suggestion is that this is probably a good opportunity to introduce the original 1996 source's style of indentation (namely 4-width tabs instead of 2-width spaces).
> 
> The 1996 source in fact includes a style document (`STYLE.DOC`) that might be good to double check with. Even though it doesn't affect anything accuracy-wise, if we know what the original code style was, we may as well use it ourselves.

Sounds good, I'll look into that and try to come as close as possible.

foxtacles:
Okay, so I've looked at `STYLE.DOC` and modified the rules to be based on that style guide. It's noteworthy however that those rules have been applied rather inconsistently (probably because they didn't have any tooling to help with it).

Major changes to the style now include:

* Tabs instead of spaces (4-width) for indentation/continuation
* `class/struct/enum` braces on first line
* Pointer alignment is left

Ramen2X:
This isn't really a code-related style issue, but the 1996 source also makes it evident that all of their filenames were created in a strict UPPERCASE format versus the all-lowercase approach we employ here. Is this something also worth adopting here purely for environmental accuracy? It doesn't really matter in the grand scheme of things, but I thought I would bring it up anyway.

foxtacles:
> This isn't really a code-related style issue, but the 1996 source also makes it evident that all of their filenames were created in a strict UPPERCASE format versus the all-lowercase approach we employ here. Is this something also worth adopting here purely for environmental accuracy? It doesn't really matter in the grand scheme of things, but I thought I would bring it up anyway.

Yeah, we can consider that. I'm not sure how many implications it would have with regards to git history etc. so I think it would be better if we decide for it to do it in a separate PR

.editorconfig
```diff
@@ -1,7 +1,7 @@
...
 [*.{cpp,h,py,txt,editorconfig}]
 
-indent_style = space
-indent_size = 2
+indent_style = tab
+tab_width = 4
 insert_final_newline = true
-trim_trailing_whitespace = true
\ No newline at end of file
+trim_trailing_whitespace = true
```
madebr, line 7: Let's keep the cmake and python scripts space indented.
(their indentation size can be changed in a later commit)

```suggestion
[*.{py,txt,editorconfig}]
indent_style = space
indent_size = 2
insert_final_newline = true
trim_trailing_whitespace = true

[*.{cpp,h}]
indent_style = tab
tab_width = 4
insert_final_newline = true
trim_trailing_whitespace = true
```

itsmattkc:
This seems like it should be good to merge. As @foxtacles says, there will be stuff in this style that we don't all necessarily agree with, but it's worthwhile to have consistent rules, and the original source is as good of a template to follow as any.

I think this gets the big stuff in. We can always tweak it more later on if we need to.

> Seems like Microsoft is all about the line count.

Heh well that is technically how they got paid early on: https://www.youtube.com/watch?v=kHI7RTKhlz0

> the 1996 source also makes it evident that all of their filenames were created in a strict UPPERCASE format versus the all-lowercase approach we employ here

Personally I think we shouldn't adopt uppercase filenames because it could introduce issues later on on non-Windows platforms (e.g. ext4 on Linux/Android is case sensitive, so we'd have to ensure all of our `#include`s are uppercase too to ensure they match the files, which feels like a hassle).

PR #241, opened 2023-10-24
implement/match CalcLocalTransform

Ramen2X:
This PR implements the global function `CalcLocalTransform`. To supplement this (and hopefully to aid the completion of other 3D functions later on), I've included a header named `vec.h` which is comprised of hundreds of extremely useful macros designed for dealing with 3D elements. This code has been publicly released as a part of the [GraphicsGems](https://github.com/erich666/GraphicsGems) collection and the license is permissive of this use-case. I have also added an additional macro which normalizes a three-dimensional vector, which is also used by `CalcLocalTransform`.

This function is currently 100% matching. Keep in mind however that this function is at high risk of de-matching due to MSVC entropy; I noticed while attempting to hit 100% that the placement of the function in the implementation file affects the match percentage.


Ramen2X:
Hate to modify code out of the scope of this PR, but I had to make a slight tweak to `MxVideoPresenter::AlphaMask::AlphaMask` to get this to build for some reason. I've confirmed this does not change the accuracy of said function.

foxtacles:
It looks like the 1996 leak has an even newer version of `VEC.H` than the one you included? Maybe we should take that one?

image

I think the folder `3rdparty` would also be a better place to put it. Looks good otherwise!

LEGO1/mxvideopresenter.cpp
```diff
@@ -135,7 +135,7 @@ MxVideoPresenter::AlphaMask::AlphaMask(const MxBitmap &p_bitmap)
     for (MxS32 i = 0; i < m_width; i++) {
       if (*t_ptr) {
         // TODO: Second CDQ instruction for abs() should not be there.
-        MxU32 shift = abs(offset) & 7;
+        MxS32 shift = abs(offset) & 7;
```
foxtacles, line 138: This seems to be because of conflicting overloads since `VEC.H` pulls in `math.h` and there is another `abs` defined in `stdlib.h`. We should probably fix this as otherwise it won't be possible anymore to call `abs` on an unsigned integer in cases like this. Let me think what would be best here.

foxtacles: I guess the easiest would just be to cast `shift` to `MxS32` in the next line (the call to `abs`) to resolve the conflict. That's what the compiler should have done implicitly before the introduction of `math.h` anyway.

Ramen2X:
Moving `CalcLocalTransform` to another file de-matched it as I suspected. Fortunately, I was able to employ another unroll trick to hopefully fix the random de-match once and for all.

foxtacles (approved):

PR #243, opened 2023-10-24
Implement Helicopter::VTable0xe4

tahg:
Clean up parent classes


tahg:
Quite a few changed files again, I'm sorry. Barely any code changes, but this is what you get when you start at the beginning of the DLL instead of logically doing a parent class before doing their children.

I also decided to organize the classes I was touching anyway.

This try should be properly synched now. I hate git sometimes.

foxtacles (approved):
Looks good!

PR #244, opened 2023-10-24
MxVideoPresenter::IsHit

disinvite:
(This is ready, but I'll keep it as a draft for now until we get the clang-format pull request done.)

Implements IsHit() for MxVideoPresenter. If available, the code will use the AlphaMask from #226 as a memoization for this check.

The match percentage here isn't great. I don't know why the original uses so many stack variables. Any ideas are welcome.


foxtacles (approved):
Looks good! I applied a minor logic fix. With regards to the stack, one thing that comes to mind is that they maybe used `MxPoint32` / `GetLocation` in some places instead of the individual x/y values - that might change the stack space used.

PR #245, opened 2023-10-24
MinGW build fixes

madebr:
These are build fixes to get isle building (further) with a modern mingw toolchain.

The only errors left are those due to taking the address of rvalues.
[These can be fixed easily](https://github.com/madebr/isle/commit/9dde9e7f9d3a2775b565a2fa9504e04afc78bf89), but it decreases the matching percentage.
Therefore, not part of this pr.

### DIFF
--- a/LEGO1/mxhashtable.h
+++ b/LEGO1/mxhashtable.h
@@ -10,6 +10,9 @@
 #define HASH_TABLE_OPT_EXPAND_MULTIPLY    2
 
 template <class T>
+class MxHashTableCursor;
+
+template <class T>
 class MxHashTableNode
 {
 public:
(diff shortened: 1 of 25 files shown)


LEGO1/mxobjectfactory.cpp
```diff
@@ -28,16 +28,14 @@ MxObjectFactory::MxObjectFactory()
 // OFFSET: LEGO1 0x100b12c0
 MxCore *MxObjectFactory::Create(const char *p_name)
 {
-  MxCore* object = NULL;
```
foxtacles, line 31: It looks like there is a a major accuracy regression with this one

LEGO1/mxhashtable.h
```diff
@@ -9,6 +9,11 @@
 #define HASH_TABLE_OPT_EXPAND_ADD         1
 #define HASH_TABLE_OPT_EXPAND_MULTIPLY    2
 
+template <class T>
+class MxHashTableCursor;
+template <class T>
+class MxList;
```
foxtacles, line 15: Is it supposed to be `MxHashTable`?

madebr: Only `MxHashTableCursor` is needed here. I got confused by the missing forwarded declarations in `mxlist.h`.

LEGO1/mxobjectfactory.cpp
madebr: I reverted the change. It was not needed for mingw compatibility.

foxtacles (approved):
Looks good!

PR #246, opened 2023-10-25
Rename some functions and members

tahg:
As usual, the naming changes affected quite a few files :( And I tried to improve some matching of files I was hitting anyways.


LEGO1/mxomni.cpp
```diff
@@ -278,18 +278,23 @@ MxResult MxOmni::Create(MxOmniCreateParam& p)
 // OFFSET: LEGO1 0x100afe90
 void MxOmni::Destroy()
 {
-	// FIXME: Stub
+	{
```
ItzSwirlz, line 281: fix indentation

LEGO1/legoinputmanager.cpp
```diff
@@ -51,9 +51,9 @@ void LegoInputManager::Destroy()
 {
 	ReleaseDX();
 
-	if (m_unk0x5c)
-		delete m_unk0x5c;
-	m_unk0x5c = NULL;
+	if (m_eventQueue)
```
ItzSwirlz, line 54: maybe put brackets here

LEGO1/legoworld.cpp
```diff
@@ -26,12 +34,35 @@ void LegoWorld::Stop()
 	TickleManager()->UnregisterClient(this);
 }
 
+// OFFSET: LEGO1 0x1001f5e0
+MxLong LegoWorld::Notify(MxParam& p_param)
```
ItzSwirlz, line 38: This *may* be an MxResult though I think other :Notify functions are MxLong's, so check on that

LEGO1/legoentity.cpp
```diff
@@ -36,35 +28,53 @@ void LegoEntity::Reset()
...
 }
 
+// OFFSET: LEGO1 0x10010650 STUB
+void LegoEntity::ResetWorldTransform(MxBool p_inVehicle)
+{
+	// TODO
+}
+
+// OFFSET: LEGO1 0x10010790 STUB
+void LegoEntity::SetWorldTransform(MxVector3& loc, MxVector3& dir, MxVector3& up)
```
foxtacles, line 38: Missing p_ prefix

LEGO1/legoentity.h
```diff
@@ -34,34 +35,34 @@ class LegoEntity : public MxEntity {
...
 	// OFFSET: LEGO1 0x10001090
-	virtual void VTable0x30(undefined4 p_unk50) { m_unk50 = p_unk50; } // vtable+0x30
-	virtual void VTable0x34();                                         // vtable+0x34
-	virtual void VTable0x38();                                         // vtable+0x38
-	virtual void VTable0x3c();                                         // vtable+0x3c
-	virtual void VTable0x40();                                         // vtable+0x40
-	virtual void VTable0x44();                                         // vtable+0x44
-	virtual void VTable0x48();                                         // vtable+0x48
-	virtual void VTable0x4c();                                         // vtable+0x4c
+	virtual void SetWorldSpeed(MxFloat p_speed) { m_worldSpeed = p_speed; } // vtable+0x30
```
foxtacles, line 45: Preferably we give the parameter the same name as the member

LEGO1/mxomni.cpp
foxtacles: This is now governed by `clang,-format`. If the action succeeds, nothing need be changed

LEGO1/legoworld.cpp
foxtacles: `MxResult` is a typedef for `MxLong`. However, unless we have enough evidence that all Notify functions restrict themselves to returning one of the two `MxResult` constants, we should leave it as-is

LEGO1/legoentity.h
tahg: > Preferably we give the parameter the same name as the member
Does this apply even for code that was in the 1996 source like this was?

foxtacles: Good question. Personally I'd say yes, it's fine if we introduce some minor extra conventions like this, even if that doesn't equal 100% what we see in the 1996 source. Many things were applied inconsistently in the original source, I don't think taking over inconcistency on this level is worthwhile.

foxtacles (approved):
Looks good to me. Merging when conflicts are solved

PR #248, opened 2023-10-25
MxStillPresenter::ParseExtra

disinvite:
Not an exact match, but wrong for all the usual reasons.

This function does an odd thing with the "extra" string length that also happens in the main `MxPresenter::ParseExtra`. The variable is word-sized, but then we mask out the more significant other bits with an `AND EAX, 0xffff`. I doubt my literal interpretation of that matches the original, but it at least gets the instruction in there.

Is this extra length + char* another structure? Why `memcpy` the data versus the more logical `strcpy`?

### DIFF
--- a/LEGO1/define.cpp
+++ b/LEGO1/define.cpp
@@ -8,3 +8,6 @@ const char* g_strWORLD = "WORLD";
 
 // 0x10102040
 const char* g_strACTION = "ACTION";
+
+// 0x101020cc
+const char* g_strVISIBILITY = "VISIBILITY";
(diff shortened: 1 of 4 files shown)


foxtacles:
> Is this extra length + char* another structure? Why `memcpy` the data versus the more logical `strcpy`?

Yeah, at this point I highly suspect it must be some sort of structure. I've come across too many places where the `extraData`/`extraLength` pair generated funny code or required shenanigans like the following for a match:

https://github.com/isledecomp/isle/blob/master/LEGO1/mxdsaction.cpp#L168

I already tried a basic struct with both members in it but didn't get anywhere. so it's still a bit of a mystery.

foxtacles (approved):
Looks good!

PR #251, opened 2023-10-27
implement/match several mxsmkpresenter functions

Ramen2X:
This PR implements and matches several MxSmkPresenter functions. MxSmkPresenter's constructor and deconstructor are both matching. `MxSmkPresenter::Init()` is also 100% matching. `MxSmkPresenter::FUN_100b3900` was 100% matching when I first completed the implementation, but it unfortunately seems to have de-matched at some point in my work and I haven't experimented with correcting it. `MxSmkPresenter::FUN_100c5d40` never matched, but it looks like it is barely off. `MxSmkPresenter::VTable0x60` was also adjusted to accommodate for the new member changes, and it currently retains its match.

To prepare to start filling out the class, I implemented a minimal version of a Smacker header, which you can find located at `3rdparty/smk/smk.h`. Originally, I had put this struct directly in MxSmkPresenter. However, throughout my research, it has become apparent to me that Mindscape had implemented their own stuff on top of the provided Smacker standard. I've given this the name of `MxSmack`, and it embeds a standard `Smack` inside of it.

Additional side note, the manual null-checks in `MxSmkPresenter::FUN_100c5d40` are intentional and seem to be required to get proper code generation.

### DIFF
--- a/3rdparty/smk/smk.h
+++ b/3rdparty/smk/smk.h
@@ -1,0 +1,54 @@
+#ifndef SMK_H
+#define SMK_H
+
+struct SmackSum {
+	unsigned long m_totalTime;
+	unsigned long m_ms100PerFrame;
+	unsigned long m_totalOpenTime;
+	unsigned long m_totalFrames;
+  	unsigned long m_skippedFrames;
+	unsigned long m_totalBlitTime;
+	unsigned long m_totalReadTime;
+	unsigned long m_totalDecompressTime;
+	unsigned long m_totalBackReadTime;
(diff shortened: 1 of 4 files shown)


foxtacles:
Looks good! I've matched the remaining functions - basically `FUN_100c5d40` had to be a static function.

Its address does suggest that the function was most likely in a separate translation unit and not part of `MxSmkPresenter`. I'd probably relocate `MxSmack` together with this function into its own unit `mxsmack.cpp/.h`, but this can be done in a separate PR.

PR #252, opened 2023-10-28
implement a few mxstream* methods

MishaProductions:
- implemented MxDiskStreamProvider::SetResourceToGet (couldn't get to match due to weird MxString allocations)
 - implemented MxRAMStreamController::Open (couldn't figure out how to fully match it)

For whatever reason, MxDiskStreamProvider::SetResourceToGet calls MxTickleThread::StartWithTarget (when MxDiskStreamProviderThread is actually an MxThread)


foxtacles:
* I've relocated the `m_target` member to `MxThread`. Based on the findings that seemed reasonable.
* Matched `MxDiskStreamProvider::SetResourceToGet` to ~96% - rest might be compiler randomness or one of the function signatures is not correct yet
* Matched `MxRAMStreamController::Open` to 100%

Also, I've noticed you just added a new commit adding another function. In the future, please add additional contributions in a new / separate PR, so that the scope of the PR doesn't change mid-review. Thank you!

PR #254, opened 2023-10-29
MxSmkPresenter vtables

MishaProductions:
I couldn't figure out how to fully match MxSmkPresenter::VTable0x88 (or what it even does)

### DIFF
--- a/3rdparty/smk/smk.h
+++ b/3rdparty/smk/smk.h
@@ -40,7 +40,7 @@ struct Smack {
 	unsigned long m_trackType[7];
 	unsigned long m_extra;
 	unsigned long m_newPalette;
-	unsigned int m_palette[193];
+	unsigned char m_palette[772];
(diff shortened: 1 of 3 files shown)


3rdparty/smk/smk.h
```diff
@@ -40,7 +40,7 @@ struct Smack {
 	unsigned long m_trackType[7];
 	unsigned long m_extra;
 	unsigned long m_newPalette;
-	unsigned char m_palette[772];
+	unsigned int m_palette[193];
```
foxtacles, line 43: Is there a reason why this needs changing? Since it's a 3rd party source it probably shouldn't be touched. @Ramen2X

Ramen2X: Yeah, I agree. Unless there's functionality that actually depends on the member being set this way/and or makes more sense like this, I would leave it how it was. This is how official implementations define this member.

PR #255, opened 2023-10-29
more MxDSBuffer functions

MishaProductions:
- turned mode into an enum based on MxDSBuffer::AllocateBuffer
TODO: match MxDSBuffer::AllocateBuffer

### DIFF
--- a/LEGO1/mxdsbuffer.cpp
+++ b/LEGO1/mxdsbuffer.cpp
@@ -29,7 +29,7 @@ MxDSBuffer::~MxDSBuffer()
 			// TODO
 		}
 		else if (m_mode == MxDSBufferType_Allocate || m_mode == MxDSBufferType_Unknown) {
-			free(m_pBuffer);
+			delete m_pBuffer;
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


LEGO1/mxdsbuffer.cpp
```diff
@@ -14,24 +17,112 @@ MxDSBuffer::MxDSBuffer()
...
 // OFFSET: LEGO1 0x100c6530
 MxDSBuffer::~MxDSBuffer()
 {
-	// TODO
+	if (m_pBuffer != NULL) {
+		if (m_mode == MxDSBufferType_Chunk) {
+			// TODO
+		}
+		else if (m_mode == MxDSBufferType_Allocate || m_mode == MxDSBufferType_Unknown) {
+			free(m_pBuffer);
```
DimaLeon2000, line 32: Should be `delete m_pBuffer`.

foxtacles:
`MxStreamer` should return references to the subclasses instead of copies. I've also fixed the `delete` call (should be `delete[]`). `AllocateBuffer` still needs work obviously but we can merge it.

PR #256, opened 2023-10-31
[Proposal] Add virtual table verification to reccmp

itsmattkc:
This is a proposed addition to reccmp that would verify the position of each vtable function using the PDB. It also breaks up some reccmp functionality into modules for easier maintenance.

The reason I'm leaving it as a draft for now is because we will need to agree on a format for the vtable comments that the script will read.

In this draft, I've copied the `// OFFSET` formatting by having `// VTABLE` in a line above vtable functions:

```
// VTABLE 0x8
virtual MxResult Tickle();
```

There's an example of this in `mxcore.h`.

To do this the `// VTABLE` comments above each class defining the address of the original vtable has had to be renamed to prevent confusion. I've chosen to rename them to `// VTABLEADDR`, but am open to other suggestions.

Formatting it this way makes things consistent and explicit, but might be a little cumbersome to write and might force each class to take up a lot of lines. Alternatively, we could make the script check for a comment at the _end_ of the line, which more closely matches our current formatting, e.g.

```
virtual MxResult Tickle(); // vtable 0x8
```

This has a benefit of being easier to write and less obtrusive, and should be equally doable, though it also may be easier to miss as a reader.

This addition also has the capability to verify the offsets of members or the size of each class (though we already have the latter as a compile step). That may be useful too, and I figure whatever formatting we use for the vtable functions we can re-use for the member offsets if we decide to add that too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ ISLE.EXE
 LEGO1.DLL
 build/
 *.swp
+__pycache__/
--- a/LEGO1/act1state.h
+++ b/LEGO1/act1state.h
@@ -3,7 +3,7 @@
 
 #include "legostate.h"
 
-// VTABLE 0x100d7028
+// VTABLEADDR 0x100d7028
 // SIZE 0x26c
 class Act1State : public LegoState {
 public:
--- a/LEGO1/act2brick.h
+++ b/LEGO1/act2brick.h
@@ -3,7 +3,7 @@
 
 #include "legopathactor.h"
 
-// VTABLE 0x100d9b60
+// VTABLEADDR 0x100d9b60
 // SIZE 0x194
 class Act2Brick : public LegoPathActor {
 public:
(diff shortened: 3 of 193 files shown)


foxtacles:
Nice! Personally I think the "relative offset notation" using lower case would be preferable:

`virtual MxResult Tickle(); // vtable 0x8`

As you said yourself, this formatting enhances readability, can be reused for other offset comments such as the ones for members, and also allow us to leave the current upper case `VTABLE` comments as they are in tact.

What do you think about the efforts in https://github.com/isledecomp/isle/pull/228? Considering we keep extending our "decomp declaration language" I think it makes sense to decouple the parsing/detection of the comments from `reccmp` so that other checks (like the one in the PR) can make use of that information as well.

itsmattkc:
My understanding is this has been or will be covered in upcoming PRs, so I will close this

PR #258, opened 2023-10-31
Add missing MxSoundManager functions

foxtacles:
This adds two missing non-virtual functions to `MxSoundManager` - it should be complete now (100% matches)

`FUN_100aecf0` was a surprise, since it accesses the `g_mxcoreCount` global which apparently isn't just a counter, but a structure with some extra data in it. So far I wasn't able to make sense of the data or its purpose, so I've just copied it verbatim from the binary.


Ramen2X (approved):
Very interesting. I wonder what the significance of the data is... LGTM

PR #259, opened 2023-11-01
LegoGameState destructor and related

disinvite:
Implements `~LegoGameState()`, one of the public symbols, along with these related functions:

- `LegoGameState::SetROIHandlerFunction`
- `LegoROI::SetSomeHandlerFunction` (called by above)
- `ROIHandlerFunction` (from LegoGameState. set as the handler by the above two)
- `LegoROI::CallTheHandlerFunction` (calls the above)
- `LegoROI::ColorAliasLookup` (called by above)

I'm sure we could come up with better names for these functions. Good match for all, most are 100%. There is an array of structs with color aliases (as I'm calling them) that map to MxS32 for RGB values and a fourth MxS32 that is always zero. The aliases are strings like "lego black", some of which already appear in `LegoGameState`.

The main point of interest is `ColorAliasLookup`. We multiply the RGB values by a constant `double` variable and I'm not sure what that's doing. The array `g_roiColorAliases` has 22 entries, and the loop that checks for a match on the string is doing the range check against the address of the following variable `g_roiConfig` rather than a fixed array size. The match percentage is good here but I doubt the devs did it this way. What am I missing? This does seem to be a do/while because we don't check the condition before starting like we do in a for-loop.

Bonus: used MxTypes for some values in `LegoROI` and reordered the functions.

### DIFF
--- a/LEGO1/legoroi.cpp
+++ b/LEGO1/legoroi.cpp
@@ -1,16 +1,90 @@
 #include "legoroi.h"
 
-// 0x10101368
-int g_roiConfig = 100;
+#include <string.h>
 
-// OFFSET: LEGO1 0x100a9e10
-void LegoROI::SetDisplayBB(int p_displayBB)
-{
-	// Intentionally empty function
-}
+// SIZE 0x14
+typedef struct {
+	const char* name;
+	MxS32 red;
+	MxS32 green;
+	MxS32 blue;
+	MxS32 unk10;
+} roi_color_alias;
+
+// 0x100dbe28
+double g_unk_roiConstant = 0.00392156862745098;
+
+// 0x101011b0
+roi_color_alias g_roiColorAliases[22] = {
+	{"lego black", 0x21, 0x21, 0x21, 0},       {"lego black f", 0x21, 0x21, 0x21, 0},
+	{"lego black flat", 0x21, 0x21, 0x21, 0},  {"lego blue", 0x00, 0x54, 0x8c, 0},
+	{"lego blue flat", 0x00, 0x54, 0x8c, 0},   {"lego brown", 0x4a, 0x23, 0x1a, 0},
+	{"lego brown flt", 0x4a, 0x23, 0x1a, 0},   {"lego brown flat", 0x4a, 0x23, 0x1a, 0},
+	{"lego drk grey", 0x40, 0x40, 0x40, 0},    {"lego drk grey flt", 0x40, 0x40, 0x40, 0},
+	{"lego dk grey flt", 0x40, 0x40, 0x40, 0}, {"lego green", 0x00, 0x78, 0x2d, 0},
+	{"lego green flat", 0x00, 0x78, 0x2d, 0},  {"lego lt grey", 0x82, 0x82, 0x82, 0},
+	{"lego lt grey flt", 0x82, 0x82, 0x82, 0}, {"lego lt grey fla", 0x82, 0x82, 0x82, 0},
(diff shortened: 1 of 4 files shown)


LEGO1/legoroi.cpp
```diff
@@ -1,16 +1,90 @@
...
+typedef struct {
+	const char* name;
+	MxS32 red;
+	MxS32 green;
+	MxS32 blue;
+	MxS32 unk10;
+} roi_color_alias;
+
+// 0x100dbe28
+double g_unk_roiConstant = 0.00392156862745098;
```
foxtacles, line 15: This seems to be `1/255`. I'm assuming it's used to normalize the RGB color values to a value in the range 0-1

LEGO1/legoroi.cpp
```diff
@@ -1,16 +1,90 @@
...
+#include <string.h>
+
+// SIZE 0x14
+typedef struct {
+	const char* name;
+	MxS32 red;
+	MxS32 green;
+	MxS32 blue;
+	MxS32 unk10;
+} roi_color_alias;
```
foxtacles, line 12: We should probably name this `RoiColorAlias` (or `ROIColorAlias`)  to stick to our naming conventions

LEGO1/legoroi.cpp
```diff
@@ -1,16 +1,90 @@
...
-int g_roiConfig = 100;
+MxS32 g_roiConfig = 100;
 
-// OFFSET: LEGO1 0x100a9e10
-void LegoROI::SetDisplayBB(int p_displayBB)
-{
-	// Intentionally empty function
-}
+// 0x101013ac
+ROI_Handler g_someHandlerFunction = NULL;
```
foxtacles, line 36: Same here with naming, underscore shouldn't be there

LEGO1/legoroi.cpp
```diff
@@ -1,16 +1,90 @@
...
 
 // OFFSET: LEGO1 0x100a81c0
-void LegoROI::configureLegoROI(int p_roi)
+void LegoROI::configureLegoROI(MxS32 p_roi)
 {
 	g_roiConfig = p_roi;
 }
+
+// OFFSET: LEGO1 0x100a9bf0
+MxBool LegoROI::CallTheHandlerFunction(char* p_param, float& p_red, float& p_green, float& p_blue, float& p_other)
```
foxtacles, line 45: I think we should start using `MxFloat`

LEGO1/legoroi.cpp
```diff
@@ -1,16 +1,90 @@
...
+	if (p_param == NULL)
+		return FALSE;
+
+	if (g_someHandlerFunction) {
+		char buf[32];
+		if (g_someHandlerFunction(p_param, buf, 32))
+			p_param = buf;
+	}
+
+	return LegoROI::ColorAliasLookup(p_param, p_red, p_green, p_blue, p_other);
```
foxtacles, line 57: prefixing the `LegoROI` class shouldn't be necessary

LEGO1/legoroi.cpp
```diff
@@ -1,16 +1,90 @@
 #include "legoroi.h"
 
+#include <string.h>
+
+// SIZE 0x14
+typedef struct {
+	const char* name;
```
foxtacles, line 7: Members should be prefixed with `m_`, even in structs like this

foxtacles (approved):
Looks good! The loop in the alias lookup is odd indeed, but I can't think of any way this can be written differently considering the source assembly right now. Merging!

disinvite:
Thanks! (And thank you for spotting the 1/255)

PR #263, opened 2023-11-04
Finish MxBackgroundAudioManager

MishaProductions:
- Renamed MxParamType Paint to c_notificationStartAction
 - implemented a few methods in MxBackgroundAudioManager
 
TODO:
  - [x] Implement 0x10015910 (appears to start playing music)
  - [x] Implement MxBackgroundAudioManager::Tickle

### DIFF
--- a/LEGO1/legoomni.cpp
+++ b/LEGO1/legoomni.cpp
@@ -212,10 +212,27 @@ GifManager* GetGifManager()
 	return LegoOmni::GetInstance()->GetGifManager();
 }
 
+// OFFSET: LEGO1 0x100158e0
+MxDSAction& GetCurrentAction()
+{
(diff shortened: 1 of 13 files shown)


foxtacles:
* Changed `MxResult Notify` to `MxLong Notify`, since a value of `1` is returned from the `Notify` function which isn't part of the `MxResult` constant set.
* Matched `MxBackgroundAudioManager::StartAction`
* Used tickle state enum in `MxBackgroundAudioManager::Tickle`
* Matched `MxBackgroundAudioManager::FUN_1007ee70` - please note that there is a naming conflict with https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-deleteobject and you have to be careful with the overload resolution.

PR #264, opened 2023-11-05
implement register/unregister scripts

MishaProductions:


foxtacles:
Looks good!

* Matched `RegisterScripts` and `UnregisterScripts` (`testScript` was duplicated)

MishaProductions:
Good catch!

PR #268, opened 2023-11-05
MxQueue and parts of LegoInputManager

disinvite:
Implements the `MxQueue` datatype that we need for `LegoInputManager`. It is a subclass of `MxList`, adding only the Enqueue and Dequeue methods. I had to open up the scope from private to protected in `MxList` to make this work. Only Dequeue is implemented right now because I don't think Enqueue is ever called. If you go off the 96 source (and I did) then it seems the `QUEUE_EVENTS` macro is not defined, and so Enqueue is not called in `QueueEvent()`.

Sidebar while on the topic of these complex datatypes: we have a few now, and all seem to require superclasses like `MxHashTableParent` that don't make much sense. I think what's happening here is that all collection datatypes have a common ancestor, say: `MxCollection`. With that, you get an `m_count` variable and a `m_customDestructor` method to clean up the template type. Each subclass builds on top of that to add whatever it needs for that datatype. For example: an `MxQueue` is just a linked list with a special API, so it extends `MxList`.

The vtables that you get in the binary are specific to whatever class was provided in the template; they are not generic vtables. I added a bunch of template defs to legoinputmanager.h to illustrate what I mean. I'm going to look into this a little more because it's too big of a change to jam into this pull request, but it should simplify a lot of what we have now.

`LegoEventNotificationParam` got an extra parameter to round out its size. The constructor used in `QueueEvent()` (as in the 96 source, and added here) leaves off this parameter, though. We will also have to address whether the type of `MxNotificationParam::m_type` should be the enum `NotificationId`. Most likely yes, because the exposed symbol for `QueueEvent` uses it. I can change this after review if we feel strongly about doing it now.

### DIFF
--- a/LEGO1/legoinputmanager.cpp
+++ b/LEGO1/legoinputmanager.cpp
@@ -9,10 +9,10 @@ DECOMP_SIZE_ASSERT(LegoInputManager, 0x338);
 // OFFSET: LEGO1 0x1005b790
 LegoInputManager::LegoInputManager()
 {
-	m_eventQueue = NULL;
+	m_unk0x5c = NULL;
 	m_world = NULL;
 	m_camera = NULL;
-	m_unk0x68 = NULL;
+	m_eventQueue = NULL;
 	m_unk0x80 = 0;
 	m_timer = 0;
 	m_unk0x6c = 0;
@@ -36,7 +36,7 @@ LegoInputManager::LegoInputManager()
 // OFFSET: LEGO1 0x1005b8b0 STUB
 MxResult LegoInputManager::Tickle()
 {
-	// TODO
+	ProcessEvents();
 	return SUCCESS;
 }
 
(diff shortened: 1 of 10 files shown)


LEGO1/legoinputmanager.cpp
```diff
@@ -46,19 +46,27 @@ LegoInputManager::~LegoInputManager()
...
+}
+
 // OFFSET: LEGO1 0x1005bfe0
 void LegoInputManager::Destroy()
 {
 	ReleaseDX();
 
+	if (m_unk0x5c)
+		delete m_unk0x5c;
+	m_unk0x5c = NULL;
```
ItzSwirlz, line 64: Did the developers actually do this? I don't think its necessary to set something to NULL after deletion

disinvite: Yeah, it's in there:
```assembly
    1005bfe8 8b 4e 5c   MOV     ECX,dword ptr [ESI + 0x5c]
    1005bfeb 85 c9      TEST    ECX,ECX
    1005bfed 74 06      JZ      LAB_1005bff5
    1005bfef 6a 01      PUSH    0x1
    1005bff1 8b 01      MOV     EAX,dword ptr [param_1]
    1005bff3 ff 10      CALL    dword ptr [EAX]    ; `scalar deleting destructor'
                    LAB_1005bff5                      XREF[1]: 1005bfed(j)  
```

`delete` or `free()` doesn't set the pointer to NULL, so you'll segfault if you dereference it again. Setting to NULL allows the code to ask "is there something here?" and if the pointer is NULL then the answer is "no."

foxtacles:
> We will also have to address whether the type of `MxNotificationParam::m_type` should be the enum `NotificationId`. Most likely yes, because the exposed symbol for `QueueEvent` uses it. I can change this after review if we feel strongly about doing it now.

Yeah, I think this would be a great opportunity to make that change. The easiest way to do that would probably be to just rename `MxParamType` to `NotificationId` and delete the old `NotificationId` enum. Thank you!

PR #273, opened 2023-11-07
Improve python tools

tntexplosivesltd:
* Use `with` statement for file access
* Use f-strings instead of modulo string formatting
* Single quotes in most places

Unfortunately I am not able to test this, I don't have a copy of Lego Island and I can't get any output asm files to test with. I am also using Linux.

Is there any easy way for me to test this? I plan on improving these python tools

### DIFF
--- a/tools/reccmp/reccmp.py
+++ b/tools/reccmp/reccmp.py
@@ -50,19 +50,19 @@ plain = args.no_color
 
 original = args.original
 if not os.path.isfile(original):
-  parser.error('Original binary does not exist')
+  parser.error(f'Original binary {original} does not exist')
 
 recomp = args.recompiled
 if not os.path.isfile(recomp):
-  parser.error('Recompiled binary does not exist')
+  parser.error(f'Recompiled binary {recomp} does not exist')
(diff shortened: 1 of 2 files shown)


foxtacles:
> Is there any easy way for me to test this? I plan on improving these python tools

Hi! Try to replicate the workflow specified in the build action: https://github.com/isledecomp/isle/blob/master/.github/workflows/build.yml This should also work on Linux. A copy of LEGO Island is not required (although a few files from it, but you can obtain them the same way the build script gets them).

tntexplosivesltd:
Okay I managed to get everything building and the scripts run now.
Had to get it working under Windows

foxtacles:
Looking at the output of `reccmp`, I think `can_resolve_register_differences` appears to be broken since it no longer recognizes the "100%*" matches.

Also rounding the final accuracy metric to show 2 digits would look better (as it was before):

`Total effective accuracy 56.33056831607717%`

tntexplosivesltd:
Ah oops found my mistake 😄 have fixed it now

PR #274, opened 2023-11-08
Bootstrap MxMediaPresenter

foxtacles:
* Adds the members to `MxMediaPresenter`. What we have here is a list of stream chunks, a current chunk, and a cursor into the stream chunk list, as well as an instance of `MxDSSubscriber`.
* Implements `Destroy` and `StartAction` and the list w/ annotations, 100% matches

For the list, I've re-used the same implementation pattern as before. @disinvite - the refactoring of the list stuff you mentioned in https://github.com/isledecomp/isle/pull/268 would be great indeed going forward since it feels like more boilerplate than it should be.


DimaLeon2000:
Looks okay to me.

PR #276, opened 2023-11-09
Implement/match MxMediaPresenter::EndAction

foxtacles:
* Implements/matches `EndAction` - 100% match
* The `unkPresenter` in `MxPresenter` is most likely a `MxCompositePresenter` considering the way it is used in `EndAction`, so I retyped the member. I've also renamend the `Flag_Parsed` flag to `Flag_World` since that seems to more closely resemble its intended meaning


DimaLeon2000:
LGTM

PR #278, opened 2023-11-10
Prepare MxMIDIPresenter vtable and refactoring

foxtacles:
Adds the virtual table stubs for `MxMIDIPresenter` in preparation for some implementation work. Also some minor refactoring in related classes.

### DIFF
--- a/LEGO1/mxaudiopresenter.h
+++ b/LEGO1/mxaudiopresenter.h
@@ -5,9 +5,10 @@
 #include "mxmediapresenter.h"
 
 // VTABLE 0x100d4c70
+// SIZE 0x54
 class MxAudioPresenter : public MxMediaPresenter {
 public:
-	MxAudioPresenter() { m_unk50 = 100; }
+	MxAudioPresenter() { m_volume = 100; }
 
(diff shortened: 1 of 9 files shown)


LEGO1/mxaudiopresenter.h
```diff
@@ -5,9 +5,10 @@
 #include "mxmediapresenter.h"
 
 // VTABLE 0x100d4c70
+// SIZE 0x54
 class MxAudioPresenter : public MxMediaPresenter {
 public:
-	MxAudioPresenter() { m_unk50 = 100; }
+	MxAudioPresenter() { m_volume = 100; }
```
ItzSwirlz, line 11: do we know where this constructor is or is it just a guess?

foxtacles: It's where generally any other inlined function are as well, in the function bodies that call them. In this case, examples include `LegoLoadCacheSoundPresenter`, `MxMusicPresenter` and `MxMIDIPresenter`, all of which match 100%.

itsmattkc:
Looks good to me, thanks!

PR #283, opened 2023-11-12
Refactor MxHashTable/MxVariableTable

disinvite:
Part 1 of the promised refactor surrounding collection classes. This introduces `MxCollection` as the superclass for `MxHashTable`.

The goal is to have `MxList` extend this class as well, but I ended up refactoring a lot of `MxHashTable` and `MxVariableTable` (both added by me in #87) to fix problems and bring the code style in line with with @foxtacles did with `MxList`. So this PR will just focus on `MxHashTable` and I will do the rest in a follow-up PR.

Summary:

- `MxHashTable` does not use pure virtual calls, so these are removed.
- Using `T` instead of `T*` for `MxHashTable`. There may not be another hash table in `LEGO1`, but if there is, it will be better to have a more generic generic.
- Adding a lot of `TEMPLATE` markers for reccmp.
- Using more Mx types.
- `MxHashTableCursor::Compare` returns an MxBool even though it is not used by `MxVariableTable`. This was confirmed by the alpha version `LEGO1D.dll` which does not inline any functions (and avoids a lot of code optimizations).
- A `Compare` helper method in `MxString` was long overdue. This is also used in the alpha.
- I moved `MxHashTable::DeleteAll` into its own function; it is called by the destructor. Curiously, the alpha shows that the destructor does call a method for cleaning up the hash table, but it does not `memset` the pointers to zero.

### DIFF
--- a/LEGO1/mxcollection.h
+++ b/LEGO1/mxcollection.h
@@ -1,0 +1,25 @@
+#ifndef MXCOLLECTION_H
+#define MXCOLLECTION_H
+
+#include "mxcore.h"
+
+template <class T>
+class MxCollection : public MxCore {
+public:
+	MxCollection()
+	{
+		m_count = 0;
+		m_customDestructor = Destroy;
+	}
+
+	virtual ~MxCollection() {}
+
+	static void Destroy(T){};
+	virtual MxBool Compare(T, T) { return 0; }
+
+protected:
+	MxU32 m_count;                 // +0x8
+	void (*m_customDestructor)(T); // +0xc
+};
+
+#endif // MXCOLLECTION_H
--- a/LEGO1/mxcollection.h
+++ b/LEGO1/mxcollection.h
@@ -15,7 +15,7 @@ public:
 	virtual ~MxCollection() {}
 
 	static void Destroy(T){};
-	virtual MxBool Compare(T, T) { return 0; }
+	virtual MxS8 Compare(T, T) { return 0; }
 
 protected:
 	MxU32 m_count;                 // +0x8
--- a/LEGO1/mxhashtable.h
+++ b/LEGO1/mxhashtable.h
@@ -67,8 +67,8 @@ protected:
 	// on the value of m_resizeOption. Hard to say whether this is how the devs
 	// did it, but a simple cast in either direction doesn't match.
 	union {
-		MxU32 m_increaseAmount;			// +0x20
-		double m_increaseFactor;		// +0x20
+		MxU32 m_increaseAmount;  // +0x20
+		double m_increaseFactor; // +0x20
 	};
 };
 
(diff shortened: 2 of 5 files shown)


LEGO1/mxcollection.h
```diff
@@ -0,0 +1,25 @@
...
+	MxCollection()
+	{
+		m_count = 0;
+		m_customDestructor = Destroy;
+	}
+
+	virtual ~MxCollection() {}
+
+	static void Destroy(T){};
+	virtual MxBool Compare(T, T) { return 0; }
```
foxtacles, line 18: Are we sure about the `MxBool` return type? Can it actually be distinguished from a `(unsigned) char` in the assembly? I'm wondering since negative values can be returned from all the Compare implementations we have so far.

LEGO1/mxhashtable.h
```diff
@@ -80,7 +62,7 @@ class MxHashTable : protected HashTableParent<T> {
 	MxHashTableNode<T>** m_slots; // +0x10
 	MxU32 m_numSlots;             // +0x14
 	MxU32 m_autoResizeRatio;
```
foxtacles, line 64: member offset annotation would be great

foxtacles:
Looks great! Just a couple comments

LEGO1/mxcollection.h
disinvite: oh duh. yeah, the one from `MxVariableTable` is just a wrapper around `strcmp`. I'll fix it

LEGO1/mxhashtable.h
```diff
@@ -93,77 +75,92 @@ class MxHashTable : protected HashTableParent<T> {
...
-		if (m_match->m_next)
-			m_match->m_next->m_prev = m_match->m_prev;
+	return m_match != NULL;
+}
 
-		m_table->m_customDestructor(m_match->m_obj);
-		delete m_match;
-		m_table->m_numKeys--;
+template <class T>
+void MxHashTableCursor<T>::DeleteMatch()
```
foxtacles, line 121: I think the whole function body is wrapped in a `if (m_match) {}` test in the alpha build (maybe you can double check). I think this check is optimized away by the compiler because it can deduce that `m_match` is not `NULL` in the one case the function is used in our decomp, so even if we add it the assembly doesn't change. iirc the match percentages of some other functions changed though. Personally I'm on the fence if we should just add it as well to match the alpha debug code, since it's obviously not guaranteed to be the same as the original final source, but I guess we could add a code comment about it at least.

disinvite: Yep, it's there (at `LEGO1D 0x1008d8bf`). I'll put that in. Match percentage seems ok, except for `MxHashTable<MxVariable *>::Resize` but that must be compiler weirdness.

PR #285, opened 2023-11-13
Implement/match MxEventPresenter

foxtacles:
* Implements `CopyData`, `ReadyTickle`, `StartingTickle` and `PutData`
* `m_data` is obviously some sort of structure but I don't have enough information yet on what it is. It can most likely be extracted from `libweaver`
* All 100% except `PutData`, which does something odd with an extra `strlen` that I can't get to match.

### DIFF
--- a/LEGO1/mxeventpresenter.cpp
+++ b/LEGO1/mxeventpresenter.cpp
@@ -1,8 +1,10 @@
 #include "mxeventpresenter.h"
 
 #include "decomp.h"
+#include "mxautolocker.h"
 #include "mxeventmanager.h"
 #include "mxomni.h"
+#include "mxvariabletable.h"
 
 DECOMP_SIZE_ASSERT(MxEventPresenter, 0x54);
 
@@ -21,13 +23,14 @@ MxEventPresenter::~MxEventPresenter()
 // OFFSET: LEGO1 0x100c2da0
 void MxEventPresenter::Init()
 {
-	m_unk50 = NULL;
+	m_data = NULL;
 }
 
 // OFFSET: LEGO1 0x100c2db0
 MxResult MxEventPresenter::AddToManager()
 {
 	MxResult ret = FAILURE;
+
 	if (EventManager()) {
 		ret = SUCCESS;
 		EventManager()->AddPresenter(*this);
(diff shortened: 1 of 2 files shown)


LEGO1/mxeventpresenter.h
```diff
@@ -24,13 +24,17 @@ class MxEventPresenter : public MxMediaPresenter {
...
+	virtual MxResult AddToManager() override;      // vtable+0x34
+	virtual void Destroy() override;               // vtable+0x38
+	virtual undefined4 PutData() override;         // vtable+0x4c
+	virtual void CopyData(MxStreamChunk* p_chunk); // vtable+0x5c
 
 private:
 	void Init();
 
-	undefined4* m_unk50;
+	MxU8* m_data; // 0x50
```
ItzSwirlz, line 37: Question: Were (and I guess, are, since I'm focusing on Java for AP CS), array declarations such as: `MxU8 m_data[];` allowed? This way its specifically being an array

LEGO1/mxeventpresenter.cpp
```diff
@@ -44,10 +47,71 @@ void MxEventPresenter::Destroy()
...
 	Init();
 
 	m_criticalSection.Leave();
 }
+
+// OFFSET: LEGO1 0x100c2e30
+void MxEventPresenter::CopyData(MxStreamChunk* p_chunk)
+{
+	m_data = new MxU8[p_chunk->GetLength()];
+	memcpy(m_data, p_chunk->GetData(), p_chunk->GetLength());
```
ItzSwirlz, line 62: m_data -> m_chunk?

LEGO1/mxeventpresenter.cpp
```diff
@@ -44,10 +47,71 @@ void MxEventPresenter::Destroy()
...
+// OFFSET: LEGO1 0x100c2ef0
+undefined4 MxEventPresenter::PutData()
+{
+	MxAutoLocker lock(&m_criticalSection);
+
+	if (IsEnabled()) {
+		if (m_currentTickleState >= TickleState_Streaming &&
+			(m_currentTickleState <= TickleState_Repeating || m_currentTickleState == TickleState_Done)) {
+			if (m_currentChunk && m_currentChunk->GetLength()) {
+				if (m_data[12] == 2) {
```
ItzSwirlz, line 99: Note to self, not a problem: What value does this hardcoded 2 mean? Is this the start of a dictionary value?

LEGO1/mxeventpresenter.h
foxtacles: That is not valid C++. You can declare an array with a static size but that's about it. Java has different fundamentals and while some of the syntax may be similar, it's not applicable here.

LEGO1/mxeventpresenter.cpp
foxtacles: `m_data` is probably a custom structure, but there isn't sufficient information right now to figure out what it is exactly. Until then I think that `m_data` is an appropriate name since for now it's based on a verbatim copy of `p_chunk->m_data`.

ItzSwirlz: Might be some sort of dictionary or Hashmap

foxtacles: This data comes from the `.si` files, and since @itsmattkc has a library documenting the types the structure can probably be extracted from it: https://github.com/itsmattkc/SIEdit/tree/master/lib

I haven't looked into it yet however, but the information is almost certainly there.

foxtacles:
Merging tomorrow if no objections.

itsmattkc:
Yep looks good to me, go ahead and merge

PR #287, opened 2023-11-16
Implement ViewROI and base classes

tahg:
Still not really sure on MxMatrix, but the changes haven't affect the match afaik, and I've already fought with it a lot.
ViewROI and base classes mostly implemented, but there might be some non-virtual functions that remain. Hard to match since not everything makes it to the final pdb/dll currently.


LEGO1/viewroi.h
```diff
@@ -0,0 +1,47 @@
...
+#include "orientableroi.h"
+#include "tgl.h"
+#include "viewlodlist.h"
+
+/*
+	ViewROI objects represent view objects, collections of view objects,
+	etc. Basically, anything which can be placed in a scene and manipilated
+	by the view manager is a ViewROI.
+*/
+class ViewROI : public OrientableROI {
```
foxtacles, line 13: In the 1996 source the base class is `RelativeROI` - has this been changed on purpose?

LEGO1/orientableroi.h
```diff
@@ -0,0 +1,52 @@
...
+
+protected:
+	// vtable + 0x14
+	virtual void VTable0x14() { VTable0x1c(); }
+	virtual void UpdateWorldBoundingVolumes() = 0;
+
+public:
+	virtual void OrientableROI::VTable0x1c();
+	// vtable + 0x20
+	virtual void OrientableROI::SetLocalTransform(const MxMatrix& p_transform);
```
foxtacles, line 37: The class name should not be prefixed

foxtacles:
Since this is mostly stuff from libraries that are separate from the rest of the codebase, we should be careful not to commingle it. The 1996 source gives us the folder/library names many of these files belong to. We already have the `realtime` one (currently in `LEGO1/realtime`) so I think we should do the same for `viewmanager`, `tgl` etc.

All the files in these folders should probably not have any dependencies to the Omni or Lego code (meaning, no `Mx*` types used).

LEGO1/viewroi.h
tahg: Somewhere along the line one of OrientableROI or RelativeROI was removed, or they got merged.  What still exists more closely matches OrientableROI, so I went with that name.

LEGO1/orientableroi.h
tahg: Oops, I'll fix this. Things always seem to get shuffled between the cpp and h while I work, and copy/paste errors like this creep in.

tahg:
Right now the build script does not pull in arbitrary extra directories on the path, and I don't know how to fix that, or add more.  This I think has the side effect of having to specify eg, realtime/realtime.h, for it to find the file.

itsmattkc:
If you want to include `realtime.h` without specifying the `realtime/` path, the best way would be to add the following to CMakeLists.txt:

```
target_include_directories(lego1 PRIVATE
    "${CMAKE_SOURCE_DIR}/lego1/realtime"
    "${CMAKE_SOURCE_DIR}/lego1/viewmanager"
    "${CMAKE_SOURCE_DIR}/lego1/tgl"
)
```
(I've added `viewmanager` and `tgl` for reference to show multiple paths, it can be kept to just `realtime` for now)

LEGO1/tglvector.h
```diff
@@ -0,0 +1,277 @@
...
+#include "math.h" // ??? sin() in  RotateAroundY()
+
+#include <stddef.h> // offsetof()
+
+namespace Tgl
+{
+
+namespace Constant
+{
+const float Pi = 3.14159265358979323846;
```
ItzSwirlz, line 13: does MSVC's own library have a pi constant? especially in math.h

itsmattkc:
Supposedly TGL is part of the public domain, according to the [resume of Melinda Green](https://superliminal.com/daniel/greenres.htm) (a former Mindscape employee):

> Also developed TGL (Thin Graphics Layer) an abstract 3D game API used initially by the Lego product to provide a porting layer on top of Direct3D and other rendering libraries. TGL has been placed in the public domain with the hope of continued development by a consortium of game companies.

I can't find anywhere it's available though. This developer also has a "[free source code](https://superliminal.com/sources/)" page containing libraries like Squat (which is also present in the 96 source), but TGL is not listed there. Perhaps I could reach out and ask where I could find a copy to use in the decomp?

LEGO1/tglvector.h
tahg: Perhaps, the code uses several slightly different versions of Pi, in different places.  And also a few other constants derived from that, like pi/180 and 180/pi for conversion.

LEGO1/lodlist.h
```diff
@@ -0,0 +1,180 @@
...
+#define LODLIST_H
+
+#include "assert.h"
+
+#include <stddef.h> // size_t
+
+class LODObject;
+
+// disable: identifier was truncated to '255' characters in the debug information
+#pragma warning(disable : 4786)
```
madebr, line 11: I don't see this warning with msvc 4.2.
Where and how is it triggered?

tahg: Templated names have the potential to generate this error.  A normal name plus namespaces generally wouldn't get close. At any rate, it was in the reference source file, so I left it in, but we may decided to clean up pragma's at some stage if we find them to be unnecessary.

madebr: Thanks. When I do my next bi-monthly mingw check-up, I'll probably `#ifdef _MSC_VER` this out :)

tahg: Yep, it's one of those strange issues that's apparently unique to MSVC.

LEGO1/tgl/tgl.h
```diff
@@ -0,0 +1,363 @@
...
+// Group
+
+class Group : public Object {
+public:
+#if 0
+    virtual Result  SetPosition(const double[3]) = 0;
+    virtual Result  SetOrientation(const double direction[3],
+                                   const double up[3]) = 0;
+#endif
+	virtual Result SetTransformation(const Matrix4&) = 0;
```
foxtacles, line 306: We should probably add a major TODO comment considering we changed the parameter type here. Seems highly unlikely this would suddenly take a `Matrix4` and not a TGL vector like the others. I know this was changed to make the `UpdateWorldData` code match but it's an odd state to be in, so I have a feeling we might be missing something here.

tahg: Yes, it should be a FloatMatrix4, as per the commented out declaration in UpdateWorldData. I can't remember now what the issue I had was with that.  I think it generated extra code.

LEGO1/realtime/roi.h
stravant, line 1: Could you add a comment in here pointing out that ROI = Region of Interest? I don't think it's obvious, even having worked a fair bit on game code I hadn't seen that acronym before and couldn't come up with what that stood for without looking it up.

PR #290, opened 2023-11-18
MxList refactor

disinvite:
Continued from #283. `MxList` template classes now extend from `MxCollection`. We also have `MxPtrList`, which is the same as `MxList` but you can omit the `*` when declaring the template. It is not used consistently as you'll see from the code changed here, but it does conveniently explain the missing link between `MxList` and the most-derived class in a few places.

I left the cursor classes alone for now. Converting the list classes is probably enough for a single PR review.

- Restored the `this` reference in places where I had mistakenly removed it in #283. This was added in #245 to support two-phase compilation. (See also: https://stackoverflow.com/questions/4643074/why-do-i-have-to-access-template-base-class-members-through-the-this-pointer/4643295#4643295)
- We did get a slight reduction in boilerplate code: `MxPtrList` sets the `m_customDestructor` member, which means we don't have to do this again manually in each subclass. There is still the case of many identical `Compare` methods, but the vtables tell us that they do not come from the superclass. A tricky `typedef` could possibly eliminate this redundancy, but the 96 source shows that the devs were happy to just reimplement this wherever it was needed.
- Moved the `// OFFSET: LEGO1 <addr> TEMPLATE` marks to .cpp files where they had been in .h files. This will assist the `checkorder` tool in #228.
- I have been adding `// VTABLE <addr> TEMPLATE` marks for template classes such as `MxList` or `MxCollection`. My thought is that @itsmattkc's proposed enhancement to `reccmp` in #256 may be able to use these, and also that it provides a reference for us to see in the code. If these aren't needed or should be moved, let me know what you think.
- One new collection class: `LegoPathControllerList` used by `LegoWorld`. I added this and two `MxPresenterList` members to `LegoWorld`. It looks like the `MxList` of each collection type gets its own constructor, but right now our generated code inlines all the vtable setups. I marked them as `// OFFSET`s for documentation purposes and in the event that it just fixes itself later.

### DIFF
--- a/LEGO1/legopathcontrollerlist.cpp
+++ b/LEGO1/legopathcontrollerlist.cpp
@@ -11,12 +11,6 @@ MxS8 LegoPathControllerList::Compare(LegoPathController* p_a, LegoPathController
 	return p_a == p_b ? 0 : p_a < p_b ? -1 : 1;
 }
 
-// OFFSET: LEGO1 0x1001d3c0
-void LegoPathControllerList::Destroy(LegoPathController* p_controller)
-{
-	delete p_controller;
-}
-
 // OFFSET: LEGO1 0x1001d230 TEMPLATE
 // MxCollection<LegoPathController *>::Compare
 
@@ -31,6 +25,12 @@ void LegoPathControllerList::Destroy(LegoPathController* p_controller)
 
 // OFFSET: LEGO1 0x1001d330 TEMPLATE
 // MxList<LegoPathController *>::~MxList<LegoPathController *>
+
+// OFFSET: LEGO1 0x1001d3c0
+void LegoPathControllerList::Destroy(LegoPathController* p_controller)
+{
+	delete p_controller;
+}
 
 // OFFSET: LEGO1 0x1001d490 TEMPLATE
 // MxCollection<LegoPathController *>::`scalar deleting destructor'
--- a/LEGO1/legopathcontrollerlist.cpp
+++ b/LEGO1/legopathcontrollerlist.cpp
@@ -1,0 +1,42 @@
+#include "legopathcontrollerlist.h"
+
+#include "decomp.h"
+#include "legopathcontroller.h"
+
+DECOMP_SIZE_ASSERT(LegoPathControllerList, 0x18);
+
+// OFFSET: LEGO1 0x1001d210
+MxS8 LegoPathControllerList::Compare(LegoPathController* p_a, LegoPathController* p_b)
+{
+	return p_a == p_b ? 0 : p_a < p_b ? -1 : 1;
+}
+
+// OFFSET: LEGO1 0x1001d3c0
+void LegoPathControllerList::Destroy(LegoPathController* p_controller)
+{
+	delete p_controller;
+}
+
+// OFFSET: LEGO1 0x1001d230 TEMPLATE
+// MxCollection<LegoPathController *>::Compare
+
+// OFFSET: LEGO1 0x1001d240 TEMPLATE
+// MxList<LegoPathController *>::MxList<LegoPathController *>
+
+// OFFSET: LEGO1 0x1001d2d0 TEMPLATE
+// MxCollection<LegoPathController *>::~MxCollection<LegoPathController *>
+
+// OFFSET: LEGO1 0x1001d320 TEMPLATE
+// MxCollection<LegoPathController *>::Destroy
+
+// OFFSET: LEGO1 0x1001d330 TEMPLATE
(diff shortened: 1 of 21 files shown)


LEGO1/legopathcontrollerlist.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+DECOMP_SIZE_ASSERT(LegoPathControllerList, 0x18);
+
+// OFFSET: LEGO1 0x1001d210
+MxS8 LegoPathControllerList::Compare(LegoPathController* p_a, LegoPathController* p_b)
+{
+	return p_a == p_b ? 0 : p_a < p_b ? -1 : 1;
+}
+
+// OFFSET: LEGO1 0x1001d3c0
```
foxtacles, line 14: Looks like it's not sorted into the right place, should be soon covered by checkorder though.

LEGO1/legoworld.h
```diff
@@ -39,9 +41,12 @@ class LegoWorld : public LegoEntity {
 	void EndAction(MxPresenter* p_presenter);
 
 protected:
-	undefined m_unk68[0x30];
-	LegoCameraController* m_camera;
-	undefined m_unk9c[0x5a];
+	LegoPathControllerList m_list0x68; // +0x68
```
foxtacles, line 44: We haven't formalized the rule yet but so far we have not put a plus sign in front of the member offset

foxtacles:
Looks good! I'm seeing a few possible regressions, did you look into these?

```
-  MxDSMultiAction::CopyFrom (0x100ca0d0) is 100.00% similar to the original
+  MxDSMultiAction::CopyFrom (0x100ca0d0) is 81.25% similar to the original
-  MxAudioManager::InitPresenters (0x100b8e40) is 100.00% similar to the original
+  MxAudioManager::InitPresenters (0x100b8e40) is 64.29% similar to the original
-  MxRegion::MxRegion (0x100c31c0) is 100.00% similar to the original
+  MxRegion::MxRegion (0x100c31c0) is 87.34% similar to the original

disinvite:
I'm gonna check out the alpha to see if there's any clue, but I like your approach. I thought it _was_ overriding the `Destroy` method, but I must've forgot to check the cpp.s version

foxtacles:
> I'm gonna check out the alpha to see if there's any clue, but I like your approach. I thought it _was_ overriding the `Destroy` method, but I must've forgot to check the cpp.s version

The `MxPtrList` class can't possibly be aware of any functions that are overridden in derived classes, so at the scope of the `MxPtrList` constructor it will always set the same `Destroy` stub.

disinvite:
After some digging, I think your suggestion _is_ the way they set the custom destructor, so thanks! I'm not sure why we can't see it in the 96 source, though.

In `LEGO1D`: `MxRegion::MxRegion` should be at 0x1007f7d0, which gives us 0x10081480 for the `MxRegionList` constructor with a vtable at 0x10117dd8.

The constructor for `MxPtrList<MxRegionTopBottom>` is at 0x10081520, and it calls down into a function at 0x100817e0 where (it seems to me) we decide between the default `Destroy()` (0x10081760) and the custom one (0x10081820). That function 0x100817e0 uses an odd instruction `SETNZ` and if you search for that you can find many other similar functions. Assuming all the above is correct, this is probably a good way to identify other `MxPtrList` instances in the alpha.

foxtacles:
> The constructor for `MxPtrList<MxRegionTopBottom>` is at 0x10081520, and it calls down into a function at 0x100817e0 where (it seems to me) we decide between the default `Destroy()` (0x10081760) and the custom one (0x10081820). That function 0x100817e0 uses an odd instruction `SETNZ` and if you search for that you can find many other similar functions. Assuming all the above is correct, this is probably a good way to identify other `MxPtrList` instances in the alpha.

Yeah, that seems to be the crucial bit. Based on the alpha they did use the `p_ownership` flag instead of passing down the `Destroy` function as a parameter.

The main question to solve the problem remains: how does the base class get access/knowledge about a `Destroy` function that - based on the 96 source - is defined in the most derived class? 

We can rule out passing the function as a parameter, as the alpha debug build clearly indicates they are using a boolean (`p_ownership`). It may come down to the same (optimized) assembly match in our decomp but it's probably not what they did.

Another way would be through `T`:

image

However this would require that `Destroy` is defined in `T`, but in the 1996 source we can see the function is defined in the list class (which is also what we did so far). We don't have a way to access the list class, so I'm at a loss how exactly they did.

Either way, I think this is good to merge. We may be able to figure it out completely later.

PR #293, opened 2023-11-19
Bootstrap decomp of Tgl rendering library

stravant:
This PR kicks off work on decompiling the D3D Retained Mode (D3DRM) rendering part of the codebase, in the form of the `Tgl` wrapper library in the existing known source code for the game.

* I did not copy the cursed template-based template method pattern used for implementations in the original Tgl source, though I did copy the class and parameter naming.

* The base Tgl library consists of a set of abstract classes which could have multiple implementations.

* In the shipped game, there is only one implementation, using D3D Retained Mode.

* There are a few significant diversions from the known source code for Tgl. The main one is that Mesh seems to have been split into separate MeshBuilder / MeshInstance classes though we need further decomping to verify that.

* Every class contains at least one non-trivial 100% matching method call to verify things, and only a few methods have been left as STUBs. In total this PR adds 86 methods, 48 of which are 100% matches.


stravant:
Didn't know about TGL. I'll do some adjustments using the TGL code to fill in some Unknowns I had here.

CMakeLists.txt
```diff
@@ -233,6 +239,7 @@ endif()
 if (ISLE_USE_DX5)
   target_include_directories(lego1 PRIVATE "${CMAKE_SOURCE_DIR}/3rdparty/dx5/inc")
   target_link_directories(lego1 PRIVATE "${CMAKE_SOURCE_DIR}/3rdparty/dx5/lib")
+  target_link_libraries(lego1 PRIVATE d3drm)
```
madebr, line 242: Mingw needs this as well, so add it to the next `target_link_libraries` unconditionally.

stravant: Done.

stravant:
Overhauled the changes to be structured like the original Tgl code.

LEGO1/tgl/tgl.h
```diff
@@ -139,98 +123,69 @@ class Renderer : public Object {
...
+	virtual Texture* CreateTexture() = 0;
 	virtual Texture* CreateTexture(
-		int width,
-		int height,
-		int bitsPerTexel,
-		const void* pTexels,
-		int pTexelsArePersistent,
-		int paletteEntryCount,
-		const PaletteEntry* pEntries
+		int p_width,
```
foxtacles, line 133: For the utility libraries, the naming conventions we have in the rest of the code base don't apply:

https://github.com/isledecomp/isle/blob/master/CONTRIBUTING.md#naming-conventions

Instead, we want to retain the original naming scheme as far as it is available (even if inconsistent), so in this case we should leave all the parameter names as they were in the 1996 source. Thank you!

LEGO1/tgl/d3drm/group.cpp
```diff
@@ -0,0 +1,129 @@
+#include "impl.h"
+
+using namespace TglImpl;
+
+// Inlined only
+GroupImpl::~GroupImpl()
+{
+	if (m_data) {
+		free(m_data);
```
foxtacles, line 9: This should be `delete`

LEGO1/tgl/d3drm/mesh.cpp
```diff
@@ -0,0 +1,166 @@
...
+DECOMP_SIZE_ASSERT(D3DRMVERTEX, 0x24);
+
+DECOMP_SIZE_ASSERT(Mesh, 0x4);
+DECOMP_SIZE_ASSERT(MeshImpl, 0x8);
+
+// Inlined only
+MeshImpl::~MeshImpl()
+{
+	if (m_data) {
+		free(m_data);
```
foxtacles, line 14: This should be `delete`

LEGO1/tgl/d3drm/mesh.cpp
```diff
@@ -0,0 +1,166 @@
...
+	newSomething->m_data->groupIndex = index;
+	target->ImplementationData()->SetVertices(index, 0, vcount, vertexBuffer);
+	target->ImplementationData()->SetGroupTexture(index, textureRef);
+	target->ImplementationData()->SetGroupMapping(index, mapping);
+	target->ImplementationData()->SetGroupQuality(index, quality);
+	Result result = ResultVal(target->ImplementationData()->SetGroupColor(index, color));
+
+	// Cleanup
+	if (faceBuffer)
+		free(faceBuffer);
```
foxtacles, line 120: This should be `delete` (below too)

LEGO1/tgl/d3drm/texture.cpp
```diff
@@ -0,0 +1,202 @@
...
+	if (p_palette != NULL) {
+		InitializePalette(p_paletteSize, p_palette);
+	}
+}
+
+// OFFSET: LEGO1 0x100a13b0
+void TglD3DRMIMAGE::Destroy()
+{
+	if (m_texelsAllocatedByClient == 0) {
+		free(m_image.buffer1);
```
foxtacles, line 84: This should be `delete` (below too)

LEGO1/tgl/d3drm/texture.cpp
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// OFFSET: LEGO1 0x100a1550
+Result TglD3DRMIMAGE::InitializePalette(int p_paletteSize, PaletteEntry* p_palette)
+{
+	// This function is a 100% match if the PaletteEntry class is copied
+	// into into the TglD3DRMIMAGE class instead of being a global struct.
+	if (m_image.palette_size != p_paletteSize) {
+		if (m_image.palette != NULL) {
+			free(m_image.palette);
```
foxtacles, line 109: This should be `delete`

LEGO1/viewmanager/viewlodlist.h
```diff
@@ -44,7 +44,7 @@ class ViewLODList : public LODList<ViewLOD> {
 // ??? for now, until we have symbol management
 typedef const char* ROIName;
 struct ROINameComparator {
-	bool operator()(const ROIName& rName1, const ROIName& rName2) const
+	int operator()(const ROIName& rName1, const ROIName& rName2) const
```
foxtacles, line 47: They did use `bool` in the 1996 source so we should leave it as is, or is there a good reason not to? (other than the compiler warning)

foxtacles (commented):
Looks good to me, just a few comments. Thank you!

LEGO1/tgl/d3drm/renderer.cpp
```diff
@@ -0,0 +1,323 @@
+#include "impl.h"
+
+using namespace TglImpl;
+
+// OFFSET: LEGO1 0x100a15e0
+Renderer* CreateRenderer()
```
foxtacles, line 6: This function should be part of the `Tgl` namespace as declared in `tgl.h`?

LEGO1/tgl/d3drm/something.cpp
```diff
@@ -0,0 +1,63 @@
+#include "impl.h"
+
+using namespace TglImpl;
+
+DECOMP_SIZE_ASSERT(Something, 0x4);
+DECOMP_SIZE_ASSERT(SomethingImpl, 0x8);
+
+// Inlined only
+SomethingImpl::~SomethingImpl()
```
foxtacles, line 9: We generally use the term "Unknown" (or `Unk`) in the rest of the code base for unknown stuff, so we should apply it here as well

itsmattkc: In general, naming anything "Something" is pretty objectively bad practice. There are hundreds if not thousands of classes we don't know yet that could all theoretically be called "Something", but how does that help us identify them? If anything, it makes it _harder_ to distinguish them than if it was just offsets.

LEGO1/viewmanager/viewlodlist.h
stravant: It's a compiler error for me, not just a warning, and I couldn't find a good header to include to get bool.

The original header includes some non-standard `std.h` which presumably defined bool. What do you think I should do?

foxtacles: I'm assuming you are not using MSVC 4.20? It only emits a warning about the reserved keyword `bool`, but otherwise compiles fine, presumably treating it as `int`. Or are you using different compiler settings?

Either way, depending on the case, we could add a macro `COMPAT_BOOL` that expands to different things based on the compiler used, similar to `COMPAT_CONST`

LEGO1/tgl/d3drm/group.cpp
stravant: Fixed, should actually have been `Release()`

LEGO1/tgl/d3drm/mesh.cpp
stravant: I don't think that's the case.

A delete call includes a null check, but not a set to null. If this were actually a delete call then the code would either end up with two nested if statements rather than a single one, or the code setting to null outside of the if statement.

MeshImpl is the odd one out, not following the pattern the others have of directly holding a Direct3DRM handle, so it's possible the people who came in later and made that change did use malloc / free for the holder struct instead of new / delete like the initial Tgl implementers did in the leak.

LEGO1/tgl/d3drm/texture.cpp
stravant: This definitely can't be a new/delete allocation because there's no null check on the frees in the destructor.

LEGO1/tgl/d3drm/renderer.cpp
stravant: Yup, missed `Tgl::` on there, forgot how I was doing the namespaces.

LEGO1/tgl/d3drm/mesh.cpp
stravant: This one could be a new/delete, changed.

LEGO1/viewmanager/viewlodlist.h
stravant: Ah, sorry, I didn't notice that it was only a warning since I was dealing with another actual error too at the time.

I don't think we should have warnings we can avoid but that doesn't need to be part of this PR, I'll take out the change.

LEGO1/tgl/d3drm/something.cpp
stravant: There's only one unknown class in the Tgl namespace, the rest of the codebase explicitly uses Tgl classes as `Tgl::` so I don't think there's an issue with using an ambiguous name.

Though I'm not attached to "something", changed it to `Unk` (Using `Unknown` breaks code completion because Microsoft in their infinite wisdom decided to use `Unknown` as an enum item name)

LEGO1/tgl/d3drm/texture.cpp
foxtacles: I might be missing something here, but if I replace both `free`'s in this function with `delete`, I'm getting the exact the same match/disassembly.

Could you elaborate where you see the difference exactly that ascertains that it must be `free`?

LEGO1/tgl/d3drm/texture.cpp
foxtacles: When substituing `free` with `delete`, the match moves from 68% to a full 100%:

image

Again, I might be missing something here, but `delete` just seems correct.

stravant: How did you make this match?? I changed the code to new/delete but it still does not for me.

LEGO1/tgl/d3drm/texture.cpp
stravant: No, you're right, I was just blundering. I got myself thinking that the ifs were the C++ `delete` adding an implicit test, but the `free` also has an embedded null check, so there's no reason for C++ to be doing that.

foxtacles:
> Though I still can't figure out how to test the scalar_deleting_destructors. Is this what you wanted?

Yeah, the way it is done for `MeshImpl` is correct. The same can/should be done for ideally every class with a virtual table there is. Sorry for the lack of documentation and guidance on this stuff, we will improve on that soon.

stravant:
Added checks for all the inline destructors.

LEGO1/tgl/d3drm/camera.cpp
```diff
@@ -5,14 +5,8 @@ using namespace TglImpl;
...
-// Inlined only
-CameraImpl::~CameraImpl()
-{
-	if (m_data) {
-		m_data->Release();
-		m_data = NULL;
-	}
-}
+// OFFSET: LEGO1 0x100a2560 TEMPLATE
+// TglImpl::LightImpl::`scalar deleting destructor'
```
foxtacles, line 9: Meant to be `CameraImpl`?

LEGO1/tgl/d3drm/view.cpp
```diff
@@ -0,0 +1,350 @@
...
+	return reinterpret_cast<ViewportAppData*>(pViewport->GetAppData());
+}
+
+inline IDirect3DRMFrame* ViewportGetLightFrame(IDirect3DRMViewport* pViewport)
+{
+	return ViewportGetData(pViewport)->m_pLightFrame;
+}
+
+// OFFSET: LEGO1 0x100a23a0 TEMPLATE
+// TglImpl::LightImpl::`scalar deleting destructor'
```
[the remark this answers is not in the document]
stravant: Aaaa.... I can't get this right. 🙃

PR #294, opened 2023-11-20
Implement/match remaining MxWavePresenter functions

foxtacles:
Implements the remaining functions in `MxWavePresenter`.

```
+  MxWavePresenter::WriteToSoundBuffer (0x100b1bd0) is 100.00% similar to the original
+  MxWavePresenter::StreamingTickle (0x100b1ea0) is 100.00% similar to the original
+  MxWavePresenter::DoneTickle (0x100b20c0) is 82.35% similar to the original
+  MxWavePresenter::AppendChunk (0x100b2130) is 100.00% similar to the original
+  MxWavePresenter::PutData (0x100b2160) is 100.00% similar to the original
+  MxWavePresenter::EndAction (0x100b2280) is 100.00% similar to the original
+  MxWavePresenter::SetVolume (0x100b2300) is 100.00% similar to the original
+  MxWavePresenter::Enable (0x100b2360) is 100.00% similar to the original
+  MxWavePresenter::ParseExtra (0x100b23a0) is 86.32% similar to the original
+  MxWavePresenter::VTable0x64 (0x100b2440) is 100.00% similar to the original
+  MxWavePresenter::VTable0x68 (0x100b2470) is 100.00% similar to the original
```

### DIFF
--- a/LEGO1/mxwavepresenter.cpp
+++ b/LEGO1/mxwavepresenter.cpp
@@ -8,8 +8,6 @@
 #include "mxomni.h"
 #include "mxsoundmanager.h"
 
-#include <limits.h>
-
 DECOMP_SIZE_ASSERT(MxWavePresenter, 0x6c);
 DECOMP_SIZE_ASSERT(MxWavePresenter::WaveFormat, 0x1c);
 
--- a/LEGO1/mxwavepresenter.cpp
+++ b/LEGO1/mxwavepresenter.cpp
@@ -37,7 +37,7 @@ void MxWavePresenter::Init()
 	m_waveFormat = NULL;
 	m_dsBuffer = NULL;
 	m_length = 0;
-	m_bytes = 0;
+	m_lockSize = 0;
 	m_unk64 = 0;
 	m_playing = FALSE;
 	m_unk66 = FALSE;
@@ -88,7 +88,7 @@ MxBool MxWavePresenter::FUN_100b1ba0()
 }
 
 // OFFSET: LEGO1 0x100b1bd0
-void MxWavePresenter::FUN_100b1bd0(void* p_audioPtr, MxU32 p_length)
+void MxWavePresenter::WriteToSoundBuffer(void* p_audioPtr, MxU32 p_length)
 {
 	DWORD dwStatus;
 	LPVOID pvAudioPtr1;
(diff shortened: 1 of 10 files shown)


LEGO1/mxwavepresenter.cpp
```diff
@@ -84,11 +87,46 @@ MxBool MxWavePresenter::FUN_100b1ba0()
...
+	LPVOID pvAudioPtr1;
+	DWORD dwOffset;
+	LPVOID pvAudioPtr2;
+	DWORD dwAudioBytes1;
+	DWORD dwAudioBytes2;
+
+	dwOffset = m_length * m_unk64;
+	m_dsBuffer->GetStatus(&dwStatus);
+
+	if (dwStatus == DSBSTATUS_BUFFERLOST) {
```
ItzSwirlz, line 103: Maybe this is a switch statement?

LEGO1/mxwavepresenter.cpp
```diff
@@ -159,63 +197,182 @@ void MxWavePresenter::StartingTickle()
...
+			}
+			break;
+		case TickleState_Repeating:
+			if (m_unk65)
+				break;
+
+			m_dsBuffer->SetCurrentPosition(0);
+
+			if (m_dsBuffer->Play(0, 0, m_action->GetLoopCount() > 1) == DS_OK)
+				m_unk65 = TRUE;
```
ItzSwirlz, line 278: m_isPlaying, my only concern is in the MxWavePresenter::Enable function where it sets it to false

LEGO1/mxwavepresenter.cpp
foxtacles: Not sure what you mean, there is no switching on multiple values here.

LEGO1/mxwavepresenter.cpp
```diff
@@ -136,7 +174,7 @@ void MxWavePresenter::StartingTickle()
 		desc.dwSize = sizeof(desc);
 
 		if (m_unk66)
```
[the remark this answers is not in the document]
foxtacles: There should be at least one instance where this variable is written to before I'd name it. I know it's likely related to `MxOmni::IsSound3D` but unless there is code to prove it, it should remain "unknown".

LEGO1/mxwavepresenter.cpp
foxtacles: Seems reasonable. I believe the reason it's set to false there is because it's meant to "reset", and it will be set to `TRUE` once `Play` has been called again.

foxtacles:
Figured out that the missing piece of `FUN_100b1bd0` was to pad the buffer with "silence data", which according to the wave format is `0x7F` when using 8-bit and `0` when 16-bit sound. Also renamed `m_unk67` to `m_silenceData`. Match is now at 100%.

LEGO1/mxwavepresenter.cpp
```diff
@@ -159,63 +194,182 @@ void MxWavePresenter::StartingTickle()
...
+		m_unk68 = TRUE;
+	}
 }
 
-// OFFSET: LEGO1 0x100b2470 STUB
+// OFFSET: LEGO1 0x100b2470
 void MxWavePresenter::VTable0x68()
 {
-	// TODO
+	if (m_unk68) {
```
ItzSwirlz, line 359: Last comment I have is maybe this is just a variable to mark if the buffer is available or not in use

foxtacles: Looks more like a stopped/paused flag to me.

itsmattkc (approved):

PR #296, opened 2023-11-22
Fix reccmp html output for template functions

disinvite:
Template functions look like this using angle brackets: `MxList<MxPresenter *>::~MxList<MxPresenter *>`

In the HTML output of reccmp, we escape the angle bracket characters by converting to `&lt;` and `&gt;`. These are displayed correctly on the page because we create the display elements using the `innerHTML` property, so they become un-escaped. The side effect of this is that the literal data contains the escaped characters, so you can't search for templates unless you escape the text yourself in the search box.

This changes the reccmp output so that we escape for JSON instead, as we are ultimately building up a javascript array and inserting it into template.html using pystache (#292). We can have the browser escape the text automatically by using the `innerText` property instead of `innerHTML`.

The diff output for each function will also contain angle brackets, as in `call <OFFSET 3>` for example. I added more code to template.html that builds this output dynamically, too, also using `innerText` for escaping.

Let me know if the `diffCssClass` function is too codegolf-y with double ternary operators.

### DIFF
--- a/tools/reccmp/reccmp.py
+++ b/tools/reccmp/reccmp.py
@@ -10,7 +10,7 @@ import logging
 import os
 import sys
 import colorama
-import html
+import json
 import re
 from isledecomp.dir import walk_source_dir
 from isledecomp.parser import find_code_blocks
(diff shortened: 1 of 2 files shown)


foxtacles:
Looks good!

PR #298, opened 2023-11-22
Python Linting and Code Formatting

tntexplosivesltd:
This is quite a large PR, and it might raise a few concerns
I have added pylint linting and black code formatting
I have also moved some classes to the utils in the isledecomp python library, and cleaned up the main reccmp file itself

I can split my changes up into their own PRs if need be


foxtacles:
Looks good to me, but since I don't have a lot of experience with Python a second review would be good.

tntexplosivesltd:
Oops, all my commit authors were incorrect.
I also took this opportunity to move Bin and SymInfo into their own files, to help make #256 easier to update

.github/workflows/format.yml
```diff
@@ -19,4 +19,22 @@ jobs:
...
\ No newline at end of file
+          LEGO1/viewmanager/*.cpp LEGO1/viewmanager/*.h
+
+  python-format:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Build isledecomp library
```
madebr, line 30: This step name is incorrect.

madebr: Also, there are 2 `pip install` invocations: here, and at the start of the next step.
Can/should we combine them?

.pylintrc
```diff
@@ -0,0 +1,634 @@
...
+
+# Regexp for a line that is allowed to be longer than the limit.
+ignore-long-lines=^\s*(# )?<?https?://\S+>?$
+
+# Number of spaces of indent required inside a hanging or continued line.
+indent-after-paren=2
+
+# String used as indentation unit. This is usually "    " (4 spaces) or "\t" (1
+# tab).
+indent-string='    '
```
madebr, line 337: I also do my python projects with 4 character indentation, but the maintainer of the python script should decide whether this is fine.

tntexplosivesltd: Who would that be?

The `black` code formatting tool already enforces 4-space indenting (which is pretty standard for Python)
I know using something like `black` can be contentious, but I think it is a good way of having consistent-looking code across the tools.

.github/workflows/format.yml
tntexplosivesltd: Should be all fixed up now, tidied it up across all the .yml files

.pylintrc
foxtacles: The main (original) authors of the scripts would be @itsmattkc and @disinvite

disinvite:
Looks good to me!

itsmattkc:
Seems good to me too, thanks!

PR #304, opened 2023-11-25
MxFlcPresenter vtable70, m_unk64

ItzSwirlz:
vtable70 is basically creating a bitmap; doing this needs m_unk64, which is the main point of interest.

It *appears* that just like MxSmkPresenter, there is a general structure of a "FLIC", at least what appears to be a FLIC header. All that matters currently with m_unk64 is that +0x8 is the width and +0xa is the height.

This PR is just a test for now, we'll have to see whether this is what m_unk64 is or isn't.

### DIFF
--- a/3rdparty/flic/flic.h
+++ b/3rdparty/flic/flic.h
@@ -1,0 +1,36 @@
+#ifndef FLIC_H
+#define FLIC_H
+
+#include <windows.h>
+
+// A basic FLIC header structure from the "EGI" documentation. Source: https://www.compuphase.com/flic.htm#FLICHEADER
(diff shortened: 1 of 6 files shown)


foxtacles:
Looks good to me.

I've applied some fixes and added another function in `MxLoopingFlcPresenter` that makes use of another attribute from the FLIC header structure, `speed`, which is described as "Delay between frames". This value is added to `m_unk68` which makes sense to me, as it would then amount to the sum of the delays after each frame.

There are bigger (more complex) in `MxFlcPresenter` that use the header and work with FLIC data, but so far the structure seems to check out so we can merge it.

PR #307, opened 2023-11-28
STL lists for stream classes

disinvite:
I've been hacking away at `MxStreamController` and related classes, but I thought I would tie off some stuff related to the class members as it probably warrants its own review.

`MxStreamController`, `MxDiskStreamController` and `MxDiskStreamProvider` all use STL lists to hold data. Much like the `MxList` hierarchy, there are four classes in total: two from the STL layer, a wrapper class, and then a final most-derived class. For example:

- `list<MxDSAction *,allocator<MxDSAction *> >::~list<MxDSAction *,allocator<MxDSAction *> >`
- `List<MxDSAction *>::~List<MxDSAction *>`
- `MxStreamList<MxDSAction *>::~MxStreamList<MxDSAction *>`
- `MxStreamListMxDSAction::~MxStreamListMxDSAction`

There are no vtables here, so it (probably) doesn't matter whether the wrapper class (`MxStreamList`) is the same regardless of the type. I've done it that way for simplicity.

The way to tell what's going on is the chain of destructor methods for the member that get tacked onto the end of a constructor. Here is an excerpt from the end of the `MxStreamController` constructor that I've cleaned up a little:

```asm
100c0c84 8b 4d ec         MOV     ECX,dword ptr [EBP + -0x14]
100c0c87 83 c1 54         ADD     ECX,0x54
100c0c8a e9 d1 00 00 00   JMP     FUN_100c0d60

100c0c8f 8b 4d ec         MOV     ECX,dword ptr [EBP + -0x14]
100c0c92 83 c1 54         ADD     ECX,0x54
100c0c95 e9 56 05 00 00   JMP     FUN_100c11f0

100c0c9a 8b 4d ec         MOV     ECX,dword ptr [EBP + -0x14]
100c0c9d 83 c1 54         ADD     ECX,0x54
100c0ca0 e9 5b 04 00 00   JMP     FUN_100c1100

100c0ca5 8b 4d ec         MOV     ECX,dword ptr [EBP + -0x14]
100c0ca8 83 c1 54         ADD     ECX,0x54
100c0cab e9 60 03 00 00   JMP     FUN_100c1010
```

The four jumps here correspond to the four list functions on `MxDSAction*` seen above, in that order. I've added the `TEMPLATE` markers for all these so we can track them with reccmp. The actual datatypes used in the lists are not obvious, so I used the debugger to lookup the vtable for all these.

This also makes it easy to tell where the same class-type was used in multiple places. This list type was used twice in `MxStreamController` and four times in `MxDiskStreamController`. Ghidra (at least on my end) doesn't decompile these blocks automatically, so you have to already know what you're looking for instead of finding all cases where something is called via the xrefs. It might be worth writing a tool that will calculate the jump destination for all these (for each known constructor) and print out a nice report.

I also filled in the rest of the member types for these classes where I could. The constructor for `MxDiskStreamController` _was_ 100% with all my other unfinished code in there, but compiler weirdness knocked it down a peg with that stuff removed.

Speaking of `TEMPLATE` offsets: I had to use `clang-format off` for one very long symbol that exceeds 120 columns. It is:
`list<MxNextActionDataStart *,allocator<MxNextActionDataStart *> >::~list<MxNextActionDataStart *,allocator<MxNextActionDataStart *> >`

There's no way to make that any shorter. The class name for the datatype has a `ClassName()` method overridden from `MxCore`, so the name is correct. We could add `ReflowComments: false` to the clang-format config, but this of course applies globally. Another option is to reshuffle the parser in the `isledecomp` library so it can handle offset-by-name cases that span two lines.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,6 +173,7 @@ add_library(lego1 SHARED
   LEGO1/mxstreamchunklist.cpp
   LEGO1/mxstreamcontroller.cpp
   LEGO1/mxstreamer.cpp
+  LEGO1/mxstreamlist.cpp
   LEGO1/mxstreamprovider.cpp
   LEGO1/mxstring.cpp
   LEGO1/mxstringlist.cpp
--- a/LEGO1/mxcriticalsection.cpp
+++ b/LEGO1/mxcriticalsection.cpp
@@ -1,6 +1,10 @@
 #include "mxcriticalsection.h"
 
+#include "decomp.h"
+
 #include <stdio.h>
+
+DECOMP_SIZE_ASSERT(MxCriticalSection, 0x1c);
 
 // 0x10101e78
 int g_useMutex = 0;
--- a/LEGO1/mxcriticalsection.h
+++ b/LEGO1/mxcriticalsection.h
@@ -3,6 +3,7 @@
 
 #include <windows.h>
 
+// SIZE 0x1c
 class MxCriticalSection {
 public:
 	__declspec(dllexport) MxCriticalSection();
--- a/LEGO1/mxdiskstreamcontroller.cpp
+++ b/LEGO1/mxdiskstreamcontroller.cpp
@@ -5,11 +5,25 @@
 #include "mxomni.h"
 #include "mxticklemanager.h"
 
-// OFFSET: LEGO1 0x100c7120 STUB
+DECOMP_SIZE_ASSERT(MxDiskStreamController, 0xc8);
+
+// OFFSET: LEGO1 0x100c7120
 MxDiskStreamController::MxDiskStreamController()
 {
-	// TODO
+	m_unk8c = 0;
 }
+
+// OFFSET: LEGO1 0x100c7330 TEMPLATE
+// list<MxDSAction *,allocator<MxDSAction *> >::_Buynode
+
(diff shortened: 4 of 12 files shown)


ItzSwirlz:
I don't know if this is related, but I kind of figured out what MxDSSubscriber is doing. Here is an image of a whiteboard I drew up of what's happening in the constructor/the classes:
image

What I said in Matrix was:
"Ok, so I think Class1 kind of is the marker and/or owns 16 bytes of data (DAT_100b5990). Then the vtable is set to Class2 which then that data is sent to a function that runs an operator delete on it. Then, Class3 has some vtable function that compares two uint's. if equal, it returns it, if one is greater than the other, it concatenates 0xFF to the end of the data"

foxtacles:
Great work, looks good to me! Thank you for the explanation too, that is really helpful.

Glad that `MxUnkList` has been figured out, it will make decompiling a fair amount of code easier. Not sure why it never occured to me that it could be STL-based, I remember I spent a few hours trying to build it from scratch but couldn't finish it. I think there is one more instance of `MxUnkList` in `MxCompositePresenter`, wondering what the list element will be in there.

With regards to `clang-format off`, I think the way you did it is the way to go, I wouldn't change any settings or adjustments to the parser.

PR #308, opened 2023-11-28
(Proposal) Adjustments to "decomp" language

foxtacles:
This PR proposes a number of changes to our "decompilation" language and syntax. I believe these changes are necessary to create a coherent foundation that we can more easily extend in the near future.

Details in this comment: https://github.com/isledecomp/isle/pull/308#issuecomment-1831969852

### Summary

The main change would be to introduce a generic syntax in the form of:

`{ENTITY}: {MODULE} {offset}`

that will be consistently applied for other entities we want to check in the future and that have an absolute offset, i.e. virtual tables and probably global variables at some point.


tahg:
For consistency, I suggest we move the names we currently append after {offset} to first class types of {ENTITY}. These would include:

- TEMPLATE: Used for templated functions
  ```
  // TEMPLATE: LEGO1 0x12345678`
  // MyClass<T>::ATemplateFunc
  ```
- STUB: Used for functions which should be tracked but not compared. Ideally I'd like to see these in progress.html, with `STUB` or similar in the `Matching` column.
  ```
  // STUB: LEGO1 0x12345678
  MyClass::AStubbedFunc()
  {
    // Todo
  }
  ```
- FUNCTION: Normal functions. This will be the bulk of our code/report.
  ```
  // FUNCTION: LEGO1 0x12345678`
  MyClass::ANormalFunc()
  {
  }
  ```
- SYNTHETIC: Used for functions which have no source code. Along with STUB, this is mostly for ensuring coverage of the binary.
  ```
  // SYNTHETIC: LEGO1 0x12345678
  // MyClass::`scalar deleting destructor'
  ```
- GLOBAL: Used for static variable definitions.
  ```
  // GLOBAL: LEGO1 0x12345678
  MyClass* g_myClass;
  ```
- VTABLE: Used for a class's virtual function table. We may use this in the future to check virtual function calls, or just for human reference.
  ```
  // VTABLE: LEGO1 0x12345678
  class MyClass {
  };
  ```

foxtacles:
Looks good to me, personally I'd be fine with both approaches. In the end I'd say it comes down to if we can write a parser that can, without excessive complexity or context dependence, match all the cases we are looking for (examples above).

Assuming @tahg approach above, the `isledecomp` library will (eventually) have to find the following entities:

* `FUNCTION`, `SYNTHETIC`, `TEMPLATE` and `STUB` belong to the category of "functions" and are required for `checkorder` and `reccmp`. `SYNTHETIC` and `TEMPLATE` are always followed by a comment and `FUNCTION` and `STUB` are not, however I'm not sure if we need different approaches of reporting for each depending on the needs of `reccmp`, for instance. Ideally all of them can be parsed in the same fashion.
* `VTABLE` are "virtual tables" and will [eventually be required](https://github.com/isledecomp/isle/pull/256) for `reccmp`. Those annotations can be followed by a comment with a class (in case of a templated class) or not. Again, not sure if both cases can be handled with the same code just fine and how `reccmp` will have to distinguish between them (if at all).
* `GLOBAL` are "global variables" and we don't use them for anything for now, but should be eventually collected for additional checks.

@disinvite could you perhaps give us your opinion on the feasibility of this approach? Thank you!

disinvite:
This all looks good! The rigid format for the markers will help a lot with parsing.

I see TEMPLATE and SYNTHETIC as basically the same thing: we want to match the function based on its name rather than the line number where it begins (or, precisely: where the opening curly bracket appears). For example, we have cases where a function is both a SYNTHETIC and a TEMPLATE:

```
// TEMPLATE: LEGO1 0x100c9e30 
// MxCollection<MxDSAction *>::`scalar deleting destructor'
```

So you could say SYNTHETIC is a superset of TEMPLATE.

I don't see a problem matching the VTABLEs, although some distinction between the SYNTHETIC and real cases would make parsing easier. The problem would come later when matching things up with the `cvdump` report. I haven't looked at the output of vtable information yet, but that would decide how the parser should report the "name" of whatever vtable we are looking at. (My thought here is: would it end up being simpler to just always use the synthetic approach for vtables?)

It might also be worth thinking about how we want to use this information. (i.e. We want to match the vtables, but how exactly?) We have the `// vtable+0x8` markers in the header file, so we could extend the parser to grab those and compare. Another way might be to specify (in the VTABLE marker) how many entries are in the class vtable. This would help cover the synthetic cases where we don't have the option to mark things in the class declaration.

I see a similar problem for the global variable case. We can compare the initialized data if there is any, but how many bytes do you check? We could derive the variable size easily by checking for our Mx* types, but it's not so easy for structs.

foxtacles:
I think this is technically ready to merge. I'm going to write a `README` about this stuff, but the gist can be found in this PR. Final review would be great @itsmattkc @tahg

PR #311, opened 2023-12-06
Read section info from PE header

disinvite:
The `Bin` class from the isledecomp/bin module reads the binary files — namely `ISLE.EXE` and `LEGO1.DLL`, in both the original and reconstructed iterations — and provides a mechanism for reading the bytes at a given virtual address. Until now, we have read all the relevant information from the PE header by jumping directly to the known location in the `.text` section.

This enhancement reads all the section information so we can read virtual addresses from anywhere in the exe. It also verifies that the MZ and PE headers are there in the event that you sent the wrong file into the function.

The python logger module used by `reccmp` is sent into the constructor for `Bin`; this change makes it an optional dependency.

A further enhancement is to use the relocation table (the `.reloc` section) to identify locations in the code that refer to a position-independent address. We have been making the educated guess that immediate values higher than the first virtual address are themselves virtual addresses, not values. For example, `LEGO1` code begins at virtual address `0x10001000`, so an instruction like `PUSH 0x1000105c` is assumed to be pushing an offset, not a big number. The relocation table should pinpoint these cases exactly.


foxtacles:
Looks good to me, thank you!

PR #312, opened 2023-12-06
(Proposal) Use alternative C4786 warning suppression

foxtacles:
Unfortunately, the workaround we use to get rid of the warning (by declaring a static class) leads to a junk artifact being inserted into the binaries (probably in several places, but I haven't checked). I've noticed that today when strangely, the recompiled offsets of the functions in `ISLE.EXE` were all wrong:

```
  IsleApp::IsleApp (0x401000 / 0x401020) is 100.00% similar to the original
  IsleApp::~IsleApp (0x4011a0 / 0x4011c0) is 100.00% similar to the original
  IsleApp::Close (0x401260 / 0x401280) is 100.00% similar to the original
...
```
(notice how the recompiled `IsleApp::IsleApp` starts at `0x401020` - this is because of that class)

For the sake of accuracy, I'm proposing to substitute the workaround with a solution in the build system. ~~I've created a basic script that wraps the compilation step and filters out all the "warning C4786" spam. Even though that is kind of ugly, it's somewhat less ugly than polluting our decompilation binary with stuff that shouldn't be in it.~~

~~Alternatively we could also do without an extra script and let everyone take care of the warning spam in their environment themselves, which I wouldn't mind either.~~

Decent solution added by @madebr, see below.

Let me know what you think about this.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,7 @@ root = true
 
 [*.{py,txt,editorconfig}]
 indent_style = space
-indent_size = 2
+indent_size = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,10 @@ jobs:
         # Use minimum supported version
         cmake-version: '3.13.x'
 
+    - name: Patch MSVC 4.2
+      run: |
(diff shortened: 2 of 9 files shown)


foxtacles:
@tntexplosivesltd Maybe you can suggest a fix for the python lint issue here. Thank you!

madebr:
See foxtacles/isle#3

PR #313, opened 2023-12-07
Refactor MxList cursors

disinvite:
Third and final `MxCollection` refactoring, following #290 and #283. I also added a new `MxList` of `LegoWorld` pointers to `LegoOmni`. Template type confirmed by debugger.

For cursors of a specific list type, we had been using a typedef on `MxListChild` or `MxListChildChild` for the specific situation. This PR replaces those with a dedicated class for the `MxList` derivative being used. `MxListChildChild` was used in conjunction with an `MxPtrList` so I replaced those with `MxPtrListCursor` and it lines up to the same number of vtables.

There is a slight added code overhead. Each cursor now needs its own initializer list to pacify the compiler. For example:
```cpp
// VTABLE: LEGO1 0x100d7e50
class MxDSActionListCursor : public MxListCursor<MxDSAction*> {
public:
    MxDSActionListCursor(MxDSActionList* p_list) : MxListCursor<MxDSAction*>(p_list){};
};
```

Fortunately, this lines up perfectly with the 96 source, so I think this is a reasonable solution to our problem.

Accuracy has fluctuated a little due to compiler weirdness, though it looks like a slight improvement overall. The biggest hit is in `MxRegion::vtable18`, but I think this is also compiler weirdness. The problem is that the previous method had all three cursor vtables assigned in a row, but we now call a constructor for `MxListCursor<MxRegionTopBottom *>` and this throws the stack out of alignment. I think this is okay, because the method right under this (`MxRegion::vtable1c`) uses the same cursor but does _not_ call a constructor. Meaning: this change didn't necessarily break anything.

Finally, I added a whole mess of decomp markers for any list cursors we were already using. I didn't always put the `// VTABLE` markers in virtual address order. I don't think we have any control over how the compiler orders these and I thought it helped readability to have the vtables for the superclass chain in a row where possible. Happy to change that after the review.

### DIFF
--- a/LEGO1/mxdsselectaction.cpp
+++ b/LEGO1/mxdsselectaction.cpp
@@ -42,7 +42,7 @@ void MxDSSelectAction::CopyFrom(MxDSSelectAction& p_dsSelectAction)
 // SYNTHETIC: LEGO1 0x100cbc90
 // MxListCursor<MxString>::`scalar deleting destructor'
 
-// SYNTHETIC: LEGO1 0x100cbd00
+// FUNCTION: LEGO1 0x100cbd00
 // MxStringListCursor::~MxStringListCursor
 
 // FUNCTION: LEGO1 0x100cbd50
--- a/LEGO1/mxdsselectaction.cpp
+++ b/LEGO1/mxdsselectaction.cpp
@@ -33,6 +33,18 @@ void MxDSSelectAction::CopyFrom(MxDSSelectAction& p_dsSelectAction)
 		this->m_unk0xac->Append(string);
 }
 
+// SYNTHETIC: LEGO1 0x100cbbd0
+// MxStringListCursor::`scalar deleting destructor'
+
+// TEMPLATE: LEGO1 0x100cbc40
+// MxListCursor<MxString>::~MxListCursor<MxString>
+
+// SYNTHETIC: LEGO1 0x100cbc90
+// MxListCursor<MxString>::`scalar deleting destructor'
+
+// SYNTHETIC: LEGO1 0x100cbd00
+// MxStringListCursor::~MxStringListCursor
+
 // FUNCTION: LEGO1 0x100cbd50
 MxDSSelectAction& MxDSSelectAction::operator=(MxDSSelectAction& p_dsSelectAction)
 {
@@ -130,3 +142,6 @@ void MxDSSelectAction::Deserialize(char** p_source, MxS16 p_unk24)
 
 	*p_source += extraFlag;
 }
+
+// TEMPLATE: LEGO1 0x100cc450
+// MxListEntry<MxString>::GetValue
--- a/LEGO1/mxregion.cpp
+++ b/LEGO1/mxregion.cpp
@@ -86,6 +86,24 @@ void MxRegion::vtable18(MxRect32& p_rect)
 	m_rect.UpdateBounds(p_rect);
 }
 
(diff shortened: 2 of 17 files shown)


LEGO1/mxregion.cpp
```diff
@@ -167,6 +200,15 @@ void MxRegionTopBottom::FUN_100c5280(MxS32 p_left, MxS32 p_right)
 	}
 }
 
+// TEMPLATE: LEGO1 0x100c54f0
+// MxListCursor<MxRegionLeftRight *>::MxListCursor<MxRegionLeftRight *>
+
+// TEMPLATE: LEGO1 0x100c5560
```
foxtacles, line 206: `FUNCTION` I think

LEGO1/mxregionlist.h
```diff
@@ -23,13 +23,17 @@ class MxRegionList : public MxPtrList<MxRegionTopBottom> {
...
-// VTABLE: LEGO1 0x100dcc10
-// class MxListCursorChildChild<MxRegionLeftRight *>
-typedef MxListCursorChildChild<MxRegionLeftRight*> MxRegionLeftRightListCursor;
+// VTABLE: LEGO1 0x100dcba0
+// class MxListCursor<MxRegionTopBottom *>
+
+// VTABLE: LEGO1 0x100dcb88
+class MxRegionListCursor : public MxPtrListCursor<MxRegionTopBottom> {
+public:
+	MxRegionListCursor(MxRegionList* p_list) : MxPtrListCursor<MxRegionTopBottom>(p_list){};
```
foxtacles, line 35: Apparently if you change the type from `MxRegionList*` to `MxPtrList<MxRegionTopBottom>*` the match of `vtable18` bounces back to 80%+. I have no earthly idea why that is, but I think I'd make the change and add a comment explaining that using `MxRegionList*` somehow leads to a significantly worse match. Maybe we can figure it out later.

foxtacles:
Looks good to me, much cleaner, just a few comments. Thank you!

LEGO1/mxdsselectaction.cpp
```diff
@@ -33,6 +33,18 @@ void MxDSSelectAction::CopyFrom(MxDSSelectAction& p_dsSelectAction)
...
+// SYNTHETIC: LEGO1 0x100cbbd0
+// MxStringListCursor::`scalar deleting destructor'
+
+// TEMPLATE: LEGO1 0x100cbc40
+// MxListCursor<MxString>::~MxListCursor<MxString>
+
+// SYNTHETIC: LEGO1 0x100cbc90
+// MxListCursor<MxString>::`scalar deleting destructor'
+
+// SYNTHETIC: LEGO1 0x100cbd00
```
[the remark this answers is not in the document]
disinvite: Ahhh yes. Thanks. I blame github

LEGO1/mxregionlist.h
disinvite: This is done. Only a minor drop in accuracy instead of a plunge. It might be having trouble on the cast to/from `MxRegionList` in the initializer. I left the decomp markers as they were; we'll be back in this file later to finish it up anyway.

foxtacles: I think you don't need the dummy class. The code you had before worked for me, I literally only changed the type the constructor took. Can you reproduce that?

foxtacles (approved):

PR #315, opened 2023-12-07
Implement/match more of MxCompositePresenter

foxtacles:
This adds 3 more functions:

```
+  MxCompositePresenter::StartAction (0x100b6410) is 100.00% similar to the original
+  MxCompositePresenter::EndAction (0x100b65e0) is 100.00% similar to the original
+  MxCompositePresenter::Notify (0x100b6760) is 100.00% similar to the original
```

`m_unk8c` in `MxDSAction` has been named `m_origin`, since it is always set to the class instance that created the action in the first place.

### DIFF
--- a/LEGO1/mxcompositepresenter.cpp
+++ b/LEGO1/mxcompositepresenter.cpp
@@ -1,7 +1,11 @@
 #include "mxcompositepresenter.h"
 
 #include "decomp.h"
+#include "mxactionnotificationparam.h"
+#include "mxautolocker.h"
+#include "mxdsmultiaction.h"
 #include "mxnotificationmanager.h"
+#include "mxobjectfactory.h"
 
 DECOMP_SIZE_ASSERT(MxCompositePresenter, 0x4c);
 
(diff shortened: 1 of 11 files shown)


LEGO1/mxcompositepresenter.cpp
```diff
@@ -43,34 +50,106 @@ MxCompositePresenter::~MxCompositePresenter()
...
+		NotificationManager()->Send(
+			action->GetOrigin(),
+			&MxEndActionNotificationParam(c_notificationEndAction, this, action, FALSE)
+		);
+	}
 }
 
-// STUB: LEGO1 0x100b6760
+// FUNCTION: LEGO1 0x100b6760
 MxLong MxCompositePresenter::Notify(MxParam& p)
```
ItzSwirlz, line 130: I might be wrong, and I feel like we've had this discussion before but given the 1996 source are these 'MxS32'?

foxtacles: The function is exported with `long` as a return type, so it must be `MxLong`.

PR #316, opened 2023-12-07
(Discussion/Proposals) Consistency regarding annotations of header-implemented functions

foxtacles:
Based on recent Matrix chats it became apparent that there's still some (important) work to be done related to our function annotations, specifically regarding the functions which have their implementation in header files.

In this PR I'll briefly summarize the issue and the approaches we have considered so far. The goal is to eventually reach consensus and attain consistency in the code base (supported by tooling).

# What is the problem?

Currently, it is undecided where we want to place the annotations for **functions that have their implementations in header files** (or none at all, such as the `scalar deleting destructor`). These sometimes get inlined, but whenever they are not, they require a function annotation. Examples include:

```
	inline virtual const char* ClassName() const // vtable+0c
	{
		return "MxCore";
	}
```

or many of the templated functions in [`mxlist.h`](https://github.com/isledecomp/isle/blob/1485e5df4711f284a1768a45ffd226f73b63736b/LEGO1/mxlist.h).

The original practice has been to attach the annotation to the function directly, or use the "comment" syntax for templated functions, but all annotations remained in `.h` files:

```
	// FUNCTION: LEGO1 0x1000bfe0
	inline virtual const char* ClassName() const override // vtable+0xc
	{
```
```
	// TEMPLATE: LEGO1 0x100c3c50
	// MxPtrListCursor<MxRegionTopBottom>::~MxPtrListCursor<MxRegionTopBottom>
```

However, this started being an issue with our new tooling, specifically`checkorder`. Some functions in header files cannot be declared in ascending order by their address, since it can be in conflict with their placement in the virtual table (or make no logical sense).

# Possible approaches

There's generally been two approaches to this problem, both of which have been implemented already in parts. However, the current inconsistency is not desirable and should be resolved asap.

## Strictly moving all annotations to `.cpp` files

Advantages include:

* No more issues with `checkorder`; all materialized functions appear in order
* Each `.cpp` file serves as documentation for a contiguous block of functions that can be found in the original binary in exactly this order. This aids in the decompilation effort and general understanding of dependencies.

Disadvantages include:

* Readability suffers somewhat due to intermittent annotations in between function implementations, and the original functions are no longer annotated directly in IDEs. Example: https://github.com/isledecomp/isle/blob/master/LEGO1/mxcompositepresenter.cpp#L22
* Finding the right compilation unit to put the function annotation in can be tricky/impossible since we still have a lot of gaps in the decompilation.

Required changes:

* Extending `checkorder`/linter to make sure no function annotations appear in `.h` files
* Moving all annotations of header-implemented functions to `.cpp` files

## Strictly moving all annotations to `.h` files

Advantages include:

* More intuitive, since annotations can be attached directly to some functions (but not all)
* No need to consider in which `.cpp` file to put the annotation

Disadvantages include:

* `checkorder` needs to be disabled for `.h` files; order of functions no longer considered
* No obvious documentation on where the functions end up being materialized by the compiler in the binary

Required changes:

* Disabling `checkorder` for `.h` files
* Moving all annotations of header-implemented functions to `.h` files

# Summary

Generally, two important concerns are being touched upon by this issue: documentation of contiguous blocks of functions and the readability/maintainability of the annotations.

I believe that there is merit to both approaches. Personally I lean a little bit towards the solution of moving everything to `.cpp` files, since it allows us to gain additional understanding of when/where header files have been pulled in (dependency discovery) directly in the code base, without having to reference an extra tool. I do realize though this comes at the cost of some readability and maintainability.

Regardless of which approach we end up deciding on, it is almost certain we will require some form of documentation as to where functions ended up in the binary. For this purpose, @tahg has created [a spreadsheet](https://docs.google.com/spreadsheets/d/1a-uMSbQTEm9k7Ci6qrTiOBmJyFZzYUkbVI4jvn0AUFo/edit#gid=0), which we could consider of adopting to the project as well, especially if it can be integrated with the results of `reccmp` based on the `master` branch.

Let me know what you think!


tahg:
As per discussion in Matrix, I feel like moving to .h files only is not an option.  The rules for ordering vtables and function order are complex and dependent on at minimum both overloading and overriding of functions.  I feel this still leaves us with one of 2 options.  We can leave stuff as it is, with functions being annotated where they are defined, or we can list all functions in a cpp. This second option also has two sub options, for a given compilation unit, we can either list functions that "belong" to that class, or list everything that we know is pulled in from that class.

Personally, I feel we should keep annotations where they are, with in an option in checkorder to "check", "warn", or "ignore" header files, with warn essentially the same as check, but not throwing an error exit code.  The workflow scripts should be things we enforce on any new or existing code, and while a perfect ordering of the binary is something we are striving for, we may not match exactly, and certainly don't currently.

I believe we currently collect enough information to compare our order to the original with a separate tool, but I feel that would be better implemented on a global scope against a master list of functions (I'm currently working on one.)

disinvite:
I would vote against the "move it all to header files" option. I think we gain a lot from having the `// FUNCTION` markers right next to the functions in .cpp files. I use my own Ghidra database (but routinely check the server too) so it's helpful to see what things have already been finished and quickly find the offset. I would imagine this is also helpful for future new contributors. It's a small thing, but just having those two pieces of information together in one file is intuitive.

Having _everything_ in the .cpp files has benefits too, and I would vote for that if these are the only two options. Everyone's editor will color the comments differently and the decomp markers stand out so it's no problem for me to scan past them visually.

I raised this in Matrix chat but another alternative (or compromise) is to put _some_ things in a third type of file. It could be a .cpp file that we don't compile as @tahg suggested. Maybe that file should only be a temporary escrow for symbols that don't yet belong anywhere else.

But if those options are all (ultimately) about pacifying `checkorder`, let's just change the tool to not enforce on header files.

foxtacles:
Just to clarify (I'm not 100% sure if I was clear), this PR is _only_ about header-implemented functions (i.e. functions that have their implementation in header files). The "regular" functions with implementations in compilation unit keep their annotations in `.cpp` files no matter which option we choose.

foxtacles:
Based on most of the feedback so far the preferred solution seems to be moving the annotations back to `.h` files, which is reasonable. I've made the changes in this PR - see the above commit.

Now, `checkorder` naturally fails for some `.h` files, which is why we need to disable it for `.h` files, or at least make it an opt-in check.

It would also be great if we can add a linter rule to the tooling that emits an error when any "comment syntax" function annotation is found in a `.cpp` file. This should never happen, since by definition* those markers are only used for header-implemented functions, and thus should not appear in `.cpp` files.

* technically you can define templates and such in `.cpp` files as well, but that is a very unusual practice and we don't do it anywhere, so I'd say for the sake of maintaining consistency we can make it a rule to only allow "comment syntax" in `.h` files.

PR #324, opened 2023-12-12
Decomp linter tool

disinvite:
(Draft for the moment because I'm still picking at this, but it is basically done. I thought I might solicit feedback because this is taking a while.)

`checkorder` is now `decomplint`.

The tool looks at the decomp markers and prints any syntax warnings or errors that we get for free by using the `DecompParser` class. The new class `DecompLinter` takes the output from `DecompParser` and runs additional checks with the markers from the entire file, and the main one is checking the order of implemented functions (i.e. those not referenced by name) in `.cpp` files.

We don't (IIRC) have any cases where a function or other symbol has multiple decomp markers, but this is possible and supported by the parser. For example:

```cpp
// FUNCTION: ISLE 0x1234
// FUNCTION: LEGO1 0x10002000
// MyClass::MyFunction
```

It would not make sense to check function order by just naively reading the markers in sequence, because the virtual address space is different. So we have to check for a single module only. This is done with the `--module` flag. If you omit this, we skip most of the "linter" checks and you just get the same syntax checking that you already got from `checkorder`.

My reasoning for calling a problem a syntax warning versus an error is this: for a given problem, if we can still gain something from reading the marker, it is a warning. If we are forced to skip something, it is an error. Implemented functions out of order is also an error. With `--enforce` mode enabled, we can cause warnings to return a failure status with the `--warnfail` flag.

Per the discussion in #316, we want to move all "byname" markers to header files. There is a new syntax warning `byname_function_in_cpp` for this situation, and there are plenty of them in the current code base.

Lastly, the error reporting for `decomplint` is (I think) a little better now. I imitated the output of `clang-format` and alerts are ordered by line number. I don't know whether we need to actually print any "details" for each alert. If we do print anything right now, it is just the text from the line where the syntax warning occurred.


foxtacles:
Looks great! Could you change the target of this PR to be merged into https://github.com/isledecomp/isle/pull/316? I'd say we should also add `--warnfail` to the GitHub action, but with the changes in this PR there shouldn't be any more issues.

Thank you!

disinvite:
Merging into #316 by way of foxtacles#4

PR #333, opened 2023-12-15
some gamestate functions

MishaProductions:
I am 100% sure that LegoGameState::m_unk0xa6 is an array, however I don't know the size of it.

### DIFF
--- a/LEGO1/legogamestate.cpp
+++ b/LEGO1/legogamestate.cpp
@@ -271,18 +271,18 @@ void FUN_1003ccf0(MxU16*, LegoFileStream&)
 }
 
 // FUNCTION: LEGO1 0x1003cdd0
-void LegoGameState::SerializeScoreHistory(MxS16 m_flags)
+void LegoGameState::SerializeScoreHistory(MxS16 p_flags)
 {
 	LegoFileStream stream;
 	MxString savePath(m_savePath);
 	savePath += "\\";
 	savePath += g_historyGSI;
 
-	if (m_flags == LegoStream::WriteBit) {
+	if (p_flags == LegoStream::WriteBit) {
 		WriteScoreHistory(&m_unk0xa6);
 	}
 
-	if (stream.Open(savePath.GetData(), (LegoStream::OpenFlags) m_flags) == SUCCESS) {
+	if (stream.Open(savePath.GetData(), (LegoStream::OpenFlags) p_flags) == SUCCESS) {
 		FUN_1003ccf0(&m_unk0xa6, stream);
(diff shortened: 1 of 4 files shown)


LEGO1/legogamestate.cpp
```diff
@@ -255,14 +261,51 @@ void LegoGameState::RegisterState(LegoState* p_state)
 	m_stateArray[targetIndex] = p_state;
 }
 
-// STUB: LEGO1 0x1003cdd0
-void LegoGameState::SerializeScoreHistory(MxS16)
+void LegoGameState::WriteScoreHistory(MxU16*)
```
foxtacles, line 264: Needs a function annotation

LEGO1/legogamestate.cpp
```diff
@@ -255,14 +261,51 @@ void LegoGameState::RegisterState(LegoState* p_state)
...
+{
+}
+
+// STUB: LEGO1 0x1003ccf0
+void FUN_1003ccf0(MxU16*, LegoFileStream&)
```
foxtacles, line 269: Since `LegoGameState::WriteScoreHistory` is part of `LegoGameState`, and both are called with `m_unk0xa6`, shouldn't both be part of the `LegoGameState` class?

MishaProductions: It doesn't seem like ```FUN_1003ccf0``` is part of the LegoGameState
image

foxtacles: Yes, but why do you think `WriteScoreHistory` is? Both functions are being called the same way:

image

MishaProductions: Oh, I see what you mean, sorry

foxtacles:
> I am 100% sure that LegoGameState::m_unk0xa6 is an array, however I don't know the size of it.

Looks like a structure to me, with `m_unk0xa6` being a length/counter follwed by an array with elements of size `0x2c` each and a length of `20`. Source: `FUN_1003c830` which appears to initialize this structure, called from the constructor of `LegoGameState`

MishaProductions:
Yeah, that makes more sense.

foxtacles:
I've added a preliminary `ScoreStruct` structure in the place of `m_unk0xa6`. The two functions `WriteScoreHistory` and ``FUN_1003ccf0` are members of this structure, which also led to `LegoGameState::SerializeScoreHistory` matching.

The structure itself appears to be initialized in the constructor of `LegoGameState` at `1003c830`, which might be the constructor of the structure.

PR #334, opened 2023-12-15
Move redist tools to central location

disinvite:
We redistribute a few programs with the python scripts in our `tools` directory so that they can work. Specifically:
- `cvdump.exe` is used by `reccmp` for parsing the PDB file
- `DUMPBIN.EXE` is used by `verexp` to quickly look at the export symbols

There are also these files in the `verexp` directory that may or may not be dependencies of `DUMPBIN.EXE`. (Is this a wine compatibility thing?)
- `LINK.EXE`
- `MSPDB41.DLL`

This change moves all these files to `tools/isledecomp/isledecomp/lib` and provides a simple module `isledecomp.lib` to easily locate the path of the tools.

The reason is this: the tools can now be used in a script from any directory, without having to look up their path either absolutely or relatively.

To use `cvdump.exe` as an example, it is referenced by `isledecomp.SymInfo`, but until now it was only possible to use that module from `reccmp`. The `get_file_in_script_dir` helper method looks at `sys.argv[0]` to decide the base path for `cvdump.exe`, and so it expects that value to be `path/to/tools/reccmp.py`.


foxtacles:
Looks good, but there seems to be an issue in the CI

disinvite:
Looks OK now. We were not pulling in the non-python files when the `isledecomp` package is installed. It worked on my system when installing the package with `pip install -e`.

PR #335, opened 2023-12-15
Additional isle functions

MishaProductions:
All implemented functions match except Isle::Create


foxtacles:
Looks good! Matched `Isle::Create` to 100%.

PR #338, opened 2023-12-16
implement Isle::Notify

MishaProductions:
couldn't figure out how to match it


foxtacles:
Matched `Isle::Notify`. It currently shows ~89%, but it's actually 100% because of the issue with jump tables that `reccmp` has.

PR #340, opened 2023-12-17
implement MxDSBuffer object creation

MishaProductions:


foxtacles:
Looks good, matched all the functions to 100%

PR #342, opened 2023-12-17
Decomp linter warning for offset reuse

disinvite:
The `decomplinter` tool will now alert to offsets being reused in a module when scanning the code base. You can call the `reset()` method with argument `True` to "fully" reset the linter's state.

I fixed the handful of files that had this new problem. Most were `// GLOBAL` markers pointing at the wrong string.


foxtacles:
Very useful, thank you!

PR #346, opened 2023-12-21
implement EnableFullScreenMovie and FUN_100ba640

MishaProductions:
Couldn't figure out how to match both functions, however the logic seems correct


foxtacles:
Matched both functions to 100%. Maybe take a look at the bitwise operations, the logic wasn't quite correct there - it's easy to get confused by what they do though. Looks good now.

PR #347, opened 2023-12-21
Add unknown render class stub

MishaProductions:
Not exactly sure what this class does, but it has a list and render settings?


LEGO1/MxUnknown100dbdbc.h
```diff
@@ -0,0 +1,16 @@
...
+#include "mxtypes.h"
+
+// VTABLE: LEGO1 0x100dbdbc
+// SIZE 0x14
+class MxUnknown100dbdbc {
+public:
+	MxUnknown100dbdbc();
+
+private:
+	char m_unk0x4[0x14]; // TODO: change to 0x10 once scalar deconstructor is added
```
foxtacles, line 13: Should be `undefined`. The file names should be all lower-case

LEGO1/legoomni.cpp
```diff
@@ -385,7 +386,7 @@ void LegoOmni::Init()
 	MxOmni::Init();
 	m_unk0x68 = 0;
 	m_inputMgr = NULL;
-	m_unk0x6c = 0;
+	m_renderMgr = 0;
```
ItzSwirlz, line 389:
```suggestion
	m_renderMgr = NULL;
```

PR #348, opened 2023-12-21
Implement/match MxSmack::LoadFrame

foxtacles:
`MxSmack::LoadFrame` loads a frame. If there is a palette chunk - which means the palette has changed - it will be processed before the actual frame will be decoded by the Smacker SDK using the function `SmackDoFrameToBuffer`. Match is at 100%.

What remains unclear is the offset of the palette, which seems to be off by 4 byte for us. Maybe the version of the Smacker SDK we have is the problem here and the structure (past the header) has been different at the time; that would probably be the most likely scenario.

### DIFF
--- a/LEGO1/mxsmack.cpp
+++ b/LEGO1/mxsmack.cpp
@@ -6,7 +6,7 @@ DECOMP_SIZE_ASSERT(SmackTag, 0x390);
 DECOMP_SIZE_ASSERT(MxSmack, 0x6b8);
 
 // FUNCTION: LEGO1 0x100c5a90
-MxResult MxSmack::LoadHeaderAndTrees(MxU8* p_data, MxSmack* p_mxSmack)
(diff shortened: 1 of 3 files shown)


LEGO1/mxsmack.cpp
```diff
@@ -140,15 +139,88 @@ void MxSmack::Destroy(MxSmack* p_mxSmack)
...
+		MxU8 palette[772];
+
+		MxU8* intoChunk = p_chunkData + 1;
+		MxU8* intoPalette = palette;
+		MxU16 paletteIndex = 0;
+		// TODO: struct incorrect, Palette at wrong offset?
+		MxU8* currentPalette = &p_mxSmack->m_smackTag.Palette[4];
+
+		do {
+			if (*intoChunk & 0x80) {
```
ItzSwirlz, line 172: This may help:
https://github.com/FFmpeg/FFmpeg/blob/45b4781e9a19cb4c9b2bf6e7018087698067fb74/libavformat/smacker.c#L268C7-L276C42

PR #353, opened 2023-12-22
Refactor WinePathConverter into PathResolver

disinvite:
The `WinePathConverter` class was used as:
1. a wrapper around calls to `winepath`, for converting to and from Wine paths
2. for converting Windows-like paths in the `cvdump` output to match Unix paths returned by `os.walk`

This refactor to `PathResolver` doesn't add much in the way of features, but I think the added separation of concerns and documentation is useful. I'm working on adding more checks and comparisons to our palette with `cvdump` and the `SymInfo` module that drives it, and the fact that there was not a unified API across platforms for this path conversion was a sticking point.

If we are not on Windows, the goal in this and the previous module is actually to _avoid_ calling `winepath` whenever possible. This call is expensive, although this is mitigated a lot by having the `wineserver` daemon running in the background.

The other problem I aim to solve here is dealing with the wonky paths produced by MSVC. For example, my PDB has references to these two paths:

`C:\isle\LEGO1\viewmanager/../realtime/vector.h`
`C:\isle\LEGO1\realtime/vector.h`

They are the same file, but the way `SymInfo` groups the line numbers from `cvdump` together will cause us to miss lines for one of the paths. (And so we actually gain one new function for `reccmp` with this change.) I suspect this would become more of a problem later on as we move some of the code files out of the root LEGO1 directory and into subdirectories. (Probably matching the way the 96 source did it.)

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -1,0 +1,36 @@
+name: Unit Tests
+
+on: [push, pull_request]
+
+jobs:
+  pytest-win:
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Install python libraries
+      shell: bash
+      run: |
+        pip install pytest -r tools/requirements.txt
+
+    - name: Run python unit tests (Windows)
+      shell: bash
+      run: |
+        pytest tools/isledecomp
+
+  pytest-ubuntu:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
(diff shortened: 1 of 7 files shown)


foxtacles:
Looks good! Do you think you could add a CI action for the `isledecomp` tests?

disinvite:
Oh yeah! Good opportunity to add that in. I'm testing on both Windows and Ubuntu because we might as well cover all the bases. I tried to think of a way to design this so the tests _would_ be platform-independent, but didn't find one that I liked.

PR #358, opened 2023-12-24
MxDisplaySurface::SetPalette()

ItzSwirlz:
Incomplete as of opening the pr


LEGO1/mxdisplaysurface.cpp
```diff
@@ -216,9 +248,63 @@ void MxDisplaySurface::Clear()
 	this->Reset();
 }
 
-// STUB: LEGO1 0x100baae0
+// FUNCTION: LEGO1 0x100baae0
 void MxDisplaySurface::SetPalette(MxPalette* p_palette)
 {
+	HDC hdc;
```
disinvite, line 254: Something must be missing here at the start, because the original grabs `0x414` from the stack and we're using about `0x400` less than that. They could be unused variables. Maybe two MxU16 arrays of 256 colors?

LEGO1/mxdisplaysurface.cpp
```diff
@@ -216,9 +248,63 @@ void MxDisplaySurface::Clear()
...
+	byte bVar6;
+	byte bVar7;
+	if (((this->m_surfaceDesc).ddpfPixelFormat.dwFlags & 0x20) != 0) {
+		this->m_ddSurface1->SetPalette(p_palette->CreateNativePalette());
+		this->m_ddSurface2->SetPalette(p_palette->CreateNativePalette());
+		if (((this->m_videoParam).Flags().GetFullScreen() & 1) == 0) {
+			lpal.palVersion = 0x300;
+			// lpal.palNumEntries = 256;
+			// FIXME: this loop may be incorrect
+			memcpy(lpal.palPalEntry, NULL, sizeof(lpal.palNumEntries));
```
disinvite, line 271: I think you want `memset` here

LEGO1/mxdisplaysurface.cpp
disinvite: Actually this might be another case like the `BITMAPINFO` wrapper in `MxBitmap` where they have defined a wrapper around `LOGPALETTE` where the `palPalEntry` array has 256 entries instead of 1. That would put the size of `lpal` at `0x404` or 1028 bytes

foxtacles:
I've made the helper functions to 100%. What @disinvite said is probably correct and has also been applied. The function still doesn't match because the stack allocation isn't quite right yet, so something needs to be further changed around. However, logically it should be accurate now.

PR #361, opened 2023-12-25
Implement MxDSBuffer::FUN_100c67b0

MishaProductions:
With this function implemented, LegoWorldPresenter is created from isle.si

### DIFF
--- a/LEGO1/mxdsbuffer.cpp
+++ b/LEGO1/mxdsbuffer.cpp
@@ -136,7 +136,7 @@ MxResult MxDSBuffer::FUN_100c67b0(
 
 	do {
 		while (TRUE) {
-			data = (MxU8*) SwapBuffers();
+			data = (MxU8*) SkipToData();
 			if (data == NULL) {
 				return SUCCESS;
 			}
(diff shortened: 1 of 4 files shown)


LEGO1/mxdsbuffer.cpp
```diff
@@ -119,15 +120,65 @@ MxResult MxDSBuffer::SetBufferPointer(MxU32* p_buffer, MxU32 p_size)
...
+				result = SUCCESS;
+			}
+		}
+
+		MxResult createObjectResult = CreateObject(p_controller, (MxU32*) data, p_action, p_streamingAction);
+		if (createObjectResult == FAILURE) {
+			return FAILURE;
+		}
+		result = createObjectResult;
+	} while (result != 1);
```
ItzSwirlz, line 179:
```suggestion
	} while (result != FAILURE);
```

MishaProductions: `FAILURE` is -1, not 1

MishaProductions: I just noticed that 1 is returned at 0x100c6b96 in ParseChunk

PR #376, opened 2023-12-27
Improve handling of variables for decomp parser

disinvite:
Another refactor of the decomp parser to provide the following features:

- Extract the name of global variables. This is so we can cross-reference the variable name with the output from cvdump. If the data is initialized, we can compare it.

- If we find a variable inside a function, assume it is a static variable and set the `is_static` flag. In the `GLOBALS` section of cvdump, these variables are identified with `S_LDATA32`. Static variables of a _class_, however, are still `S_GDATA32` like regular global variables. We will not be able to match the names of those exactly without making the parser a lot more aware of class/function structure. We might be able to match them anyway by process of elimination. We'll cross that bridge when we come to it.

- Add a new marker type `// LIBRARY` to mark functions like `_malloc` and `_free` that come from the standard library. We (probably) don't need to compare these so this is effectively an alias for `// STUB`. We can't match these yet anyway, because it requires us to parse the `PUBLICS` section from cvdump, and that piece is unfinished. When it is time to annotate these, we could add a .h file to the codebase (so it will be picked up by the parser) but not include it in any .cpp files.

- Add a new marker type `// STRING` to indicate constant string values that appear on the next line. Right now, though, we are just ignoring these markers. My reason for changing these is to make them distinct from `// GLOBAL` markers that indicate a variable.

We don't actually need to use the `// STRING` markers, but I thought that we might as well keep them around since they are already here in the code. To match strings up, we can take advantage of two facts:
1. String constants are set into `.rdata` only once, and then referenced by pointer wherever they are used.
2. Strings are null-terminated.

So my strategy (in an upcoming PR) will be:
1. Use cvdump to find all string constants in the recomp. There is a specific mangled symbol prefix that will identify these.
2. In the original binary, look at relocated vaddrs from .rdata.
3. Try to match up string data from 1 with locations in 2.

The medium-long term goal here is to have a complete map of symbols between the original binary and the recomp. That will serve two main purposes:
1. When comparing code, we are substituting virtual addresses with strings like `<OFFSET1>` to help with text comparison via difflib. If we know that a certain address in the original maps to a certain address in the recomp, we can say for certain that the address used in the recomp is correct. For example: we can check whether a call to `_malloc` or any other function in the original is the same in the recomp.
2. We can compare symbols that are not code, like variables and vtables.

### DIFF
--- a/ISLE/define.h
+++ b/ISLE/define.h
@@ -8,9 +8,9 @@ class IsleApp;
 
 extern IsleApp* g_isle;
 extern int g_closed;
-// GLOBAL: ISLE 0x4101c4
+// STRING: ISLE 0x4101c4
 #define WNDCLASS_NAME "Lego Island MainNoM App"
-// GLOBAL: ISLE 0x4101dc
+// STRING: ISLE 0x4101dc
 #define WINDOW_TITLE "LEGO\xAE"
 extern unsigned char g_mousedown;
 extern unsigned char g_mousemoved;
--- a/LEGO1/act1state.h
+++ b/LEGO1/act1state.h
@@ -12,7 +12,7 @@ public:
 	// FUNCTION: LEGO1 0x100338a0
 	inline virtual const char* ClassName() const override // vtable+0x0c
 	{
-		// GLOBAL: LEGO1 0x100f0154
+		// STRING: LEGO1 0x100f0154
 		return "Act1State";
 	};
 
--- a/LEGO1/act2brick.h
+++ b/LEGO1/act2brick.h
@@ -16,7 +16,7 @@ public:
 	// FUNCTION: LEGO1 0x1007a360
 	inline virtual const char* ClassName() const override // vtable+0x0c
 	{
-		// GLOBAL: LEGO1 0x100f0438
+		// STRING: LEGO1 0x100f0438
 		return "Act2Brick";
 	}
 
--- a/LEGO1/act2policestation.h
+++ b/LEGO1/act2policestation.h
@@ -12,7 +12,7 @@ public:
 	// FUNCTION: LEGO1 0x1000e200
 	inline virtual const char* ClassName() const override // vtable+0x0c
 	{
-		// GLOBAL: LEGO1 0x100f03fc
+		// STRING: LEGO1 0x100f03fc
 		return "Act2PoliceStation";
 	}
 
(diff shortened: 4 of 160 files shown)


foxtacles:
Looks good!

Would the `// LIBRARY` marker also be appropriate for functions like the ones from the Smacker lib? https://github.com/isledecomp/isle/blob/master/LEGO1/mxsmack.h#L15

With regards to the `// STRING` marker, I know there are more strings embedded in functions such as: https://github.com/isledecomp/isle/blob/master/LEGO1/mxstreamcontroller.cpp#L273 - would it make sense to annotate them as well, and would it be technically be possible to compare their contents eventually?

disinvite:
Hmm... maybe the answer is to just not skip `// LIBRARY` functions after all. Should I switch that around?

That case doesn't fit nicely as a name-referenced `// FUNCTION` marker because there's no connection to any code that we wrote. Also: looking at cvdump, the Smacker functions are all under `PUBLICS`, same as the MSVC library stuff. So maybe `// LIBRARY` just serves to direct `reccmp` for how to match up the symbol, but we still compare it and add it to our total of identified functions.

That would leave us with:

| Marker | Referenced by... | Should skip | User code |
|---|---|---|---|
| FUNCTION | line or name | | yes |
| STUB | line or name | yes | yes |
| LIBRARY | name | | |

As for strings, I'm pretty confident we can match everything without the markers. I didn't want to just delete all the ones we already have because it still can be valuable documentation.

That string "MxDSMultiAction" is also under `PUBLICS` in cvdump, and it looks like: `??_C@_0BA@ONG@MxDSMultiAction?$AA@`

The `??_C` prefix means a constant string.

foxtacles:
> That would leave us with:
> 
> Marker	Referenced by...	Should skip	User code
> FUNCTION	line or name		yes
> STUB	line or name	yes	yes
> LIBRARY	name		

Yeah I think that would be great, especially if `reccmp` can actually compare the library functions as well. There will be more use cases for `LIBRARY` soon since they did use another C library for FLIC decoding as well which we will add eventually.

Detecting unmarked strings would be good and I'd make it a requirement as soon as we have that capability (can be a separate PR though). Being able to compare the contents and thus verifying the correctness can save tons of hours of debugging in the future, so I think it's worth it.

foxtacles:
Merging, comparison of library functions, string annotation checking etc. can well be done in separate PRs.

PR #377, opened 2023-12-27
Finish LegoPalettePresenter

MishaProductions:
100% match of both functions

### DIFF
--- a/LEGO1/legopalettepresenter.h
+++ b/LEGO1/legopalettepresenter.h
@@ -25,9 +25,10 @@ public:
 		return !strcmp(p_name, ClassName()) || MxVideoPresenter::IsA(p_name);
 	}
 
-	MxResult ParsePallete(MxStreamChunk* p_chunk);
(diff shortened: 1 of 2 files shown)


LEGO1/legopalettepresenter.h
```diff
@@ -25,13 +25,15 @@ class LegoPalettePresenter : public MxVideoPresenter {
 		return !strcmp(p_name, ClassName()) || MxVideoPresenter::IsA(p_name);
 	}
 
-	virtual void Destroy() override; // vtable+0x38
+	MxResult ParsePallete(MxStreamChunk* p_chunk);
```
foxtacles, line 28: General remark, it would be nice to separate virtual from non-virtual functions with a blank line. The way we did it in most places was to first list all virtual functions in a block, then followed by another block with the other functions.

I will check if `clang-format` can help with this, but in the meantime it would be nice to do it manually for readability.

PR #383, opened 2023-12-28
Cvdump parser and comparing library functions

disinvite:
This adds a more structured framework for parsing the output from `cvdump` when analyzing our recomp PDB files.

To use it, you can import `Cvdump` from `isledecomp.cvdump` and use the fluent interface to set which options you want. e.g.

```python
    cv = (
        Cvdump(pdb)
        .lines()
        .symbols()
        .publics()
        .section_contributions()
        .run()
    )
```

The final call to `run()` will return the parser object with the parsed results already loaded.

You can omit the component calls to subsections like `.lines()` and it will run `cvdump` with all options enabled. This is the same thing the command line `cvdump` does if you don't specify any flags.

As a proof of concept, we are now comparing against a select group of library functions found under `PUBLICS`, and annotated in `library_msvc.h`. After comparing a few of those functions, we found that `ISLE` was missing some of the symbols that it should have, and that its `entry` method (registered as the symbol `_WinMainCRTStartup`) did not match. The solution is to link the multi-threaded `LIBC` library, `LIBCMT.LIB`, and so I have changed `CMakeLists.txt` accordingly.

We can't compare the Smacker library functions yet. They do not have any size information registered in the `SECTION CONTRIBUTIONS` piece from `cvdump`. Also, they are linked in their own segment `_UNSTEXT` instead of `.text`. The solution is to get all the offsets from `_UNSTEXT` listed under `PUBLICS` in `cvdump` and (possibly) guess where each function ends. Refining the output from the parser will be its own project in another PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -292,7 +292,7 @@ if (MSVC)
   # game was originally built with) and tweaked slightly to produce more debugging info for reccmp.
   # They ensure a recompilation that can be byte/instruction accurate to the original binaries.
   if (ISLE_BUILD_APP)
-    target_compile_options(isle PRIVATE "/ML$<$<CONFIG:Debug>:d>")
+    target_compile_options(isle PRIVATE "/MT$<$<CONFIG:Debug>:d>")
   endif()
   target_compile_options(lego1 PRIVATE "/MT$<$<CONFIG:Debug>:d>")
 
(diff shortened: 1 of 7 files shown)


foxtacles:
Looks good, thank you! Would be great if we can eventually compare the Smacker stuff too, but this is a very good start.

PR #391, opened 2023-12-30
Implement MxDisplaySurface::VTable0x28

foxtacles:
This adds the implementation for `VTable0x28`, which was the last piece necessary to render something from the initial Smacker video in `nocd.si`.

The match isn't great at ~30% and it will probably take a lot of further experimentation to get it right, since it's a very large function again and minor differences in variable placement/usage, inline functions etc. are highly important. Functionally I believe it should be accurate, although it's easy to miss something with a function of this size, so I'm inviting everyone to check it out and look for any issues. The 16-bit branches are the one currently most relevant.

Many variable names are also still placeholders or should be improved - suggestions welcome.

https://github.com/isledecomp/isle/assets/1135351/160b91bb-d31c-4aa9-87ab-1fbb29bdd428


foxtacles:
Match is at 70%+ now - merging

PR #402, opened 2024-01-04
implement MxDirect3D::GetZBufferDepth

MishaProductions:
doesn't fully match due to some issue with the _D3DDeviceDesc / MxAssignedDevice structure. Based off of the leaked source code


foxtacles:
Since I'm working on the same stuff right now, I already have a fully matching version of this function - pushed it. Also already mostly into the remaining stuff in the area.

PR #405, opened 2024-01-04
Reccmp comparison engine refactor

disinvite:
Up until now, our m.o. for comparing the original and recomp files has been:
1. Populate temp structures in `syminfo` by reading the output of `cvdump`.
2. Read each decomp marker from each file in the source directory and try to match the function.

This refactor changes that around so we now read all possible information first, then run comparisons on each symbol where we have addresses in both binaries. The direct benefit is that we can verify that pointer values from the original and recomp point at the same thing, or that both point to `NULL`. This will help with comparing:

- Assembly that refers to a pointer or calls a function
- Pointer variables
- Vtables

This PR doesn't add any of those new comparisons yet (sorry!) but I think enough files are changed here that it deserves its own review. Deciding the output for the comparisons (in both the terminal and HTML versions) will be its own project.

Summary of changes:
- The `isledecomp.compare` module has two databases — one for symbols and one for line references checked with `// FUNCTION` markers. For this, I used an actual in-memory `sqlite` database. It performs slightly better (on my end) than the previous implementation.
- Each module gets its own `logging` instance where needed, instead of passing around the same object.
- The `DecompCodebase` class aggregates all decomp markers from a single module (`LEGO1` or `ISLE`) so that the code for reading those is more idiomatic. We also now have the option to import all "lineref" markers first; I think this will sidestep the potential problem with `MxDeviceEnumerateElement` or other symbols with a duplicated name.
- The `CvdumpAnalysis` module collects the results gained from parsing cvdump output. We still need the raw output from `CvdumpParser` to get at the `LINES` section. One role of the analysis module is to take all the symbols and estimate their approximate size by taking the distance between offsets. The information from `SECTION CONTRIBUTIONS` also does this, but I found it will overshoot the mark in some cases. We take the minimum of this value and our own calculation to get the symbol "size" if we do not have a confirmed size to use.
- We are doing some limited demangling of MSVC symbols to get information about string constants and vtables. This may be all we need, but we have the option to add the command line `undname` tool if there is more information worth getting.
- On the topic of string constants, we can match these up relatively easily with the original binary by checking likely string locations (via `.reloc`) in the `.data` or `.rdata` sections. The reason to match these up is that many of our pointer references point at strings — all those `ClassName()` functions, for example. When we extend the parser to look at `// STRING` markers, we can eliminate the need for this brute force lookup.

The biggest external change will be the output from `reccmp`. Because we are reading everything first before comparing, the `[ERROR]` messages when we fail to match all appear up front, followed by the comparisons. The functions are now listed in the order of their virtual address from the original, so the diff for `progress.txt` in the github action is going to be wonky for this PR. None of the match percentages should change, though.


foxtacles:
Looks good, great work. As always, thank you for the comprehensive summary! Looking forward to all the extra comparisons this will enable.

PR #410, opened 2024-01-06
Fixes for building isle with mingw on Linux

madebr:
Functional difference should be zero.
The messages of the commits explain the changes.

As an extra, I added resources to LEGO1.DLL.


foxtacles:
Looks good. I've applied the following changes:

* The references to `Destroy` in `MxPtrList<T>` were mixed up
* `MxStreamListMxDSAction::Find` requires the `iterator it` to be declared in the loop, otherwise vastly different assembly will be generated with MSVC4. Since I couldn't find a workaround, I introduced the `COMPAT_MODE` macro to allow specifying alternative code in (rare) cases like these
* Naming/formatting

Let me know if it still compiles, then it should be good to merge.

madebr:
It still compiles, with these warnings:
```
isle/LEGO1/tgl/d3drm/impl.h:82:21: warning: inline function 'void TglImpl::RendererImpl::Destroy()' used but never defined
   82 |         inline void Destroy();
      |                     ^~~~~~~
isle/LEGO1/mxstreamer.h:17:16: warning: inline function 'MxStreamerSubClass1::MxStreamerSubClass1(undefined4)' used but never defined
   17 |         inline MxStreamerSubClass1(undefined4 p_size);
      |                ^~~~~~~~~~~~~~~~~~~
```

(After applying https://github.com/madebr/isle/commit/a08bab8a7e8646ffc19b72d679f3e987df28515c)

foxtacles:
OK, feel free to make a new PR where you apply the changes in https://github.com/madebr/isle/commit/a08bab8a7e8646ffc19b72d679f3e987df28515c to the code base using `COMPAT_MODE` if you want.

Maybe the use of this would even be preferable over changing interfaces with `COMPAT_CONST`.

PR #411, opened 2024-01-06
Fix all mingw errors & warnings

madebr:
When building with mingw gcc 10 from ubuntu 22.04, I saw a couple more warnings.

In the last 2 commits, I also fixed the "inline function used but not defined warning".
If those decrease accuracy, perhaps these aren't inlined in the first place?

(I did not yet test the resulting exe + dll on Windows)


foxtacles:
Looks good to me.

To match functionality precisely, you should add a scope in those instances where a "temporary" is passed by reference, otherwise the destructor will not be called at the right time. So instead of:

```
#ifdef COMPAT_MODE
	MxAtomIdCounter id_counter(m_internal);
	MxAtomIdCounterSet::iterator it = AtomIdCounterSet()->find(&id_counter);
#else
	MxAtomIdCounterSet::iterator it = AtomIdCounterSet()->find(&MxAtomIdCounter(m_internal));
#endif
```

it would be:

```
#ifdef COMPAT_MODE
	{
		MxAtomIdCounter id_counter(m_internal);
		MxAtomIdCounterSet::iterator it = AtomIdCounterSet()->find(&id_counter);
	}
#else
	MxAtomIdCounterSet::iterator it = AtomIdCounterSet()->find(&MxAtomIdCounter(m_internal));
#endif
```

This most likely won't make a difference, but could theoretically introduce bugs in the "compat" version if not changed.

madebr:
Thanks! When I boot my Windows box, I'm going to test building with msys2 and Visual Studio 2019,
and of course also try executing them.

Thanks for the quick reviews and fast responses! It's really nice to have somebody on top of things.

PR #415, opened 2024-01-07
refresh readme

Ramen2X:
This PR updates the project README to be more in line with the project's current status. 

I've made several changes:
* Indicated that we are explicitly working from LEGO Island's 1.1 English release. It isn't very common knowledge, but there are *multiple* iterations of LEGO Island 1.1. Some localization teams opted to recompile the entire game rather than just updating the assets, producing a different, unique binary (though this isn't true for all localized versions of LEGO Island). I've also added hashes for the version of ISLE and LEGO1 that we have derived this code from to the bottom of the README, so contributors can make absolutely sure that there won't be any conflicts with what they got.
* Reworded the static report of LEGO1's decompilation status to no longer claim that the library is "very much incomplete". I feel that this is now inaccurate; especially as we are starting to visually see some progress. The claim that it "cannot be used" also *could* imply that it isn't useful to anyone, including developers. Instead, I've simply made it clear that it isn't usable for gameplay.
* Changed some technical wording structure from singular to plural. We are working on two binaries in this project rather than one, so I feel that this makes more sense.
* Added LEGO Island's registry paths to mentions of the registry. I felt that it was odd before that there was indication that you could create a portable environment for LEGO Island by updating the registry but that there was no further elaboration on this. I've addressed this.

Additionally, I've made several grammar adjustments and fixed some typos.

Of course, I'd like to hear what everyone else thinks of these adjustments. I'm open to any suggestions or changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ If you have a CMake-compatible IDE, it should be pretty straightforward to use t
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


foxtacles (approved):
Looks good to me!

Ramen2X:
Merging - if anyone wants to request any changes, you're welcome to open a PR.

PR #419, opened 2024-01-08
Implement most of LegoCameraController

MishaProductions:
Implemented most of the subclasses and LegoCameraController. Most of the code was present in the leaked source code. The LegoCameraController class just seems to interface with an 3d view. 100% matches on all functions except the constructor of LegoCameraController.


LEGO1/lego/legoomni/include/legocameracontroller.h
```diff
@@ -1,17 +1,21 @@
...
+#include "legopointofviewcontroller.h"
 #include "mxcore.h"
+#include "mxpoint32.h"
 #include "realtime/matrix.h"
 #include "realtime/vector.h"
 
 // VTABLE: LEGO1 0x100d57b0
 // SIZE 0xc8
-class LegoCameraController : public MxCore {
+class LegoCameraController : public LegoPointOfViewController {
```
ItzSwirlz, line 12: Is it called LegoPointOfViewController in the source or just LegoPOVController

MishaProductions: LegoPointOfViewController

PR #421, opened 2024-01-10
Support building isle with modern MSVC + msys2

madebr:
- Enable decomilation assertions when Visual C++ version <= 5.0
  Visual C++ 4.2.0 == cl version 10.20 ([wikipedia](https://en.wikipedia.org/w/index.php?title=Microsoft_Visual_C%2B%2B&section=4#Internal_version_numbering))
- Disable assertions for modern C++ + mingw: the size assertions did not fail in minw@Linux, but failed when building with Visual Studio 2019 or msys2 mingw.
- Pass `-Zc:__cplusplus` to make the `__cplusplus` macro contain the actual `__cplusplus` version (needed for `static_assert`)
- `LegoCameraController` overrides methods that are not declared/implemented in its parents
- Don't forward declare namespaced classes as `class Tgl::Device` but as `namespace Tgl { class Device; }`.
  msys2 mingw complained with the following warning: `declaration 'class Tgl::Device' does not declare anything`
- `ViewLODList::Dump` does not have an implementation
- `Tgl::Device::GetDrawnTriangleCount` does not exist: my fix is WRONG
- Remove `APP_ICON` from lego/res/resource.h`

Using these patches, I can run `ISLE.EXE` + `LEGO1.DLL` built with mingw + Visual Studio, and let them fail identically.

For me, the stacktrace is:
```
LegoVideoManager::Create(MxVideoParam &, unsigned int, unsigned char) legovideomanager.cpp:109
LegoOmni::Create(MxOmniCreateParam &) legoomni.cpp:510
IsleApp::SetupLegoOmni() isleapp.cpp:138
IsleApp::SetupWindow(HINSTANCE__ *, char *) isleapp.cpp:588
WinMain(HINSTANCE__ *, HINSTANCE__ *, char *, int) isleapp.cpp:218
invoke_main() 0x0000000000e171cd
__scrt_common_main_seh() 0x0000000000e17027
__scrt_common_main() 0x0000000000e16ebd
WinMainCRTStartup(void *) 0x0000000000e17248
```
`driver` in [`LegoVideoManager::Create`](https://github.com/isledecomp/isle/blob/5a2be934be68db6eea90b2711a3cad07284788a8/LEGO1/lego/legoomni/src/video/legovideomanager.cpp#L109) is `NULL`

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -4,6 +4,7 @@ on: [push, pull_request]
 
 jobs:
   decomplint-isle:
+    name: 'ISLE annotations'
     runs-on: ubuntu-latest
 
     steps:
@@ -18,6 +19,7 @@ jobs:
         python3 tools/decomplint/decomplint.py ISLE --module ISLE --warnfail
 
   decomplint-lego1:
+    name: 'LEGO1 annotations'
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,45 @@ name: Build
 on: [push, pull_request]
 
 jobs:
+  build-current-toolchain:
+    name: 'Current ${{ matrix.compiler.name }}'
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        compiler:
+        - { name: 'msvc', setup-ninja: true, setup-cmake: true }
+        - { name: 'msys2 mingw32', setup-msys2: true }
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup cmake
+        if: matrix.compiler.setup-cmake
+        uses: jwlawson/actions-setup-cmake@v1.13
+        with:
+          # Use minimum supported version
+          cmake-version: '3.13.x'
+
+      - name: Setup ninja
+        if: matrix.compiler.setup-ninja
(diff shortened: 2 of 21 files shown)


foxtacles:
Looks good, my changes:

* The overrides were a mistake, they were indeed new functions introduced in `LegoCameraController`. Another example why we should add a modern compiler to our GitHub test suite to catch issues like these.
* `SetWorldTransform` is actually `const` in the original but wasn't in ours; fixed and thus no need for `COMPAT_MODE`
* Fixed offset annotations
* `LegoView1 is 16 bytes bigger then LegoView ==> 4 32-bit pointers`: with regards to this: I don't know why, but it appears MSVC 4.20 is adding an extra 4 byte padding at the end. There is no 4th member. The size of the class remains `0x88`, no matter if you add the padding member or not. Removed for this reason. Let me know if this causes any issues.

madebr:
I added 2 GitHub ci jobs building isle using latest MSVC + msys2 mingw32.

madebr:
>  Let me know if this causes any issues @madebr - if so, we could only add it with `COMPAT_MODE`.

This pr disables all `DECOMP_SIZE_ASSERT`'s when building with mingw32 and newer msvc, so it does not matter much anymore.
It looks like the size of STL structures vary greatly within the mingw world.

foxtacles:
Thank you, very useful!

PR #426, opened 2024-01-12
Matrix/vector refactor

foxtacles:
This is a somewhat larger PR that refactors the existing matrix and vector classes. Those classes have long been a pain point; it wasn't obvious how they were originally structured/designed, a lot of code using them barely matched, inconsistencies arose when it came to interfacing them with TGL and other modules, and their internals didn't seem consistent overall.

What I can say for certain is that most of the matrix/vector code underwent a major refactor between the 1996 source and retail release. In conjunction with investigation into the alpha debug build (which signifcantly differs in parts to the 1996 source) I've made several findings, which resulted in what is more or less a rewrite of those classes.

* Almost certainly, all matrix and vector functions were defined in the header files. This has become evident after I've ran into several functions that made use of the functions, but with the code inlined. Keep in mind that the compiler only *sometimes* inlines the functions whenever it can and deems appropriate (since they are all virtual functions, the complete type must be known) - as such, I've moved all the functionality to headers. This enabled matches in several functions.
* Many of the vector/matrix functions use overloads for their basic operations. Overloaded virtual functions are placed in the virtual table in reverse order, but the code is generally materialized in the binary in the order of appearance of declarations. I've rewrote functions as far as I could; however, there's at least 1-2 cases where apparently, overloads exist on so far unknown alias types. Since these are yet to be determined, I've left those functions with different names instead.
* Both vector and matrix classes are unlikely to have used a storage vector/matrix class as their internal data representation, for two reasons: a) it makes their operation implementations much uglier; b) no traces of such proxy classes exist in the alpha debug source. Since the debug version never inlines any functions, there'd need to be calls to `operator[]` at the very least to access the members of a vector/matrix internal representation. Such calls are nowhere to be found. (also assuming they didn't directly access a public member; but that seems unlikely to me as well). A third reason is that building the base classes on top of a storage class essentially renders their apparently intended use case - which is to generically wrap an existing, outside raw storage - much more cumbersome.
* Instead, they directly operate with float arrays `float[3]`, `float[4]` or `float[4][4]` and their respective pointers. Access operators are defined directly in their base classes. This enables all previous use cases we had so far the same way as before.
* They generally used `memcpy` and `memset` to copy vector/matrix storage or set it to zero, not compiler-implemented copy-construction of storage classes or similar.
* The most derived classes, which we so far have called `Vector3Data`, `Matrix4Data` etc. were most likely called `Mx3DPointDouble` and `MxMatrix` originally. Evidence for those names exists in the 1996 source.
* Originally, the aforementioned classes were used sparingly across the code base. Most notably, they were only used in Omni/LegoOmni, but modules like TGL, Realtime, Viewmanager etc. made use of their own vector/matrix implementations, such as `Tgl::FloatMatrix4` or `Vector3`. This has changed drastically; in the retail version, the classes with Omni origin are now used in all of those places; with the exception of TGL.
* The `Tgl::DoubleMatrix4` (which we renamed `FloatMatrix4` due to their switch to `float`) is no longer being used in retail. Instead, it is now simply defined as `float[4][4]`. Again, no wrapper class appears to be used, at least based on code in the alpha debug and matches so far.

In consequence, I've rebranded the existing classes as follows:

`Vector2Impl` → `Vector2`
`Vector3Impl` → `Vector3`
`Vector4Impl` → `Vector4`
`Matrix4Impl` → `Matrix4`

Those classes remain part of `realtime`. Their main purpose is to be able to allow vector/matrix operations, with the data source provided externally by passing in a pointer to floats. It is hard to say if they were really part of `realtime` after their refactor; I've no conclusive evidence for either option.

`Vector3Data` → `Mx3DPointFloat`
`Vector4Data` → `Mx4DPointFloat`
`Matrix4Data` → `MxMatrix`

Those classes are now part of the `mxgeometry` module. Headers such as `mxgeometry3d.h` and `mxmatrix.h` existed based on the 1996 source. They are based on the above classes and come with internal storage, and also define assignment/copy-constructions from their base class instances.

Based on all the changes, I've matched the following "vector/matrix-heavy" functions to 90-100% (and several more close to it) to prove correct implementation as far as possible:

* `LegoCameraController`: `GetWorldUp`, `GetWorldLocation`, `GetWorldDirection`
* `Helicopter::VTable0xd8`
* `ViewROI::UpdateWorldData`
* `LegoView1::Create`

Not all details are completely clear yet. The usage patterns are rather inefficient in many places, with unnecessary copies and implict conversions between `float*`, base classes and Omni classes - probably in part because of the major refactor they performed on those classes in 1996. That makes it rather hard to figure out their exact implementation, but I think we are reasonably close now.

Generally, when using those classes, one should keep the following non-obvious details in mind:

* Virtual functions may be inlined if the type is known - but not always. Exact conditions are unknown.
* It is generally not obvious whether arguments to functions which are references to a vector/matrix use the base class or the `Mx` class. Code may not match if it's not the right one.
* Local instantiations of the base classes are sometimes optimized by the compiler so that no virtual table is assigned, which makes it seem like no object is constructed at all (due to inline constructors); however, this happens in a few places, where the base class is used as a wrapper for a separate data source only for the operations.


Ramen2X (approved):
Looks good! Thanks for taking the time to research, organize, and consolidate all of this. Dealing with the matrix classes was a nightmare working on #241 and I had a feeling what we were seeing in the '96 source wasn't necessarily representative of the final implementation. The current structure was also fueling my hesitation to look into any further logic that used them.

PR #427, opened 2024-01-12
fix LegoPathPresenter vtable

MishaProductions:


LEGO1/lego/legoomni/include/legopathpresenter.h
```diff
@@ -22,6 +23,8 @@ class LegoPathPresenter : public MxMediaPresenter {
 		return !strcmp(p_name, LegoPathPresenter::ClassName()) || MxMediaPresenter::IsA(p_name);
 	}
 
+	virtual void ReadyTickle();               // vtable+0x18
```
foxtacles, line 26: Should have override

PR #432, opened 2024-01-13
Implement Isle::VTable0x50

MishaProductions:
Isle::VTable0x50 doesn't match 100% because of this weird padding that I couldn't figure out in Act1State:
image
This PR allows the InfoCenter script to be started


foxtacles:
> Isle::VTable0x50 doesn't match 100% because of this weird padding that I couldn't figure out in Act1State

Doesn't seem to be padding, just a series of individual bytes, probably each of them representing a boolean. I've added them to the constructor since they are initialized as well (which remains incomplete; just for the sake of illustration)

PR #433, opened 2024-01-13
Reccmp: Use symbol names in asm output

disinvite:
Until now, we have been replacing addresses in the assembly output with a placeholder string `<OFFSETX>`. "X" is an incrementing id that correponds to the order in which we have replaced something in the given function. We do this because the virtual addresses used in original and recomp have no chance of matching until the binaries are at similar size; i.e. when we are nearly finished with the decomp. We'd like to know whether things match sooner than that.

The pitfall of using placeholders is that things like function calls can appear to match even though they point at two different things in the respective asm output. This change adds the ability to use the name of the referenced symbol instead of the placeholder string, so it will be obvious when something doesn't match.

To make this possible, we are now calling the `capstone` function `disasm_lite` with the starting address of the given function. This has the effect of resolving any relative offsets referenced in the code with their absolute address. For `call` instructions, this helps a lot because our lookup is very simple. However, all jump instructions will no longer match, which is the exact problem we are trying to solve. They had matched previously because we assumed both original and recomp functions started at address 0, and also because we replaced the immediate operand for `jmp` instructions with the placeholder string.

To fix that, I changed the output for jumps so you now get the relative offset (or displacement), regardless of the starting address of the function.

In other words: assume we have a jump that creates an infinite loop:
```asm
start:
  jmp start
```
This will give you the machine code `eb fe`. If this function started at 0x1000, and we provided that to `capstone`, we would get `jmp 0x1000` as the output. We now get `jmp -0x2`, which is a little more helpful to see why a jump is not taking you to the right place.

However... `jmp` is also used in the unwind section of some functions, which is part of exception handling. Here the intent is to jump directly to the start of a destructor function. With this change, if we have the name of that function, we use it, but otherwise we show the (probably very large) jump offset instead. These had been silently matching before, so it will be helpful to see where we have gaps in annotation coverage.

Finally, I moved things around in `reccmp` for (hopefully) the final time so we have a nice structure for doing comparisons and producing terminal/html/svg output. I plan to add more comparison types after this and we needed a better structure to have concise code.

You may notice that the accuracy has dropped for ISLE and LEGO1. I've added some missing decomp markers to mitigate that, but there are more things to add that probably merit a review on their own. That brings us to:

__What doesn't match (and why)__

- Imports. **We are now loading and matching these.** ~These are the `__imp__GetFileSize@8` functions. We can read these by checking the `.idata` section of the original file and then matching up with the PDB. This should be an easy thing to enable and I have it half implemented in the stash.~
- Class static variables. They look like this: `MxTimer::g_lastTimeCalculated`. The regular `// GLOBAL` marker will register that variable as `g_lastTimeCalculated` because that's all you see on the line of code. This might be tricky to get because we'll have to rework the parser to add state-awareness of when we are inside of a class or struct. That probably involves counting the curly braces and potentially doing two passes on each file.
- The empty string. The symbol name is `??_C@_00A@?$AA@`, which indicates a zero length string with a null terminator. In other words: a zero byte, which could be anywhere. The correct location in `LEGO1` is `0x100f01d4`, but we can't detect this automatically. We will have to just identify this with a `// STRING` marker and enhance the parser to use them.
- Variables that are hidden to us. One example is the absurd `_Tree<MxAtomIdCounter *,MxAtomIdCounter *,set<MxAtomIdCounter *,MxAtomIdCounterCompare,allocator<MxAtomIdCounter *> >::_Kfn,MxAtomIdCounterCompare,allocator<MxAtomIdCounter *> >::_Nil`. To get these and others, we will need to enable synthetic `// GLOBAL` markers. Easy to do.
- Jump tables from a `switch` statement. These have never matched. The `jmp` instruction into an index of the table is very characteristic of these cases, and (iirc) the jump tables themselves are referenced somewhere in the `cvdump` output. With that knowledge we could sort of guess at where these are supposed to be for each function and add a new comparison type.


foxtacles:
Great changes, this is a major improvement. Being able to see the symbol names in the diff instead of random `OFFSET` markers is very helpful. I can immediately spot bugs that were previously undiscovered, such as:

image

I have a few questions, looking at this example:

image

You mentioned that imports cannot be matched up yet (the ones starting with `__imp`). Is it the same when it comes to a symbol such as:

`call ?VideoManager@@YAPAVLegoVideoManager@@XZ`

If yes, I think it would be best if we can add this as soon as possible, since the overall match percentage of `ISLE` is going down by a lot (no surprise since it's using a lot of imported functions) - however, this makes it much harder to quickly discover regressions that may be accidentally introduced during development, because they'd be drowned out in all the mismatching imports noise. (notwithstanding compiler randomness in some functions, I usually look for drops from 100% to <100%)

Side note: would it be possible to show the demangled symbol in the diff instead?

image

I'm not sure how the diff works, but In this example (`ISLE 0x4023e0`) I'd expect only the calls to be marked as different, not the instructions in between. Is that a limitation or could it be improved?

image

I'm assuming this is a case of a function that we simply didn't annotate yet. Would it be possible for `reccmp` to also, in addition, show the original address of that function in `LEGO1`? This would make it faster for us to craft the annotation necessary to fix the documentation issue. I suppose the tool could even generate the entire annotation for us, although just adding the address to the `(FUNCTION`) bit at the end of the line would be just fine for now.

image

I probably just don't understand the inner workings, but shouldn't the `OFFSET` markers be identical? What causes the diff (`17` vs `18`) and would it possible to rectify? (if not it's not a big deal, mostly wondering what the reasons are)

Really great work so far, I'll keep checking this out and possible add more comments.

LEGO1/omni/include/mxsemaphore.h
```diff
@@ -6,6 +6,7 @@
 #include <windows.h>
 
 // SIZE 0x08
+// VTABLE: LEGO1 0x100dccf0
```
foxtacles, line 9: `VTABLE` should come before `SIZE`

LEGO1/omni/include/mxvideopresenter.h
```diff
@@ -80,7 +80,7 @@ class MxVideoPresenter : public MxMediaPresenter {
 	// FUNCTION: LEGO1 0x1000c800
 	virtual MxS32 GetHeight() { return m_alpha ? m_alpha->m_height : m_bitmap->GetBmiHeightAbs(); }; // vtable+0x84
 
-	// SIZE 0xc
```
foxtacles, line 83: `SIZE` removed on purpose?

disinvite:
In order:

1. Yes, those are all imports. It seems like it's the `LEGO1` symbols have the mangled name, but the ones from Windows libraries get the `__imp__` prefix. I'm not sure why. The mangled ones say `(UNK)` right now because we can't guarantee what that symbol is from the `cvdump` analysis alone. That's by design: we can assume some things about the symbols by context or name but depend on the decomp markers to say how we are supposed to match/compare something.

You make a very good point that the time between pulling this and the follow-up PR with imports will be a problem. I'll look into finishing up the imports feature so we can add it onto this PR (if you agree).

2. I can look into demangling the names but that will probably be a second PR because we can still match if it is mangled. In this case, we could just assign a more friendly name because we know these are the imported functions.

3. I'm not sure what's happening here either. We are relying on `SequenceMatcher` from `difflib`, which is the [LCS algorithm](https://en.wikipedia.org/wiki/Longest_common_subsequence). Maybe the fact that the string with the symbol name is a lot longer is throwing off the calculation? There is an option to pass an `isjunk` method into `SequenceMatcher` that will direct the algorithm to ignore some differences. Maybe ignoring space characters would help here, if it is not doing that already. I can try playing around with this.

4. We can include the original address, good idea. Ideally we would have a tool that would generate the new decomp markers pending a review by the user. The design I landed on with encapsulating everything in the `ParseAsm` object might help here because we can just compare the `replacements` dict. We had talked before about detecting these cases and providing a special error (or maybe just set match percentage to zero) and that's probably the way we can get that feature too.

5. This is a tricky one. The placeholder index number should tick up whether we find the actual name or use a placeholder, and that happens here for the recomp diff. The problem is that we did _not_ use a placeholder or name in the original diff, because we don't want to make an assumption of what this is. Now, in this case, the jump offset is huge, so it would be reasonable to say that we should use a placeholder if `abs(offset) > some_limit`, but we can get unlucky if the destructor is right under the unwind section.

6. For those `// SIZE` markers, I wasn't sure whether one of my changes to the parser had caused it to no longer ignore those comments, but I can check this again to be sure.

disinvite:
Correction to # 4 above. We _may_ be able to report the original address, but only if we have it. In the cases from your screenshot, we are missing the annotations on that function, so there's nothing to match.

We could maybe get this by comparing the replacements list afterward, but that would be reliant on the offset numbers lining up. I might be able to fix the jump offset from # 5 by just ticking the placeholder number up at every jump whether we use a placeholder or not, but if the recomp function genuinely has fewer offsets than the original then this won't come together automatically.

foxtacles:
> You make a very good point that the time between pulling this and the follow-up PR with imports will be a problem. I'll look into finishing up the imports feature so we can add it onto this PR (if you agree).

Yeah, that sounds good to me!

Regarding the issue #4: that's too bad, but no big deal. Shouldn't happen too often anyway and if there is a mismatch we can get the address manually just fine.

disinvite:
Another correction: The `__imp__*` symbols are the entries in the Import Address Table from `.idata`. The corresponding symbols without that prefix are the thunks: `jmp dword ptr [__imp__*]`. The reason our `LEGO1` symbols are mangled is that they are C++ symbols; the others are C symbols, which are also mangled but... less so.

_Finding_ the thunks is another challenge. I'm debating whether we should just annotate these in a file (as we have with the library functions) and move on. I think the only other alternative is to just search for blocks of `jmp` instructions in the original and check whether those jump into `.idata`.

foxtacles:
One issue I've found is that globals declared in namespaces can't be properly matched up (or so it seems). Example is `Tgl::CreateRenderer (0x100a15e0)`. If I move the `g_pD3DRM` in the global namespace, it works fine, but as long as it's in `TglImpl::` it doesn't seem to.

foxtacles:
Something I've noticed, some symbols seem to be cutoff:

image

Is that a MSVC limitation?

disinvite:
Yes, globals inside of classes aren't matching because we can't match on the full `Class::Member` name. We could cheat and use a partial match, provided that the variables names themselves are unique. I think `g_instance` is used in a few different classes, though.

If that symbol name is cut off at 256 characters, I wonder if this is the downstream effect of our favorite warning C4786.

For the thunks, I think what I'll do is write the code to search through the binary for the `jmp` instructions and see how it does. If it performs well, we can just add it to the script. If not, I'll put the output into an annotations file and check that in.

foxtacles:
Alright, I'm about halfway through fixing up missing annotations as much as possible. A `reccmp` pass that would only search for missing or wrong symbols, cases like these:

image

would certainly help a lot. As discussed before, I think we should then also be able to emit a hard error in the CI action in case those are found, since those cases are either bugs or fall into the missing annotation category (which is also important).

foxtacles:
The `print-rec-addr` flag to `reccmp` doesn't seem to be working anymore. Apart from that, this is good to merge in my opinion!

PR #444, opened 2024-01-16
begin implementation of InfoCenter class

MishaProductions:
With this class, the game plays the intro animation. MxDisplaySurface::VTable0x44 needs to return something other than NULL for this to work.

### DIFF
--- a/LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
+++ b/LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
@@ -1,9 +1,37 @@
 #include "infocenter.h"
 
-// STUB: LEGO1 0x1006ea20
+#include "infocenterstate.h"
+#include "legocontrolmanager.h"
+#include "legogamestate.h"
+#include "legoinputmanager.h"
+#include "legoomni.h"
+#include "legoutil.h"
+#include "legovideomanager.h"
+#include "mxactionnotificationparam.h"
+#include "mxbackgroundaudiomanager.h"
+#include "mxnotificationmanager.h"
+#include "mxstillpresenter.h"
+
+DECOMP_SIZE_ASSERT(Infocenter, 0x1d8)
+
+// FUNCTION: LEGO1 0x1006ea20
 Infocenter::Infocenter()
 {
-	// TODO
+	m_unk0xfc = 0;
+	m_unk0x11c = 0;
+	m_infocenterState = NULL;
+	m_unk0x11c = 0;
+	m_unk0x104 = 0;
+	m_unk0xf8 = -1;
+	m_currentCutScene = -1;
+	memset(&m_entries, 0, sizeof(InfocenterUnkDataEntry) * 7);
+	m_unk0x1c8 = -1;
+	SetAppCursor(1);
+	NotificationManager()->Register(this);
+	m_unk0x1d0 = 0;
+	m_unk0x1d2 = 0;
+	m_unk0x1d4 = 0;
+	m_unk0x1d6 = 0;
 }
 
 // STUB: LEGO1 0x1006ec90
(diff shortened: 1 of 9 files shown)


LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
```diff
@@ -15,20 +43,135 @@ Infocenter::~Infocenter()
 // STUB: LEGO1 0x1006ed90
 MxResult Infocenter::Create(MxDSAction& p_dsAction)
 {
-	return FAILURE;
+	OutputDebugString("create called\n");
```
ItzSwirlz, line 46: for reference: make sure we remove this before putting into the PR (or do a compiler macro to enable/disable debug strings)

LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
```diff
@@ -15,20 +43,135 @@ Infocenter::~Infocenter()
...
+
+	if (m_worldStarted) {
+
+		switch (((MxNotificationParam&) p_param).GetNotification()) {
+		case 0:
+			return HandleNotification0(p_param);
+		case c_notificationEndAction:
+			return HandleEndAction(p_param);
+		case c_notificationKeyPress:
+			return HandleKeyPress(((LegoEventNotificationParam&) p_param).GetKey()) & 0xff;
```
ItzSwirlz, line 81: Typically, & 0xff bit masks (at least from what I know) suggests types of u8's (MxU8)

LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
```diff
@@ -50,8 +260,73 @@ MxBool Infocenter::VTable0x5c()
...
+		action.SetObjectId(m_unk0xf8);
+		action.SetAtomId(*g_infomainScript);
+		action.SetUnknown24(-2);
+		DeleteObject(action);
+		m_unk0xf8 = -1;
+	}
+}
+
+// FUNCTION: LEGO1 0x100714a0
+void Infocenter::FUN_100714a0()
```
ItzSwirlz, line 316: rename to "StartSoundAnimScript"?

ItzSwirlz (approved):

LEGO1/lego/legoomni/include/legoscripts.h
```diff
@@ -0,0 +1,32 @@
...
+	InfoMainScript_PepperCharacterSelect = 541,
+	InfoMainScript_MamaCharacterSelect = 542,
+	InfoMainScript_PapaCharacterSelect = 543,
+	InfoMainScript_OfficierCharacterSelect = 544,
+	InfoMainScript_LoraCharacterSelect = 545,
+};
+
+typedef enum SndAmimScript {
+	SndAmimScript_BookWig = 400
+};
```
Ramen2X, line 30: Small styling nitpick - should these be UPPERCASE like other enums; such as `TransitionType` in MxTransitionManager?

madebr: Also in c++, you don't need to add the typedef.

tahg: My personal opinion is UPPERCASE should only be used for constants. (That is, where the value has some meaning other than just uniquely identifying a set member.)  The style guide says these should be like `e_nameOfEnumMember`, but I think this is clear enough.

Ramen2X: Didn't realize we had one in the style guide - that's my bad. It's good to get this all sorted out though.

foxtacles:
I've corrected all the enums we had and applied the Mindscape style guide:

> Items defined with the enum keyword should be prefixed with c_ or e_ depending on usage. If used to define a value, enums should be prefixed with c_. If used to define a list of related labels, enums should be prefixed with e_

This is also now enforced by the naming convention checker.

Improved accuracy for a few functions in Infocenter and fixed a few bugs as well

PR #446, opened 2024-01-17
Decomp parser: indirect globals and string markers

disinvite:
The `// GLOBAL` annotation marks where we have a variable with global scope that is declared on the next line of code. This PR adds the option to identify these variables indirectly, either because they are in generated code or library code. The syntax is the same as other annotations that allow you to pinpoint something by name reference:

```cpp
// GLOBAL: ISLE 0x4108e8
// __osver
```

Variables identified this way do not require the `g_` prefix that we use for variables in our code, for obvious reasons.

String literals are annotated with `// STRING`. It is possible to track these down without an annotation, which is what we have been doing, but it is nice to have the documentation right in the code. These are now enabled for use by `reccmp` comparison. We will still search the original binary for any strings not annotated, so 100% coverage is not a high priority item. If the string you annotated is _wrong_, we will alert you to that, too.

In many cases there is a global variable initialized to a string. You can now get the two-fer-one annotation special as in the following:

```cpp
// GLOBAL: LEGO1 0x101020b0
// STRING: LEGO1 0x10101f20
const char* g_strSOUND = "SOUND";
```

This only works for the first C string that appears on the line, so it is probably not a good idea to annotate a line like this where it would be ambiguous:

```cpp
m_backgroundColor = new LegoBackgroundColor("backgroundcolor", "set 56 54 68");
```

Lastly, strings are deduplicated in the binary to save space. If you annotate the same string in multiple places, the linter will allow it, provided that they are actually the same string.

As a bonus: annotations for the SmartHeap library for ISLE, and some examples of `// STRING` markers.

### DIFF
--- a/ISLE/library_msvc.h
+++ b/ISLE/library_msvc.h
@@ -1,0 +1,63 @@
+#ifdef 0
+// For ISLE symbols only
+
+// aka `operator new`
+// LIBRARY: ISLE 0x402f80
+// ??2@YAPAXI@Z
+
+// aka `operator delete`
+// LIBRARY: ISLE 0x402fa0
+// ??3@YAXPAX@Z
+
+// LIBRARY: ISLE 0x406dd0
+// _malloc
+
+// LIBRARY: ISLE 0x406f00
+// _free
+
+// LIBRARY: ISLE 0x407ec0
+// ___CxxFrameHandler
+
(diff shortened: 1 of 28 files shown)


foxtacles:
Looks great, thank you!

PR #450, opened 2024-01-17
implement LegoOmni::FUN_1005b4f0

MishaProductions:


foxtacles:
I've made some adjustments and matched the function 100%. What's odd is that all existing calls only pass flags together with the boolean `FALSE`, which means the flags won't be used in the function, and vice-versa. Not sure where the sense is in that right now.

PR #452, opened 2024-01-17
reccmp: vtable comparison

disinvite:
Virtual table contents are now compared by reccmp. The technique is as follows:
1. Read the vtable from `cvdump` output. The mangled symbols all begin with `??_7` and it is straightforward to derive the class name and optional template type.
2. Read the `// VTABLE` annotations from the code files, which also gives us the class name.
3. Match these in our database, where we also have the size of the recomp symbol. The size should be a multiple of 4 because it is just a list of pointers.
4. Read the data from original and recomp binaries. For each pointer in the original, consult the database to see whether the matched recomp address equals the actual recomp address read from the recomp binary.
5. If we have a match, we are done. If we could not match for any reason (i.e. there is no entry in the database for the original address, or the pointers don't point at the same thing), mark this as a diff.

To save time and avoid changing the `reccmp` terminal or html output, we just return a standard diff report for vtables as we are doing with functions. The output will tell where things aren't right. For example (InfocenterState):

```diff
-  vtable0x00  :  0x10071900 from orig not annotated.
+  vtable0x00  :  (no orig    / 0x10018120)  :  InfocenterState::`scalar deleting destructor'
   vtable0x04  :  (0x100ae1f0 / 0x10031510)  :  MxCore::Notify
   vtable0x08  :  (0x10001f70 / 0x10001090)  :  MxCore::Tickle
   vtable0x0c  :  (0x10071840 / 0x10018060)  :  InfocenterState::ClassName
   vtable0x10  :  (0x10071850 / 0x10018070)  :  InfocenterState::IsA
-  vtable0x14  :  0x10071830 from orig not annotated.
+  vtable0x14  :  (0x10005f90 / 0x100100a0)  :  LegoState::VTable0x14
   vtable0x18  :  (0x10005fa0 / 0x100100b0)  :  LegoState::SetFlag
   vtable0x1c  :  (0x10005fb0 / 0x100100c0)  :  LegoState::VTable0x1c
```

The first entry should be `InfocenterState::scalar deleting destructor`. The problem is that it was not annotated in the file, so there is no match in our database for the original address (`0x10071900`) and we also do not have a complete match for the (current) recomp address (`0x10018120`).

The entry `vtable0x14` has a different problem. The recomp shows `LegoState::VTable0x14` which is from the superclass. We do not have an annotation for the original address `0x10071830` because we have not implemented the override function `InfocenterState::VTable0x14`.

I've added all the missing `scalar deleting destructor` functions. There are (at least) three problem cases where the first vtable entry points at the regular destructor instead: LegoPathController, GifManagerBase, GifManager. These need more investigation.

The final match percentage reported by `reccmp` is still for functions only, but each vtable has its own percentage reported.

### DIFF
--- a/LEGO1/lego/legoomni/include/act1state.h
+++ b/LEGO1/lego/legoomni/include/act1state.h
@@ -29,6 +29,9 @@ public:
 
 	void FUN_10034d00();
 
+	// SYNTHETIC: LEGO1 0x10033960
+	// Act1State::`scalar deleting destructor'
+
 protected:
 	undefined m_unk0x8[0x10]; // 0x8
 	MxU32 m_unk0x18;          // 0x18
--- a/LEGO1/lego/legoomni/include/act2brick.h
+++ b/LEGO1/lego/legoomni/include/act2brick.h
@@ -25,6 +25,9 @@ public:
 	{
 		return !strcmp(p_name, Act2Brick::ClassName()) || LegoEntity::IsA(p_name);
 	}
+
+	// SYNTHETIC: LEGO1 0x1007a450
+	// Act2Brick::`scalar deleting destructor'
 };
 
 #endif // ACT2BRICK_H
--- a/LEGO1/lego/legoomni/include/act2policestation.h
+++ b/LEGO1/lego/legoomni/include/act2policestation.h
@@ -21,6 +21,9 @@ public:
 	{
 		return !strcmp(p_name, Act2PoliceStation::ClassName()) || LegoEntity::IsA(p_name);
 	}
+
+	// SYNTHETIC: LEGO1 0x1000f610
+	// Act2PoliceStation::`scalar deleting destructor'
 };
 
 #endif // ACT2POLICESTATION_H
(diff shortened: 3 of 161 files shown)


tahg:
Those classes should be changed, but the cause is known.  All of 3 of those at least, and almost certainly the case if any other destructor points into the vtable, is a case where someone marked an inline destructor with an offset that really should be the ddtor.  In some cases, there may not actually be an explicit destructor in the code.

foxtacles:
Looks great! I've fixed the virtual tables you mentioned.

The script we discussed that would (probably separately from `reccmp`) provide errors/warnings specifically for mismatching virtual tables (and other such entities like strings, function calls) would be very helpful to be able to start fixing all of the current issues. We can then add this to the CI to have a clear indication of whether those checks passed or not. Currently, since this stuff is all part of `reccmp`, it's hard to separate them out (in case of function calls and strings impossible since it's subsumed in the function accuracy scores). But this can be done in a later PR.

PR #457, opened 2024-01-19
implement/match InfoCenter::~InfoCenter()

Ramen2X:
This PR implements InfoCenter's destructor. This function is 100% matching.

It remains to be seen what the true purpose of `m_buffer` is, but I confirmed through dynamic analysis that this loop in the destructor was resolving to `MxStillPresenter::Enable()` through it; I've made this member an array of MxStillPresenters to accommodate.

### DIFF
--- a/LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
+++ b/LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
@@ -11,8 +11,8 @@
 #include "mxbackgroundaudiomanager.h"
 #include "mxnotificationmanager.h"
 #include "mxstillpresenter.h"
-#include "mxtransitionmanager.h"
 #include "mxticklemanager.h"
+#include "mxtransitionmanager.h"
 
 DECOMP_SIZE_ASSERT(Infocenter, 0x1d8)
 DECOMP_SIZE_ASSERT(InfocenterUnkDataEntry, 0x18)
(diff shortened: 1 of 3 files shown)


LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
```diff
@@ -46,10 +47,29 @@ Infocenter::Infocenter()
...
 
-// STUB: LEGO1 0x1006ec90
+// FUNCTION: LEGO1 0x1006ec90
 Infocenter::~Infocenter()
 {
-	// TODO
+	BackgroundAudioManager()->Stop();
+
+	MxS16 i = 0;
+	do {
```
foxtacles, line 56: Did you try if a regular for-loop matches as well? do/while loops are relatively rare

LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
```diff
@@ -46,10 +47,29 @@ Infocenter::Infocenter()
...
+		if (m_infocenterState->GetInfocenterBufferElement(i) != NULL) {
+			m_infocenterState->GetInfocenterBufferElement(i)->Enable(FALSE);
+		}
+		i++;
+	} while (i < 7);
```
foxtacles, line 61: I'd use a sizeof on the array instead of hard coding the number here

foxtacles (approved):

PR #458, opened 2024-01-19
Finish Bitmap

tahg:
Some if blocks are being reordered still.

### DIFF
--- a/LEGO1/omni/include/mxbitmap.h
+++ b/LEGO1/omni/include/mxbitmap.h
@@ -46,8 +46,24 @@ public:
 	// FUNCTION: LEGO1 0x1004e0d0
 	virtual int VTable0x28(int) { return -1; };
 
-	virtual void VTable0x2c(int, int, int, int, int, int, int);
-	virtual void VTable0x30(int, int, int, int, int, int, int);
+	virtual void BitBlt(
+		MxBitmap* p_src,
+		MxS32 p_left,
+		MxS32 p_top,
+		MxS32 p_right,
+		MxS32 p_bottom,
+		MxS32 p_width,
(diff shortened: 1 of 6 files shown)


LEGO1/omni/include/mxbitmap.h
```diff
@@ -46,8 +46,24 @@ class MxBitmap : public MxCore {
...
+	);
+	virtual void BitBltTransparent(
+		MxBitmap* p_src,
+		MxS32 p_left,
+		MxS32 p_top,
+		MxS32 p_right,
+		MxS32 p_bottom,
+		MxS32 p_width,
+		MxS32 p_height
+	);
```
ItzSwirlz, line 66: // vtable+30

LEGO1/omni/src/video/mxbitmap.cpp
```diff
@@ -207,14 +208,83 @@ MxResult MxBitmap::LoadFile(HANDLE p_handle)
...
+			&p_dstLeft,
+			&p_dstTop,
+			&p_width,
+			&p_height
+		)) {
+		MxU8* srcStart = p_src->GetStart(p_srcLeft, p_srcTop);
+		MxU8* dstStart = GetStart(p_dstLeft, p_dstTop);
+		MxLong srcStride = p_src->GetAdjustedStride();
+		MxLong dstStride = GetAdjustedStride();
+		for (; p_height; --p_height) {
```
ItzSwirlz, line 240: This feels a little weird, I'll investigate later

tahg: Maybe it's a while loop instead. There isn't much difference in compiled code.

PR #460, opened 2024-01-19
GifManager: use STL members

disinvite:
`GifManager` and its superclass `GifManagerBase` are using the STL types `<map>` and `<list>`. I've added those in and removed cases where we re-implemented the STL functions.

We aren't generating the `_Insert` method yet, but that should be a match based on other similar STL types. We probably won't see it until we build out `LegoTexturePresenter`.

`LegoOmni::Destroy` calls something to walk the `<map>` tree and delete nodes. This feels like an inline function that should be called, but I'm not sure which one.

Also: I modified the compare database to truncate at 255 characters when matching symbol names. This is all that gets generated in the PDB, per our favorite compiler warning C4786. This will allow us to use the unadulterated versions of the names in the decomp markers.

### DIFF
--- a/LEGO1/lego/legoomni/include/gifmanager.h
+++ b/LEGO1/lego/legoomni/include/gifmanager.h
@@ -3,73 +3,135 @@
 
 #include "compat.h"
 #include "decomp.h"
+#include "mxstl/stlcompat.h"
 #include "mxtypes.h"
 
 #include <d3drmobj.h>
 #include <ddraw.h>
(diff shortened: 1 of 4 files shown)


foxtacles:
Looks good - the issue with `goto` was easily resolved by moving the code into the loop. The compiler will generate the same code.

Not sure what's at `0xc` at the iterator either though, afaik there shouldn't be anything. Odd.

PR #461, opened 2024-01-19
Add library symbols for flic and smack

foxtacles:
This adds the symbol annotations for the flic and smack library functions.

@disinvite For `reccmp` to correctly match `_DecodeFLCFrame`, I've added a dummy symbol in `flic.asm` that designates the end of the section. This way it could figure out the length of the function accurately and match it.

For `smack.lib`, I think we have a similar problem that the function sizes can't be accurately determined and comparisons go past the actual function end, which leads to lower accuracy scores. Not sure if anything can be done about it.


disinvite:
Good idea to add that symbol. For the Smacker lib where we can't do this, one idea I had was to add a decomp annotation "hint" that will be used if present. For example:
```cpp
// LIBRARY: LEGO1 0x100cd782
// SIZE: 0x6
// _SmackGetSizeTables

// LIBRARY: LEGO1 0x100cd7e8
// SIZE: 0x1f1
// _SmackDoTables

// etc...
```
We'd just have to decide on the syntax. We already use `// SIZE` all over the place, so that name might be overloaded at this point.

foxtacles:
> Good idea to add that symbol. For the Smacker lib where we can't do this, one idea I had was to add a decomp annotation "hint" that will be used if present.

Yeah, we could do that, but I'm not sure if it's worth it - I think at some point we may want to unpack both the Smacker and Smartheap libraries and apply changes so that they match completely, and when/if we do that, we are also free to add whatever structure/symbols we need so that everything can be matched by `reccmp`.

PR #464, opened 2024-01-20
Enforce vtable match

disinvite:
A new CI action will make sure that vtables match from here on out. We do this because there is no compiler weirdness to worry about. The vtables either match or they don't. This PR also fixes the vtables we already have.

There's something odd going on with methods that have the same name but different parameters. It looks like MSVC reverses their order in the vtable. Take this example from `RendererImpl`, vtable at `0x100db910`:

```
vtable0x20 : 0x100a1e90 : TglImpl::RendererImpl::CreateUnk
vtable0x24 : 0x100a20d0 : TglImpl::RendererImpl::CreateTexture_void
vtable0x28 : 0x100a1f50 : TglImpl::RendererImpl::CreateTexture_args
vtable0x2c : 0x100a2270 : TglImpl::RendererImpl::SetTextureDefaultShadeCount
```

The two CreateTexture methods need to be in this (reverse) order in the class declaration from the abstract superclass `Renderer`:

```cpp
virtual Texture* CreateTexture(
    int width,
    int height,
    int bitsPerTexel,
    const void* pTexels,
    int pTexelsArePersistent,
    int paletteEntryCount,
    const PaletteEntry* pEntries
) = 0;
virtual Texture* CreateTexture() = 0;
```


foxtacles:
Yeah, I've commented before on the overloaded functions issue:

https://github.com/isledecomp/isle/blob/master/LEGO1/realtime/vector.h#L16

It's an important detail for sure.

disinvite:
This is ready. Hopefully all the newly created .cpp files are where they're supposed to be.

foxtacles:
Looks good! Thanks for putting in all that work, this is a big help going forward.

PR #467, opened 2024-01-20
Implement MxDisplaySurface::VTable0x44

MishaProductions:
Currently it matches at 53.74%. With this function implemented (and if you remove Enable(FALSE) in MxStillPresenter::ParseExtra), the infocenter background gets drawn:
image

### DIFF
--- a/LEGO1/omni/include/mxdisplaysurface.h
+++ b/LEGO1/omni/include/mxdisplaysurface.h
@@ -80,10 +80,15 @@ public:
 		MxS32 p_top2,
 		MxS32 p_width,
 		MxS32 p_height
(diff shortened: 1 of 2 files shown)


foxtacles:
I've applied some fixes:

* The entire logic of finding the bitmap source start is an inline function `GetStart` - this matches better (although it's not completely correct yet)
* No unsigned values are used for the loops. You can tell this is the case when you see comparisons such as `JLE`
* Fixed a few minor bugs

On paper the accuracy went down to ~48%, however this is mostly due to the stack allocation being different now. The instructions themselves match better. There's probably just a few minor key changes necessary to get the match up by a lot, but it will require a lot of try and error.

PR #469, opened 2024-01-20
Begin work on Police class

ItzSwirlz:
Currently, when setting the script to show Police, nothing happens (at least for me)

### DIFF
--- a/LEGO1/lego/legoomni/src/police/police.cpp
+++ b/LEGO1/lego/legoomni/src/police/police.cpp
@@ -1,5 +1,6 @@
 #include "police.h"
 
+#include "jukebox.h"
 #include "legocontrolmanager.h"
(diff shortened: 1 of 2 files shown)


LEGO1/lego/legoomni/src/police/police.cpp
```diff
@@ -19,3 +59,11 @@ MxLong Police::Notify(MxParam& p_param)
 
 	return 0;
 }
+
+// FUNCTION: LEGO1 0x1005e530
+void Police::VTable0x50()
+{
+	LegoWorld::VTable0x50();
+	PlayMusic(0xc);
```
Ramen2X, line 67: Can you use the JukeBoxScript enum rather than hardcoding a value here? `e_policeStation` would be equivalent to `0xc`

PR #470, opened 2024-01-20
Implement Lego3DView::Render

MishaProductions:
100% match. This allows for the sky to be rendered
image

### DIFF
--- a/LEGO1/lego/sources/3dmanager/lego3dview.cpp
+++ b/LEGO1/lego/sources/3dmanager/lego3dview.cpp
@@ -138,15 +138,12 @@ BOOL Lego3DView::Moved(ViewROI& rROI)
 	return TRUE;
 }
 
-// STUB: LEGO1 0x100ab270
+// FUNCTION: LEGO1 0x100ab270
 double Lego3DView::Render(double p_und)
 {
-	// assert(m_pViewManager);
-
-	// m_pViewManager->Update(m_previousRenderTime);
-
-	// m_previousRenderTime = LegoView1::Render();
-
+	assert(m_pViewManager);
+	m_pViewManager->Update(m_previousRenderTime, p_und);
+	m_previousRenderTime = TglSurface::Render();
 	return m_previousRenderTime;
 }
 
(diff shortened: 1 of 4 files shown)


LEGO1/lego/sources/3dmanager/lego3dview.cpp
```diff
@@ -138,15 +138,12 @@ BOOL Lego3DView::Moved(ViewROI& rROI)
 	return TRUE;
 }
 
-// STUB: LEGO1 0x100ab270
+// FUNCTION: LEGO1 0x100ab270
 double Lego3DView::Render(double p_und)
```
Ramen2X, line 142: Is there any reason this function isn't using MxDouble?

Ramen2X (approved):
LGTM

LEGO1/lego/sources/3dmanager/lego3dview.cpp
```diff
@@ -138,15 +138,12 @@ BOOL Lego3DView::Moved(ViewROI& rROI)
 	return TRUE;
 }
 
-// STUB: LEGO1 0x100ab270
-double Lego3DView::Render(double p_und)
+// FUNCTION: LEGO1 0x100ab270
+MxDouble Lego3DView::Render(MxDouble p_und)
```
foxtacles, line 142: The `mxtypes` should only be used within Omni and LegoOmni. For all other modules, please use the intrinsic types.

We should also improve the build system so that these sub modules are compiled separately, and I believe then CMake also has the option to restrict the include paths so that it will no longer be possible to pull in Omni/LegoOmni/etc. dependencies into these sources. Maybe @madebr could look into this?

madebr: `LEGO1/realtime/orientableroi.h` depends on `MxMatrix`.
Is this inentional?

madebr: Never mind, I did not see `MxMatrix` was declared in `mxgeometry.h`.

foxtacles: Yes. There are some header dependencies between the modules outside Omni/LegoOmni; for instance, `viewmanager` also depends on `realtime`. Those dependencies can also be seen in the 1996 source.

The only hard separation seems to be between those and Omni/LegoOmni. In other words:

LegoOmni can have access to everything;
Omni can access the top level modules only;
The modules in `lego/sources` have access to the top level modules only;
The top level modules can have access to each other only (exact relationships not yet clear)

PR #471, opened 2024-01-20
Implement MxDisplaySurface::CreateCursorSurface

MishaProductions:
Matches 56.27% currently, the logic is the same, however couldn't figure out how to get stack to be ordered like the original. Also renamed some members in LegoVideoManager


foxtacles:
There were a few bugs, after fixing those and restructuring it's at 100% now

PR #479, opened 2024-01-21
Roadmap tool to compare binary structure

disinvite:
Our `reccmp` script provides the micro-level comparison of each symbol between original and recomp. We would also like to have the macro perspective to know:

1. What things are in each file?
2. Where are they?
3. How far off are those items (in the recomp) from where they should be?

To do that, I am introducing the `roadmap` script. This takes the same (basic) arguments as `reccmp` but only looks at the position of the elements rather than doing a comparison.

The default text output will show the original address (in `section:offset` format) of each item, followed by the number of bytes ahead or behind this same item can be found in the recomp. This displacement value is computed based on the section where the item lives, so we can still make a useful comparison with `LEGO1.dll` where the files are not yet of comparable size.

After that, we have a description of what symbol type this is (if it is known), the size in bytes (calculated from all available information) and the name (if we have one).

For example, from `ISLE`:
```
0001:000074e0        -16  fun  189     __except_handler3
0001:000075c0        -16  fun  103     _sprintf
0001:00007630        -16  fun  31      _abort
0001:00008110        -16  fun  96      __mtinit
0001:00008190        -16  fun  111     __getptd
0001:0000d5d0        -16  fun  6       _RtlUnwind@16
0002:00000018         64  vta  4       MxParam
0002:000000a0         64  dat  18      _szLibName
0002:00000140         12  str  15      'runtime error '
0002:00000154         12  str  14      'TLOSS error\r\n'
0002:00000164         12  str  13      'SING error\r\n'
0002:00000174         12  str  15      'DOMAIN error\r\n'
```

Adding `--verbose` will also show the recomp address, along with some section information before showing the results. That looks like this:

```
0001:00000000   0001:00000000          0  fun  411     IsleApp::IsleApp
0001:000001a0   0001:000001a0          0  fun  180     IsleApp::~IsleApp
0001:00000260   0001:00000260          0  fun  333     IsleApp::Close
0001:000003b0   0001:000003b0          0  fun  252     IsleApp::SetupLegoOmni
0001:000004b0   0001:000004b0          0  fun  124     MxOmniCreateParam::~MxOmniCreateParam
0001:00000530   0001:00000530          0  fun  7       MxParam::~MxParam
0001:00000540   0001:00000540          0  fun  31      MxParam::`scalar deleting destructor'
```

Verbose output also shows all the items found only in the recomp. These are either items that need an annotation, or (in the `LEGO1.dll` case) they are extraneous symbols that would be removed by adding `/OPT:REF` (see #476)

My thought with having two different text output modes is this: the recomp addresses are always changing, so if we wanted to have a CI action that compares the output for each pull request, it would be better to omit the recomp address to get a reasonable diff.

You can also add `--csv` to output the results as (you guessed it) a CSV. Some initial testing with Google Sheets showed that the section:offset strings can be misinterpreted as a time value. I would expect the same from Excel, but I don't have that here to test. If anyone finds this output useful (@tahg maybe?) then let me know if we should change this output to make the spreadsheet tools play nicely.

The CSV includes the module name for the given item. This would be too much to display on the terminal output, so it's CSV only for now. This should help show where things need to be rearranged in the CMakeLists.txt file. For example, I've found that swapping the order of `isleapp.cpp` and `define.cpp` puts the global variables of `ISLE` in a better spot, so I've added that with this PR.

HTML output coming in a later update.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -299,8 +299,8 @@ set_property(TARGET lego1 PROPERTY SUFFIX ".DLL")
 if (ISLE_BUILD_APP)
   add_executable(isle WIN32
     ISLE/res/isle.rc
-    ISLE/isleapp.cpp
     ISLE/define.cpp
+    ISLE/isleapp.cpp
   )
 
   target_compile_definitions(isle PRIVATE ISLE_APP)
--- a/tools/isledecomp/isledecomp/bin.py
+++ b/tools/isledecomp/isledecomp/bin.py
@@ -1,6 +1,6 @@
 import logging
 import struct
-from typing import List, Optional
+from typing import List, Optional, Tuple
 from dataclasses import dataclass
 from collections import namedtuple
 
@@ -364,6 +364,14 @@ class Bin:
         """Convenience function for converting section:offset pairs from cvdump
         into an absolute vaddr."""
         return self.get_section_offset_by_index(section) + offset
+
+    def get_relative_addr(self, addr: int) -> Tuple[int, int]:
+        """Convert an absolute address back into a (section, offset) pair."""
+        for i, section in enumerate(self.sections):
+            if section.contains_vaddr(addr):
+                return (i + 1, addr - section.virtual_address)
+
+        return (0, 0)
 
     def get_raw_addr(self, vaddr: int) -> int:
         """Returns the raw offset in the PE binary for the given virtual address."""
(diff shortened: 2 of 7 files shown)


foxtacles:
Very useful! I think a potential next step for this might be to automatically suggest in which order we have to compile the units to improve the accuracy of the positioning. Maybe based on the first address in each class/compilation unit it could output something that we could insert into `CMakeLists`.

PR #480, opened 2024-01-22
LegoWorld STL members

disinvite:
`LegoWorld::LegoWorld` is now at 100% match with the addition of 3 STL members and filling in some annotations.

The two `<set>` members appear to contain pointers to `MxPresenter`. The key comparison is just looking at the pointer values, but I suppose if the goal is to have a unique list of presenters then that's valid.

The `<list>` at +0xe0 contains pointers to `AutoROI`, confirmed via the debugger. The only vtable I saw for the list elements matched what we have for `AutoROI`. It could be a generic `ViewROI*` list, too. I didn't want to insert the dependency in between `LegoROI` and `ViewROI` right now so it's just mocked up.

The `AutoROI` object comes from `LegoModelPresenter`. The function at `0x1007ff70` adds to the `+0xe0` list near the instruction `0x10080031`, where we take the `AutoROI` from `LegoModelPresenter +0x64`.

`0x100801b0` (aka `LegoModelPresenter::ParseExtra`) removes from the list by decrementing `LegoWorld +0xe8`. That function is also responsible for setting `LegoModelPresenter +0x64` from a value returned by `LegoUnkSaveDataWriter::FUN_10083500`.

I added this annotation for `Mx3DPointFloat`:

```
// SYNTHETIC: LEGO1 0x1001d170
// Mx3DPointFloat::Mx3DPointFloat
```

Is that truly a synthetic constructor or just MSVC not taking the `inline` suggestion for the constructor we do have?

### DIFF
--- a/LEGO1/lego/legoomni/include/legomodelpresenter.h
+++ b/LEGO1/lego/legoomni/include/legomodelpresenter.h
@@ -3,6 +3,8 @@
 
 #include "mxvideopresenter.h"
 
+class AutoROI;
+
 // VTABLE: LEGO1 0x100d4e50
 // SIZE 0x6c (discovered through inline constructor at 0x10009ae6)
 class LegoModelPresenter : public MxVideoPresenter {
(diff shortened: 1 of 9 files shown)


foxtacles:
Looks good! This should allow us to start implementing more functions in this class, which seems to be kind of a centerpiece in the game. There's several (large) functions that touch upon seemingly all of those lists/sets at the same time so this will help a lot.

PR #481, opened 2024-01-22
Implement FLIC

tahg:
FLIC library implemented, but this still needs some decisions made if we are to merge it.
I arbitrarily put flic.cpp in the root LEGO1 directory, just so the tools would find it. It probably belongs in 3rd party, but we don't compile those directories currently.
Matching is decent (8/16 at 100%), and better than the percents on progress might suggest.  Mostly it's register swaps, but if anyone wants to try their hand on matching it further, that would be appreciated.
I think naming of everything is decent, but any thoughts on that are appreciated as well.


foxtacles:
I've merged in master and performed some overall cleanup on the code. Since the code is located in the middle of `omni`, we should not consider it an extra library. I've moved it into `omni`.

The current implementation is not working; it doesn't correctly decode the book animation and crashes the game a few seconds after.

If somebody wants to test this, make sure to disable reading of animations here: https://github.com/isledecomp/isle/blob/master/LEGO1/lego/legoomni/src/video/legoanimpresenter.cpp#L92 - then the game should boot into the Infocenter just fine.

tahg:
All animations should be working correctly now.

PR #482, opened 2024-01-23
Implement MxImage/MxColor

tahg:


foxtacles:
Those classes are not part of `omni`; they are part of the utility/misc classes located in `lego/sources`. This can be easily confirmed by the 1996 source, which also has a full copy of them:

image

Since original sources exist, it will be preferable copy-pasting and preserving them (the same way we did with other classes) instead of rolling our own version of them.

tahg:
I suspected they might be Lego classes, but I didn't know we had source for them. I'll see what I can do to arrange the hierarchy, and double check no changes were made from the source we have.

PR #484, opened 2024-01-24
Spinoff some sources to static libraries

madebr:


CMakeLists.txt
```diff
@@ -235,61 +317,31 @@ add_library(lego1 SHARED
   LEGO1/omni/src/video/mxvideoparam.cpp
   LEGO1/omni/src/video/mxvideoparamflags.cpp
   LEGO1/omni/src/video/mxvideopresenter.cpp
-  LEGO1/realtime/matrix.cpp
-  LEGO1/realtime/orientableroi.cpp
-  LEGO1/realtime/realtime.cpp
   LEGO1/realtime/realtimeview.cpp
```
madebr, line 320: `link.exe` does only export symbols from objects, not from static libraries.
Because realtimeview.cpp exports some symbols, it cannot be added to a realtime static library.
So perhaps this source belongs in LegoOmni instead?

foxtacles: Maybe I misunderstand, but doesn't it seem to work? Otherwise there should be an error in the verexp tool regarding a missing symbol?

madebr: In 76debb7e5b9c9ce177acde36849c415291ce27d6, I moved it back from realtime to lego01.

foxtacles:
Would it also be possible to do the same for Omni and LegoOmni?

CMakeLists.txt
foxtacles: ok, let me see what the best course of action is here. The 1996 has it as its own library, but maybe they changed it

foxtacles: Hmm, if there really is no other way to make this work we'll have to leave it like that I guess. The odd thing is that the functions in `realtimeview.cpp` appear in between several other static libs, so if this was really part of LegoOmni then everything that comes after it would also have to have been compiled as part of LegoOmni, which seems unlikely.

foxtacles: I'm assuming using a .DEF file wouldn't make a difference?
https://learn.microsoft.com/en-us/cpp/build/reference/module-definition-dot-def-files?view=msvc-170

madebr: A `.def` could work, but that will only work for MSVC.
mingw mangles the c++ symbols differently.

madebr: It it possible to build omni and legoomni as static libraries, but at the end, only their objects will be used (not the .lib)

foxtacles: Based on the 1996 sources I think only Omni was built as a library, and then linked at the very end.

Roughly speaking, we are seeing the code in this order:

1) LegoOmni
2) MSVC libraries
3) Top-level libraries and lego/sources libraries
4) Omni

maybe @tahg could give us a default accounting of especially the libraries in the middle since he has a spreadsheet with everything

madebr: Makes sense, if and only if the omni libraries don't export. Otherwise they would had to use some kind of `.def`.

foxtacles: Considering I can't find a single `dllexport` in the 1996 sources I have a feeling they probably did use those .def files instead. Omni does export a lot of functions, so I can't think of any other way

madebr: It makes also sense because the number of exported symbols is kept to the very minimum.
They were very careful to not export too many symbols.

tahg: We don't have source for anything that was in a DLL as far as I know.  If you look at the LEGOAPP folder, you can see that quite a lot more was in the main exe.

madebr: Because mingw's thiscall and msvc's thiscall convention are incompatible, msvc and mingw have their own `.def` file.
I removed the `COMPAT_CONST` version of `MxVideoParam::MxVideoParam` so current MSVC can keep using the same def as msvc4.

Can you reformat?

foxtacles: Ran the format.

I've found a few more "complete" link instructions from the 1996 sources:

`LEGO/LEGOOMNI/SRC/DLL/LEGOOMNI.MAK:# ADD LINK32 omni2dSH.lib libcmt.lib winmm.lib dsound.lib kernel32.lib user32.lib gdi32.lib winspool.lib comdlg32.lib advapi32.lib shell32.lib ole32.lib oleaut32.lib uuid.lib odbc32.lib odbccp32.lib miscd.lib roid.lib animd.lib geomd.lib tglrl40d.lib d3drm.lib 3dmanagerd.lib realtimed.lib viewmanagerd.lib pcubed.lib oldnames.lib /nologo /subsystem:windows /dll /debug /machine:I386 /nodefaultlib /out:"..\..\bin\LegoOmni1dSH.dll" /implib:"..\..\lib\LegoOmni1dSH.lib"`

and

`LEGO/LEGOOMNI/SRC/DLL/LEGOOMNI.MAK:# ADD BASE LINK32 libcmt.lib winmm.lib omni2d.lib dsound.lib kernel32.lib user32.lib gdi32.lib winspool.lib comdlg32.lib advapi32.lib shell32.lib ole32.lib oleaut32.lib uuid.lib odbc32.lib odbccp32.lib miscd.lib roid.lib animd.lib geomd.lib tglrl20d.lib rlddf.lib rlddfglu.lib rlff.lib 3dmanagerd.lib realtimed.lib viewmanagerd.lib pcubed.lib /nologo /subsystem:windows /dll /debug /machine:I386 /nodefaultlib /out:"..\..\bin\LegoOmni1d.dll" /implib:"..\..\lib\LegoOmni1d.lib"`

It would seem that they generate a separate file with `SH` appended for the SmartHeap version, which is a detail we could cover but not too important.

We stilll need to add yet another library called `3dmanager` based on `lego/sources/3dmanager`. At that point we should have all of it (for now).

Moving `realtimeview.cpp` to the the realtime lib seems to work now with the `.DEF` files.

foxtacles: Also found the reference to the `.DEF` they used:

`LEGO/LEGOOMNI/SRC/DLL/LEGOOMNI.DEF:; LegoOmni.def : Declares the module parameters for the OMNI.DLL.`

madebr: Let's ignore the `SH` for now, because that would also mean those libraries are built without smartheap's `operator new`.
For example, does that mean 3dmanager is always built without smartheap?
Mixing allocators is tricky.

LEGO1/LegoOmni.mingw.def
```diff
@@ -0,0 +1,172 @@
+; LegoOmni.def : Declares the module paarameters for the LEGO1DLL.
+
+DESCRIPTION  " Lego OMNI Windows Dynamic Link Library"
+
+EXPORTS
+
+; EXPORTs really required only.
+
+_ZN19LegoBackgroundColorC1EPKcS1_@8 = _ZN19LegoBackgroundColorC1EPKcS1_
```
madebr, line 9: I generated this file using mingw64's gendef script, and then fixed up these stdcall functions.
I think this list can be pruned to (approximately) the same size as the msvc one, but that's a job for another time :)

CMakeLists.txt
madebr: The various `__declspec(dllexport)`'s can also be removed now.

CMakeLists.txt
```diff
@@ -16,12 +16,209 @@ if (MSVC)
...
+target_include_directories(DirectX5::DirectX5 INTERFACE "${CMAKE_SOURCE_DIR}/3rdparty/dx5/inc")
+target_link_directories(DirectX5::DirectX5 INTERFACE "${CMAKE_SOURCE_DIR}/3rdparty/dx5/lib")
+
+add_library(Smacker::Smacker STATIC IMPORTED)
+set_property(TARGET Smacker::Smacker PROPERTY IMPORTED_LOCATION "${CMAKE_SOURCE_DIR}/3rdparty/smacker/smack.lib")
+set_property(TARGET Smacker::Smacker PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_SOURCE_DIR}/3rdparty/smacker")
+
+add_library(FLIC::FLIC STATIC IMPORTED)
+set_property(TARGET FLIC::FLIC PROPERTY IMPORTED_LOCATION "${CMAKE_SOURCE_DIR}/3rdparty/flic/flic.lib")
+set_property(TARGET FLIC::FLIC PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_SOURCE_DIR}/3rdparty/flic")
```
madebr, line 40: For #481, just remove all references to `FLIC::FLIC`, and add the flic sources where it is appropriate.

CMakeLists.txt
```diff
@@ -16,12 +16,209 @@ if (MSVC)
...
+target_link_libraries(viewmanager PRIVATE Vec::Vec)
+
+add_library(mxdirectx STATIC
+  LEGO1/mxdirectx/mxdirect3d.cpp
+  LEGO1/mxdirectx/mxdirectdraw.cpp
+)
+register_lego1_target(mxdirectx)
+set_property(TARGET mxdirectx PROPERTY ARCHIVE_OUTPUT_NAME "MxDirectX$<$<CONFIG:Debug>:d>")
+target_include_directories(mxdirectx PRIVATE "${CMAKE_SOURCE_DIR}/LEGO1" "${CMAKE_SOURCE_DIR}/util")
+target_link_libraries(mxdirectx PRIVATE ddraw)
```
madebr, line 95: I added ddraw here because `DirectDrawCreate` is called inside the mxdirectx sources,
but if you want to have more control in the order you can remove them here and add them all to lego1.

CMakeLists.txt
foxtacles: I've removed them, but this caused the msys build to break it seems. Otherwise this looks pretty good now.

CMakeLists.txt
```diff
@@ -47,14 +47,6 @@ set_property(TARGET SmartHeap::SmartHeap PROPERTY IMPORTED_LOCATION "${CMAKE_SOU
 set_property(TARGET SmartHeap::SmartHeap PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_SOURCE_DIR}/3rdparty/smartheap")
 set_property(TARGET SmartHeap::SmartHeap PROPERTY INTERFACE_COMPILE_OPTIONS "/FI${CMAKE_SOURCE_DIR}/3rdparty/smartheap/SMRTHEAP.HPP")
 
-add_library(mxgeometry STATIC
-  LEGO1/mxgeometry/mxgeometry3d.cpp
-  LEGO1/mxgeometry/mxmatrix.cpp
```
madebr: Don't forget to add these to somewhere else.

foxtacles: These are actually obsolete and served only the purpose of having the `DECOMP_SIZE_ASSERT`s. At some point we should move away from this construct and verify the sizes through the PDB so we don't have this issue with header-only libraries. For now I moved the asserts to an unrelated compilation unit.

CMakeLists.txt
```diff
@@ -143,155 +332,37 @@ add_library(lego1 SHARED
...
-target_link_directories(lego1 PRIVATE "${CMAKE_SOURCE_DIR}/3rdparty/flic")
+target_include_directories(lego1 PUBLIC "${CMAKE_SOURCE_DIR}/util")
+target_include_directories(lego1 PUBLIC "${CMAKE_SOURCE_DIR}/LEGO1")
+target_include_directories(lego1 PUBLIC "${CMAKE_SOURCE_DIR}/LEGO1/omni/include")
+target_include_directories(lego1 PUBLIC "${CMAKE_SOURCE_DIR}/LEGO1/lego/sources")
+target_include_directories(lego1 PUBLIC "${CMAKE_SOURCE_DIR}/LEGO1/lego/legoomni/include")
 
 # Link libraries
-target_link_libraries(lego1 PRIVATE ddraw dsound dinput dxguid winmm d3drm smack flic)
+target_link_libraries(lego1 PRIVATE tglrl viewmanager realtime mxdirectx roi FLIC::FLIC Vec::Vec dinput dxguid Smacker::Smacker misc 3dmanager omni)
```
foxtacles, line 356: Do we still need to specify FLIC and Smacker here since they are already linked into Omni?

madebr: LEGO1/omni/include/mxflcpresenter.h includes `flic.h`. the smacker can go.

foxtacles:
Excellent, looks good. We'll add more of those libraries over time. Ordering everything can be done later as well.

PR #505, opened 2024-01-29
Finish MxControlPresenter

MishaProductions:
Using MxBitmap::GetStart() caused an access violation in MxControlPresenter::FUN_10044270 for some reason.


foxtacles:
> Using MxBitmap::GetStart() caused an access violation in MxControlPresenter::FUN_10044270 for some reason.

Seems to work for me, or am I missing something - did you try this?

`start = bitmap->GetStart(p_x - first->GetLocation().GetX(), p_y - first->GetLocation().GetY());`

MishaProductions:
Thanks, that works! I used bitmap->GetStart(p_x, p_y) which caused that problem

PR #507, opened 2024-01-29
roadmap: Suggest order of modules

disinvite:
This adds a new command line option `--order` to the `roadmap` tool that will suggest the order of modules (i.e. translation units). This is broken out by build target, so for `LEGO1` each library will show its own list of modules.

By default, we are looking for functions from each module and where they first appear in the binary. We also show the average displacement for each function from that module (i.e. how far off it is in the recomp from where it should be.) `Score::Score` is the first function from `LEGO1` so naturally the `score.cpp` module appears first in the list. You can provide an argument to `--order` to pick which symbol type you want to look at. For example: `--order vtable` will look at only vtables.

*Caution:* This will _only_ show modules that are currently used in the final binary. You should not copy the list from `roadmap` directly into `CMakeLists.txt` because this will omit any modules that are not referenced yet. Some modules are marked "no suggestion", which means that it is being used, but provides no symbols matching the type that you chose. To give an example, `LEGO1/realtime/matrix.cpp` is included in the list, but no symbols from the .obj file are used in `LEGO1` at the moment.

If we think it's worth applying the results directly to `CMakeLists.txt` now, we can do that. I think combining the results from checking functions _and_ vtables would be a good starting point.

### DIFF
--- a/tools/roadmap/roadmap.py
+++ b/tools/roadmap/roadmap.py
@@ -56,10 +56,17 @@ class ModuleMap:
 
     def get_module(self, addr: int) -> Optional[str]:
         i = bisect.bisect_left(self.contrib_starts, addr)
+        # If the addr matches the section contribution start, we are in the
+        # right spot. Otherwise, we need to subtract one here.
+        # We don't want the insertion point given by bisect, but the
+        # section contribution that contains the address.
 
-        # Clamp to final section contribution for a very high address
-        if i >= len(self.section_contrib):
-            i = len(self.section_contrib) - 1
+        (potential_start, _, __) = self.section_contrib[i]
+        if potential_start != addr:
+            i -= 1
+
+        # Safety catch: clamp to range of indices from section_contrib.
+        i = max(0, min(i, len(self.section_contrib) - 1))
 
         (start, size, module_id) = self.section_contrib[i]
         if start <= addr < start + size:
--- a/tools/roadmap/roadmap.py
+++ b/tools/roadmap/roadmap.py
@@ -29,6 +29,7 @@ class ModuleMap:
     def __init__(self, pdb, binfile) -> None:
         cvdump = Cvdump(pdb).section_contributions().modules().run()
         self.module_lookup = {m.id: (m.lib, m.obj) for m in cvdump.modules}
+        self.library_lookup = {m.obj: m.lib for m in cvdump.modules}
         self.section_contrib = [
             (
                 binfile.get_abs_addr(sizeref.section, sizeref.offset),
@@ -38,6 +39,9 @@ class ModuleMap:
             for sizeref in cvdump.sizerefs
             if binfile.is_valid_section(sizeref.section)
         ]
+
+    def get_lib_for_module(self, module: str) -> Optional[str]:
+        return self.library_lookup.get(module)
 
     def get_all_cmake_modules(self) -> List[str]:
         return [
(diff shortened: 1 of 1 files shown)


foxtacles:
Looks good! Would it be possible to sort the modules by their first address as well?

foxtacles:
Also, I'm not exactly sure how why we have this case for instance:

image

`0x10005530` is `MxDSObject::SetAtomId`. I'm not sure why it reports it as part of `MxDSAction`, but setting this aside, the eventual goal is to know in which order we have to compile the units. Inline functions should be disregarded here, because functions like `MxDSObject::SetAtomId` end up in the "LegoOmni" space but that does not mean `mxdsaction.cpp` for instance should be compiled in there.

Is there a way to possibly exclude functions that were declared inline originally?

disinvite:
But it's not inline, right? It's a member of the vtable for `MxDSObject`. I think it's getting considered as part of `MxDSAction` because that vtable comes before `MxDSObject` or any of its children:

```
0x100dc098 - MxDSAction
0x100dc868 - MxDSObject
0x100dccf8 - MxDSObjectAction
0x100dcd40 - MxDSMediaAction
0x100dcd88 - MxDSAnim
0x100dcdd0 - MxDSSound
0x100dce18 - MxDSEvent
0x100dce60 - MxDSStill
0x100dcef0 - MxDSMultiAction
0x100dcf38 - MxDSSerialAction
0x100dcf80 - MxDSParallelAction
0x100dcfc8 - MxDSSelectAction
0x100dd088 - MxDSStreamingAction
```

That doesn't explain why it appears so early in `LEGO1` because none of the other vtable functions are pulled out of sequence.

If a function is only used inline, it would not be registered in any of the places where we're looking for symbols/functions in cvdump.

disinvite:
The order seems better if we take the average with outliers removed instead of just picking the earliest. See what you think.

Still working on suggested library order, I'll pick that up tomorrow.

foxtacles:
> The order seems better if we take the average with outliers removed instead of just picking the earliest. See what you think.

Yeah looks better now! A bit confusing that there is an outlier like `mxdsaction.cpp` in between now:

image

Maybe if that can't be changed/fixed it should be highlighted/marked somehow. The spot of `mxdsaction.cpp` seems right though.

foxtacles:
> It's still showing the earliest address for that module, but they are ordered based on removing the outliers. Do we show a different address there instead? The address range for the module? Mark the out-of-order address with an asterisk?

I think something simple as an asterisk should do just fine

foxtacles:
Perhaps the library order can also be based on an average? Or maybe the earliest address that doesn't have an asterisk. Right now it shows `omni.lib` as the 2nd although it is definitely one of the last:

image

disinvite:
Let me check on that. I had thought the reason would be that there are no public functions to grab in the pdb/cvdump output, but our markers in `library_smack.h` should provide the match.

We need the markers for that because the cvdump analysis code doesn't make an assumption about kind of symbol it is based on its section. For example: we could assume that symbols from `.text` are all functions. The smacker functions are (currently) in their own section `_UNSTEXT`, but we should still be able to look at the flags to say whether this is a section for code.

If we change that approach (by making an assumption on symbol type) then it could lead to better information. I just didn't know if we could definitively say that everything in `.text` is a function. Would any of our C libraries (if they had .asm code units) put data there? I've seen that in my own decomp projects with the old Borland compilers for DOS for example.

disinvite:
Got it. It was a bug introduced by my using the `bisect` python library to speed up the `get_module` method. This affected `smack.lib` because it has no public symbols, so we are taking addresses from the function markers and checking against the single entry from cvdump `SECTION CONTRIBUTIONS` that represents the entire module. The same problem would occur if the address of a marker is wrong, which may be the case for `g_colorSaveData`.

foxtacles:
Great, I think this is a great starting point to get the compilation units in order, thanks!

PR #510, opened 2024-01-31
Implement various infocenter functions

MishaProductions:


foxtacles:
Please try to keep future PRs a bit smaller, especially when there are multiple functions with less than 100% accuracy - thank you!

PR #520, opened 2024-02-02
Implement some Act1State functions

madebr:
Also test automatic clang-format reformatting

### DIFF
--- a/LEGO1/lego/legoomni/include/act1state.h
+++ b/LEGO1/lego/legoomni/include/act1state.h
@@ -54,6 +54,7 @@ public:
 
 		inline void SetName(const char* p_name) { m_name = p_name; }
 		inline const MxString* GetName() const { return &m_name; }
+
 		// FUNCTION: LEGO1 0x100344d0
 		MxS32 Serialize(LegoFile* p_file)
 		{
@@ -69,6 +70,7 @@ public:
 				p_file->ReadVector3(&m_point2);
 				p_file->ReadVector3(&m_point3);
 			}
+
 			return 0;
 		}
 
--- a/LEGO1/lego/legoomni/include/act1state.h
+++ b/LEGO1/lego/legoomni/include/act1state.h
@@ -48,7 +48,7 @@ public:
 		// FUNCTION: LEGO1 0x10033800
 		NamedPlane() {}
 
-		inline void SetName(const char* name) { m_name = name; }
+		inline void SetName(const char* p_name) { m_name = p_name; }
 		inline const MxString* GetName() const { return &m_name; }
 		int Serialize(LegoFile* p_file);
 
--- a/LEGO1/lego/legoomni/include/act1state.h
+++ b/LEGO1/lego/legoomni/include/act1state.h
@@ -56,7 +56,7 @@ public:
 		inline const MxString* GetName() const { return &m_name; }
 
 		// FUNCTION: LEGO1 0x100344d0
-		MxS32 Serialize(LegoFile* p_file)
+		MxResult Serialize(LegoFile* p_file)
 		{
 			if (p_file->IsWriteMode()) {
 				p_file->FUN_10006030(m_name);
@@ -71,7 +71,7 @@ public:
 				p_file->ReadVector3(m_point3);
 			}
 
-			return 0;
+			return SUCCESS;
 		}
 
 	private:
--- a/LEGO1/lego/legoomni/include/act1state.h
+++ b/LEGO1/lego/legoomni/include/act1state.h
@@ -3,6 +3,11 @@
 
 #include "legostate.h"
 
+#include "misc/legotexture.h"
+#include "roi/legoroi.h"
+
+class LegoTexture;
+
 // VTABLE: LEGO1 0x100d7028
 // SIZE 0x26c
 class Act1State : public LegoState {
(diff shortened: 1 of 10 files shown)


LEGO1/lego/legoomni/src/act1/act1state.cpp
```diff
@@ -1,34 +1,295 @@
 #include "act1state.h"
 
+DECOMP_SIZE_ASSERT(Act1State, 0x26c)
+DECOMP_SIZE_ASSERT(Act1State::NamedPlane, 0x4c)
+DECOMP_SIZE_ASSERT(Act1State::NamedTexture, 0x14)
+
+#define isle_asize(ARR) (sizeof(ARR) / sizeof((ARR)[0]))
```
madebr, line 7: This macro would probably be useful in other locations (and should be upper cased)

foxtacles: there is one here: ttps://github.com/isledecomp/isle/blob/master/util/decomp.h#L17 We can change the name of it

LEGO1/lego/legoomni/include/act1state.h
```diff
@@ -25,29 +29,85 @@ class Act1State : public LegoState {
...
 
+	// SIZE 0x4c
+	class NamedPlane {
+	public:
+		// FUNCTION: LEGO1 0x10033800
+		NamedPlane() {}
+
+		inline void SetName(const char* p_name) { m_name = p_name; }
+		inline const MxString* GetName() const { return &m_name; }
+		int Serialize(LegoFile* p_file);
```
foxtacles, line 53: Instead of `int` it should be `MxS32`. there's a few more places

LEGO1/lego/legoomni/src/act1/act1state.cpp
madebr: Thanks, I'll re-use `_countof`. But added some parentheses around it.

LEGO1/lego/legoomni/src/act1/act1state.cpp
```diff
@@ -1,34 +1,295 @@
...
+
+// Forward declarations
+Act1State::NamedTexture* ReadNamedTexture(LegoFile* p_file);
+void WriteNamedTexture(LegoFile* p_file, Act1State::NamedTexture* p_texture);
+void FUN_1003f540(LegoFile* p_file, const char* p_filename);
+
+enum {
```
foxtacles, line 12: We generally declare enums inside the classes

madebr: I didn't because it was not part of the public interface,  but I'll your suggestion.

LEGO1/lego/legoomni/include/act1state.h
madebr: Those are goners now.

LEGO1/lego/legoomni/include/act1state.h
```diff
@@ -25,29 +35,88 @@ class Act1State : public LegoState {
...
+	undefined m_unk0x022;     // 0x022
+	undefined m_unk0x023;     // 0x023
+	NamedPlane m_unk0x024;    // 0x024
+	NamedPlane m_unk0x070;    // 0x070
+	NamedPlane m_unk0x0bc;    // 0x0bc
+	NamedPlane m_unk0x108;    // 0x108
+	NamedTexture* m_unk0x154; // 0x154
+	NamedTexture* m_unk0x158; // 0x158
+	NamedTexture* m_unk0x15c; // 0x15c
+	MxCore* m_unk0x160;       // 0x160
```
madebr, line 104: This should obvious be a subclass of `MxCore`, but I guess `MxCore` will do for now?

LEGO1/lego/legoomni/src/act1/act1state.cpp
```diff
@@ -1,34 +1,287 @@
...
 	// TODO
 	return SUCCESS;
 }
 
-// STUB: LEGO1 0x100346d0
+// FUNCTION: LEGO1 0x100344d0
+MxS32 Act1State::NamedPlane::Serialize(LegoFile* p_file)
+{
+	if (p_file->IsWriteMode()) {
+		p_file->FUN_10006030(m_name);
```
madebr, line 185: This method should be named `WriteString` btw.

LEGO1/lego/legoomni/src/act1/act1state.cpp
```diff
@@ -1,34 +1,287 @@
 #include "act1state.h"
 
+DECOMP_SIZE_ASSERT(Act1State, 0x26c)
+DECOMP_SIZE_ASSERT(Act1State::NamedPlane, 0x4c)
+DECOMP_SIZE_ASSERT(Act1State::NamedTexture, 0x14)
+
+// Forward declarations
+Act1State::NamedTexture* ReadNamedTexture(LegoFile* p_file);
```
foxtacles, line 8: Those three functions are very likely part of the `LegoUtil` collection. You can tell by their address.

LEGO1/lego/legoomni/src/act1/act1state.cpp
```diff
@@ -1,34 +1,287 @@
...
 }
 
-// STUB: LEGO1 0x10033ac0
+// FUNCTION: LEGO1 0x10033ac0
 MxResult Act1State::VTable0x1c(LegoFile* p_legoFile)
 {
+	if (p_legoFile->IsWriteMode()) {
+		p_legoFile->FUN_10006030(ClassName());
+	}
+	m_unk0x024.Serialize(p_legoFile);
```
foxtacles, line 56: This function appears to be inlined here at least a few times - so the `Serialize` function should be implemented in the header or at least declared inline (although we generally prefer having the implementation located in the header)

LEGO1/lego/sources/misc/legostorage.h
```diff
@@ -72,6 +73,37 @@ class LegoFile : public LegoStorage {
...
+	LegoStorage* ReadVector3(Vector3* p_vec3)
+	{
+		Read(&(*p_vec3)[0], sizeof(float));
+		Read(&(*p_vec3)[1], sizeof(float));
+		Read(&(*p_vec3)[2], sizeof(float));
+		return this;
+	}
+
+	// FUNCTION: LEGO1 0x100343d0
+	LegoStorage* WriteVector3(Vector3 p_vec3)
```
foxtacles, line 86: This takes a `Mx3DPointFloat` as a parameter. `100343a0` is the constructor called in `Act1State::NamedPlane::Serialize` before it is passed to `WriteVector3`. In the constructor you can tell by the virtual table address that is being assigned among other things. The generated constructor should also be annotated (see the bottom of `act1state.h` for an example)

madebr: Yup, you're right.

PR #527, opened 2024-02-04
parser: Performance enhancements

disinvite:
Been going down the performance rabbit hole for our python tooling. Here are a few quick things that should help a lot. No change to match percentage or `reccmp` output.

* In `isledecomp.bin`, read the entire binary file at once and use python `memoryview` instead of buffer file I/O. Changed some of the API methods to allow access to the individual sections if any scripts want them.
* Minor change to `roadmap` script now that we are decoding the ASCII version of the section name instead of providing the raw 8 bytes from the PE header.
* Added unit test for `isledecomp.bin`. This pulls down the original `LEGO1.DLL` so we can test the output from that module. I added this to the github CI.
* Memoize expensive `pathlib` calls in `isledecomp.compare.lines`. This is the module that relates the filename/line pairs from `cvdump` with the decomp annotations.
* Added a SQL index on the "name" column in our main comparison DB: `isledecomp.compare.db`.
* In `isledecomp.compare.asm.parse`, memoize the "from_hex" operation which is not particularly expensive but gets called _a lot_. You can imagine that many of these calls are passing in register names that are obviously not hex numbers. Added some code to avoid the expensive regex operations if there is nothing for us to sanitize. This section probably needs a refactor later but I will roll that into the upcoming jump table enhancement.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,6 +110,7 @@ jobs:
       id: cache-original-binaries
       uses: actions/cache/restore@v3
       with:
+        enableCrossOsArchive: true
         path: legobin
         key: legobin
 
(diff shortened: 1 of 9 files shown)


foxtacles:
Looks good, it's noticeably faster now!

PR #528, opened 2024-02-04
Starting removing `this->` from `LEGO1/lego/legoomni/`

danielhejduk:
This PR does:
- Removing all unnecessary `this->` keywords from folder `LEGO1/lego/legoomni/audio/`
- Removing all unnecessary `this->` keywords from folder `LEGO1/lego/legoomni/build/`


danielhejduk:
It seems that I need to learn more stuff about C++ and using it in decomp projects. So I am just cleaning up the existing code from `this->` keywords that are unnecessary.

foxtacles:
Refactoring of this kind should be applied to the entire code base, where applicable, at once - not piecemeal. As with most style choices we also need to consider the maintenance effort required to stay consistent. Generally this means we require a tool that enforces that consistency to avoid burdening maintainers.

PR #530, opened 2024-02-05
Match static function variables

disinvite:
After some analysis with the `roadmap` tool on `ISLE`, I came to the conclusion that two global variables `g_lastFrameTime` and `g_startupDelay` used only in the `IsleApp::Tick` function are actually static variables _from_ that function.

Before:
```
0003:0000005c          0  dat  4       g_targetHeight
0003:00000060          0  dat  4       g_targetDepth
0003:00000064          0  dat  4       g_reqEnableRMDevice
// omitted
0003:0000011c          8  str  19      'LEGO® Island Error'
0003:00000130          8  str  120     '"LEGO® Island" is not detecting...
0003:000001a8          8  str  18      'Lego Island Error'
0003:000001bc       -340  dat  4       g_startupDelay
0003:000001c0       -340  dat  4       g_lastFrameTime
0003:000001c4          0  str  24      'Lego Island MainNoM App'
```

To explain the output from `roadmap`, the items are ordered as they appear in the original, with the address from original shown in section:offset format. The first three global variables listed are exactly where they should be. They have the same offset in the `.data` section in both original and recomp. The strings that follow are ahead in the recomp by 8 bytes, and then we have the two aforementioned variables. These two are quite a ways behind where they should be.

After making those variables static (and reversing their order):
```
0003:0000005c          0  dat  4       g_targetHeight
0003:00000060          0  dat  4       g_targetDepth
0003:00000064          0  dat  4       g_reqEnableRMDevice
// omitted
0003:0000011c          0  str  19      'LEGO® Island Error'
0003:00000130          0  str  120     '"LEGO® Island" is not detecting...
0003:000001a8          0  str  18      'Lego Island Error'
0003:000001bc          0       4       ?g_startupDelay@?1??Tick@IsleApp@@QAEXH@Z@4HA
0003:000001c0          0       4       ?g_lastFrameTime@?1??Tick@IsleApp@@QAEXH@Z@4JA
0003:000001c4          0  str  24      'Lego Island MainNoM App'
```

The offsets line up exactly.

The challenge with matching these static function variables is that they are not public, per se. If the symbols appear in `cvdump` at all, you get the mangled name only (as you see above). My approach to solving this is:

1. The parser can identify a static function variable when it sees a `// GLOBAL` marker. This is based on context, meaning that we don't know whether we are in a function unless the `// FUNCTION` marker has just been parsed. Since our goal is to annotate all functions, while this is a limitation of the parser it is a solvable one. (`libclang` would eliminate this problem.)
2. Since we know which function we are in when we report a static function variable, capture the function's original address too.
3. When it's time to match up this variable, check the database for the function address and find the decorated name for that function. Try to match on the variable based on a combination of the variable name and the decorated function name.

This works well, but it all depends on the variable appearing in `cvdump`. This is not the case for `g_isInsideError` from `MxDirectDraw::Error` and I'm not sure why. Like `IsleApp::Tick`, it is a public function of its class.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -415,8 +415,8 @@ set_property(TARGET lego1 PROPERTY SUFFIX ".DLL")
 if (ISLE_BUILD_APP)
   add_executable(isle WIN32
     ISLE/res/isle.rc
-    ISLE/isleapp.cpp
     ISLE/define.cpp
+    ISLE/isleapp.cpp
   )
 
   target_compile_definitions(isle PRIVATE ISLE_APP)
--- a/ISLE/define.cpp
+++ b/ISLE/define.cpp
@@ -32,9 +32,3 @@ int g_targetDepth = 16;
 
 // GLOBAL: ISLE 0x410064
 int g_reqEnableRMDevice = 0;
-
-// GLOBAL: ISLE 0x4101bc
-int g_startupDelay = 200;
-
-// GLOBAL: ISLE 0x4101c0
-MxLong g_lastFrameTime = 0;
--- a/ISLE/define.h
+++ b/ISLE/define.h
@@ -21,7 +21,5 @@ extern int g_targetWidth;
 extern int g_targetHeight;
 extern int g_targetDepth;
 extern int g_reqEnableRMDevice;
-extern int g_startupDelay;
-extern MxLong g_lastFrameTime;
 
 #endif // DEFINE_H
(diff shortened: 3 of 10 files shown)


foxtacles:
Ah, it's great to see this confirmed this way. I have a local branch with a bunch of alterations to `ISLE`, and among other things, it includes having those static variables in that function - the reason being that there is a code fragment in the 1996 in the `MAIN` client that also placed them there.

PR #533, opened 2024-02-06
Decompiled CONFIG.EXE executable

madebr:
While looking into an issue with `ISLE.EXE` (see chat), I started reversing `CONFIG.EXE`.
It is a small MFC app, that re-uses `LEGO1/mxdirectx/mxdirectx.cpp` and also copies some registry-related functions from `ISLE.EXE`.
It behaves identical to the original one. (Or at least it should)

I have not been able yet to run `reccmp.py` on the resulting executable because it fails with the following exception and traceback:
```
Traceback (most recent call last):
  File "/home/maarten/projects/isle/tools/reccmp/reccmp.py", line 296, in <module>
    raise SystemExit(main())
                     ^^^^^^
  File "/home/maarten/projects/isle/tools/reccmp/reccmp.py", line 214, in main
    with Bin(args.original, find_str=True) as origfile, Bin(
  File "/home/maarten/projects/isle/tools/isledecomp/isledecomp/bin.py", line 165, in __enter__
    self._populate_imports()
  File "/home/maarten/projects/isle/tools/isledecomp/isledecomp/bin.py", line 318, in _populate_imports
    self.imports = list(iter_imports())
                   ^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/projects/isle/tools/isledecomp/isledecomp/bin.py", line 313, in iter_imports
    symbol_name = self.read_string(name_ofs).decode("ascii")
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/projects/isle/tools/isledecomp/isledecomp/bin.py", line 409, in read_string
    b = self.read(offset, chunk_size)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/projects/isle/tools/isledecomp/isledecomp/bin.py", line 422, in read
    (section_id, offset) = self.get_relative_addr(vaddr)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/projects/isle/tools/isledecomp/isledecomp/bin.py", line 387, in get_relative_addr
    raise InvalidVirtualAddressError(hex(addr))
isledecomp.bin.InvalidVirtualAddressError: 0x804014ab
```

I think the executables are close. They differ by 4 kiB.
```
$ ls -al build/CONFIG.EXE
-rwxr-xr-x. 1 maarten maarten 99840  6 feb 19:56 build/CONFIG.EXE
$ ls -al original/CONFIG.EXE
-rw-r--r--. 1 maarten maarten 95744  8 sep  1997 original/CONFIG.EXE
```

This app is perhaps not completely in scope of this reversing project,
but I believe it is useful anyways to understand display related logic.


madebr:
`CONFIG.EXE` can be reversed in ghidra.
This is how I made Ghidra to use readable symbols instead of ordinals:
1. closing all instances of ghidra
2. copy `mfc42.exports` from [this archive](https://github.com/isledecomp/isle/files/14185209/mfc42.zip) into `~/.ghidra/.ghidra_10.4.1_PUBLIC/symbols/win32`. You might need to replace the `PATH` attribute of `LIBRARY` within the xml with the path of your `MFC42.DLL`.
3. Open a Ghidra project
4. Add `CONFIG.EXE`
5. Analyze `CONFIG.EXE`

I don't know whether it is possible to use these symbols after analysis has been done.

foxtacles:
Looks good! @disinvite - if you could make sure the existing tooling works with this 3rd module, I think we can add this to the repository.

madebr:
Fun fact, but not tested: this could be the first part of isle to be built in 64-bit mode :)

CONFIG/AboutDlg.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+// FUNCTION: CONFIG 0x00403c20
+CAboutDialog::CAboutDialog() : CDialog(IDD)
+{
+}
+
+// FUNCTION: CONFIG 0x00403c90
+void CAboutDialog::BeginModalState()
+{
+	::EnableWindow(this->m_hWnd, FALSE);
```
ItzSwirlz, line 15: we still enforcing style for not using this->?

foxtacles:
As mentioned in the Matrix chat, since we have the tooling for it, I think we should put `CONFIG.EXE` on the same level as `ISLE` and `LEGO1` and add them to the build pipelines, and all the other actions (such as `analyze`) that we have. We probably don't necessarily need a SVG on the front page of the repository for the progress (though I wouldn't mind either).

CMakeLists.txt
```diff
@@ -60,6 +63,7 @@ message(STATUS "MSVC for decompilation: ${MSVC_FOR_DECOMP}")
 
 option(ISLE_WERROR "Treat warnings as errors" OFF)
 option(ISLE_BUILD_APP "Build ISLE.EXE application" ON)
+cmake_dependent_option(ISLE_BUILD_CONFIG "Build ISLE.EXE application" ON "NOT MINGW" OFF)
```
foxtacles, line 66: should mean "Build CONFIG.EXE", right?

CONFIG/config.cpp
```diff
@@ -0,0 +1,436 @@
...
+#include "ConfigCommandLineInfo.h"
+#include "MainDlg.h"
+#include "detectdx5.h"
+
+#include <direct.h> // _chdir
+#include <mxdirectx/mxdirect3d.h>
+#include <process.h> // _spawnl
+
+// FIXME: are these tags correct?
+// FIXME: how to tag static in-method variables?
```
foxtacles, line 12: @disinvite

CMakeLists.txt
madebr: Oops. (Who reads this thing anyway :) )

CONFIG/config.cpp
disinvite: These should go in config.h:
```
// FUNCTION: CONFIG 0x402c30
// CConfigApp::GetMessageMap

// GLOBAL: CONFIG 0x406008
// CConfigApp::messageMap

// GLOBAL: CONFIG 0x406010
// CConfigApp::_messageEntries
```

LEGO1/mxdirectx/mxdirect3d.cpp
```diff
@@ -538,13 +547,15 @@ void MxDeviceEnumerate::BuildErrorString(const char* p_format, ...)
 	OutputDebugString(buf);
 }
 
+// FUNCTION: CONFIG 0x00401bf0
 // FUNCTION: LEGO1 0x1009c4f0
 HRESULT CALLBACK MxDeviceEnumerate::DisplayModesEnumerateCallback(LPDDSURFACEDESC p_ddsd, LPVOID p_context)
```
madebr, line 552: Are these correct? Or do I need to put the following elsewhere?

```
// FUNCTION: CONFIG 0x00401bf0
MxDeviceEnumerate::DisplayModesEnumerateCallback
```

CONFIG/AboutDlg.cpp
madebr: Consider yourself an enforcer now :)
It's removed now officer ;)

CONFIG/config.cpp
madebr: Thanks applied this to all CWnd subclasses.

foxtacles:
Was https://github.com/isledecomp/isle/blob/d114437bb7ce1a808656d7bdf2060954c2ba6181/tools/ninja/Windows/ninja.exe added on purpose? Is it used by any script?

LEGO1/mxdirectx/mxdirect3d.cpp
foxtacles: That should be correct afaik. @disinvite can you confirm?

.github/workflows/build.yml
```diff
@@ -136,27 +143,31 @@ jobs:
     - name: Summarize Accuracy
       shell: bash
       run: |
+        python3 tools/reccmp/reccmp.py -S CONFIGPROGRESS.SVG -T 4252 --svg-icon tools/reccmp/config.png -H CONFIGPROGRESS.HTML legobin/CONFIG.EXE build/CONFIG.EXE build/CONFIG.PDB . | tee CONFIGPROGRESS.TXT
         python3 tools/reccmp/reccmp.py -S ISLEPROGRESS.SVG --svg-icon tools/reccmp/isle.png -H ISLEPROGRESS.HTML legobin/ISLE.EXE build/ISLE.EXE build/ISLE.PDB . | tee ISLEPROGRESS.TXT
-        python3 tools/reccmp/reccmp.py -S LEGO1PROGRESS.SVG -T 4252 --svg-icon tools/reccmp/lego1.png -H LEGO1PROGRESS.HTML legobin/LEGO1.DLL build/LEGO1.DLL build/LEGO1.PDB . | tee LEGO1PROGRESS.TXT
+        python3 tools/reccmp/reccmp.py -S LEGO1PROGRESS.SVG --svg-icon tools/reccmp/lego1.png -H LEGO1PROGRESS.HTML legobin/LEGO1.DLL build/LEGO1.DLL build/LEGO1.PDB . | tee LEGO1PROGRESS.TXT
```
foxtacles, line 148: The -T flag should stay with LEGO1 only

.github/workflows/build.yml
```diff
@@ -136,27 +143,31 @@ jobs:
...
     - name: Test Exports
       shell: bash
       run: |
-        python3 tools/verexp/verexp.py legobin/LEGO1.DLL build/LEGO1.DLL
+        tools/verexp/verexp.py legobin/LEGO1.DLL build/LEGO1.DLL
 
     - name: Check Vtables
       shell: bash
       run: |
+        python3 tools/vtable/vtable.py legobin/CONFIG.EXE build/CONFIG.EXE build/CONFIG.PDB .
```
foxtacles, line 170: should we remove python3 here as well?

madebr: Not sure, these jobs are running in a Windows VM.

LEGO1/mxdirectx/mxdirect3d.cpp
disinvite: The `CONFIG` annotations on the MxDirect3d functions all look fine. You don't need the leading zeroes on the hex address (to pad to 8) but I think the parser will allow it.

madebr:
> Was https://github.com/isledecomp/isle/blob/d114437bb7ce1a808656d7bdf2060954c2ba6181/tools/ninja/Windows/ninja.exe added on purpose? Is it used by any script?

No, it was a remainder of https://github.com/isledecomp/isle/pull/74

LEGO1/mxdirectx/mxdirect3d.cpp
madebr: My personal preference is to keep the leading zeros, but I'll change them if requested.

madebr:
I'm not sure about the annotations warning.
Usually MFC apps declare the global static `CWinApp` subclass in the global namespace.

CONFIG/config.cpp
```diff
@@ -0,0 +1,434 @@
...
+{
+	if (m_device_enumerator) {
+		delete m_device_enumerator;
+		m_device_enumerator = NULL;
+	}
+	return CWinApp::ExitInstance();
+}
+
+// GLOBAL: CONFIG 0x00408e50
+CConfigApp theApp;
```
disinvite, line 434: should be `CConfigApp g_theApp;`

madebr:
This is a summary of the vtable problem.
There are 4 classes, here is the class diagram (`>` == "is a subclass of")
```
CMainDialog            > CDialog    > CCmdTarget > CObject
CAboutDialog           > CDialog    > CCmdTarget > CObject
CConfigApp             > CWinThread > CCmdTarget > CObject 
CConfigCommandLineInfo                 > CObject
```
All objects of CObject implement various methods, under which:
```
void Serialize(CArchive& ar);
void AssertValid() const;
void Dump(CDumpContext& dc) const;
```

When I don't add an override to any class, then the table will contain `CObject::Serialize`, `CObject::AssertValid` and `CObject::Dump`.
But the Serialize/AssertValid/Dump present in the vtables in CONFIG.EXE of all 4 classes contain the same function (they are empty implementations).  

I tried the following, but `CObject::Serialize` (etc) remains in the vtable
```
class CommmonParent {
public:
  void Serialize(CArchive& ar);
  void AssertValid() const;
  void Dump(CDumpContext& dc) const;
}
class CMainDialog: public CDialog, public CommonParent {
public:
  using CommonParent::Serialize;
  using CommonParent::AssertValid;
  using CommonParent::Dump;
};
class CAboutDialog: public CDialog, public CommonParent {
public:
  using CommonParent::Serialize;
  using CommonParent::AssertValid;
  using CommonParent::Dump;
};
...
```

Making the methods in the parent virtual is no solution either, because then the size of all  structs increase (extra vfptr of the CommmonParent).

The logs show also a problem with `BeginModalState` and `EndModelState`. It is a similar problem, but can be fixed by a common ancestor for `CMainDialog` and `CAboutDialog`.

PR #538, opened 2024-02-07
Implement Infocenter::ReadyWorld

foxtacles:
Match is at ~84%, but all the functionality should be accurate.

### DIFF
--- a/LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
+++ b/LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
@@ -345,7 +345,7 @@ void Infocenter::ReadyWorld()
 
 	switch (GameState()->GetUnknown10()) {
 	case 0:
-		bg->Enable(1);
+		bg->Enable(TRUE);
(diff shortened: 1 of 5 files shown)


LEGO1/lego/legoomni/src/infocenter/infocenter.cpp
```diff
@@ -343,54 +345,170 @@ void Infocenter::ReadyWorld()
...
 		}
+
+		LegoAct2State* state = (LegoAct2State*) GameState()->GetState("LegoAct2State");
+		GameState()->FUN_1003ceb0();
+
+		if (state && state->GetUnknown0x08() == 0x68) {
+			bg->Enable(TRUE);
+			PlayCutscene(e_badEndMovie, TRUE);
+			m_infocenterState->SetUnknown0x74(0);
+			return;
```
ItzSwirlz, line 421: break?

foxtacles: no

PR #540, opened 2024-02-08
Add CONFIG.EXE to README

UnBeatWaterGH:
Currently I have only added the SVG, however we could add text explaining the purpose of CONFIG.EXE.

There are a few ways of displaying it, I have gone with this one:
config-isle-lego1
Though we could also use:
isle-config-lego1
Or:
isle-lego1-config
There are also more possibilities, but I'm only showing a few of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This is a **work-in-progress** decompilation of LEGO Island (Version 1.1, Englis
 
 ## Status
 
-<img src="https://legoisland.org/progress/ISLEPROGRESS.SVG" width="50%"><img src="https://legoisland.org/progress/LEGO1PROGRESS.SVG" width="50%">
+<img src="https://legoisland.org/progress/CONFIGPROGRESS.SVG" width="33%"><img src="https://legoisland.org/progress/ISLEPROGRESS.SVG" width="33%"><img src="https://legoisland.org/progress/LEGO1PROGRESS.SVG" width="33%">
 
 Currently, `ISLE.EXE` is completely decompiled and behaves identically to the original. A handful of stubborn instructions are not yet matching; however, we anticipate they will as more of the overall codebase is implemented.
 


madebr:
Question about the report: we use 640x640 bitmaps for the icons, which is wasteful as they are upscaled versions of 16x16 or 32x32 icons. Simply scaling those does not work because of blurring, but is it possible in css/html to scale fractionally or use the nearest color instead of interpolation?

ItzSwirlz:
I feel like CONFIG.EXE is so subtle that it isn't as important as ISLE or LEGO1 in terms of the overall repo. How about maybe having the ISLE and LEGO1 like it currently is, then just a simple progress bar, no SVG, a horizontal progress bar labeled "CONFIG.EXE" that shows the current accuracy percentage

UnBeatWaterGH:
> Question about the report: we use 640x640 bitmaps for the icons, which is wasteful as they are upscaled versions of 16x16 or 32x32 icons. Simply scaling those does not work because of blurring, but is it possible in css/html to scale fractionally or use the nearest color instead of interpolation?

In CSS you can use `image-rendering: pixelated` to get nearest-neighbour scaling.

Make sure to check the MDN page for it (linked [here](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering))

UnBeatWaterGH:
> I feel like CONFIG.EXE is so subtle that it isn't as important as ISLE or LEGO1 in terms of the overall repo. How about maybe having the ISLE and LEGO1 like it currently is, then just a simple progress bar, no SVG, a horizontal progress bar labeled "CONFIG.EXE" that shows the current accuracy percentage

Good idea, one thing, we would need an SVG (or of other image format) for the progress bar. Make of that what you want.

foxtacles:
I agree that just adding a third progress indicator in the same style as the others doesn't look great and takes away the focus from the main goal of the project (which is the completion of `LEGO1`). Maybe it's time to create an entirely different progress indication visualization that takes this and other things into account. We are certainly open to suggestions and pull requests in this matter, just remember since this comes down to (subjective) personal tastes it may or may not be approved depending on the feedback of the community.

Closing this for now, thank you for your initiative though.

PR #545, opened 2024-02-09
Match some names/values

tahg:
Align library to the 96 source code that we have. (As much as possible given changes between that and the final release.)


foxtacles:
Please add a description to the pull request. What's the basis for the changes to the names, code styles etc.?

PR #546, opened 2024-02-09
Implement LegoModelPresenter::ReadyTickle

MishaProductions:
Functionality seems to be correct, but different registers are being used for some reason which results in a 57.83% match


foxtacles:
Had one functional bug, now it's 100%

PR #554, opened 2024-02-10
reccmp: Unique addresses and stub reporting

disinvite:
The new feature from this PR is that stub functions are now reported in both the text and HTML outputs from `reccmp`. (attn: @tahg)

I added a `--silent` option to reccmp that omits the text output. This is more for my benefit when doing performance testing with the profiler so that there's less stuff printed to the terminal.

The behind-the-scenes work has to do with address uniqueness. Addrs from the original should be unique in our database because they only ever point at one thing. The decomp linter will detect if the address is reused, so we have protection on pull requests. Addrs from the recomp should be unique too, but there are situations where one would be duplicated in the data read from the PDB. For example: the `strlwr` function gets two symbols: `_strlwr` and `__strlwr`. 

Enforcing uniqueness is simple enough. I found that doing this the "correct" way with a unique index in sqlite is a lot slower than just checking whether the address has been used. Maybe the queries aren't using the indices effectively? I can look into that later.

Our addresses were already almost totally unique anyway, except for the two cases that I'll explain now:

One problem is name matching. If the PDB has two symbols with unique address but the same name, we need to match only one of them. We were using an `UPDATE` statement for this, but this will match all rows that have the given name. It _is_ possible for sqlite to update a single row in this situation (with `LIMIT 1`) but not unless the `SQLITE_ENABLE_UPDATE_DELETE_LIMIT` option is enabled. This is not the case for our built-in python sqlite, so the solution is to just select the single recomp address that meets our matching criteria, and then do the update. This funnels all "matching" through the `set_pair` function, but I think this is a good thing because it means we can change our implementation more easily. The database internals should be abstracted away as much as possible.

(A possible enhancement here would be to _not_ match _unless_ only a single row comes back from our query.)

This problem bit us in the `_Construct` function. I have changed the decomp marker from `legoworld.h` to point at the decorated name `?_Construct@@YAXPAPAVMxCore@@ABQAV1@@Z` because this is what we were trying to match. Real name: `void __cdecl _Construct(class MxCore * *,class MxCore * const &)`. I added the option for the compare database to match using the decorated name if that's obviously what we have, rather than using the "least obtuse name" as we would normally.

The other odd case has to do with strings and their null terminator. I wrote more about this in the code comments, but basically we need to distinguish between the empty string `"\x00"` and the mostly empty string `"\x00\x00"`. Any string function would handle both as being empty, but they are two different symbols in the PDB for some reason. I'm not sure where the case with two nulls is used; maybe in one of the MSVC libraries.

### DIFF
--- a/LEGO1/lego/legoomni/include/legoworld.h
+++ b/LEGO1/lego/legoomni/include/legoworld.h
@@ -155,7 +155,7 @@ protected:
 // _Tree<MxCore *,MxCore *,set<MxCore *,CoreSetCompare,allocator<MxCore *> >::_Kfn,CoreSetCompare,allocator<MxCore *> >::find
 
 // TEMPLATE: LEGO1 0x10022360
-// _Construct
+// ?_Construct@@YAXPAPAVMxCore@@ABQAV1@@Z
 
 // GLOBAL: LEGO1 0x100f11a0
 // _Tree<MxCore *,MxCore *,set<MxCore *,CoreSetCompare,allocator<MxCore *> >::_Kfn,CoreSetCompare,allocator<MxCore *> >::_Nil
--- a/tools/isledecomp/isledecomp/compare/core.py
+++ b/tools/isledecomp/isledecomp/compare/core.py
@@ -20,6 +20,7 @@ logger = logging.getLogger(__name__)
 
 @dataclass
 class DiffReport:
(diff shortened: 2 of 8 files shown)


foxtacles:
Looks good to me! I'll wait for feedback of @tahg before merging this.

PR #560, opened 2024-02-13
LegoCacheSound and related STL

disinvite:
This started as an exploration into the STL members of `LegoUnknown100d6b4c`, and then expanded out into `LegoCacheSound` and a new related class `LegoUnknown100d5778`.

My goal in pulling in those other classes was to try to understand how `LegoUnknown100d6b4c` works. I don't have the _why_ but I'm more confident about the members being a set and list. It didn't really come together until I reversed the function at `0x1003d290`. The list and set are both using the same "value type" for the nodes, and you can tell that because they both use the same destructor at `0x1003d030`. The unwind section shows this object being instantiated as the arg for the STL functions. (In typical MSVC fashion, I had this function at 100% but it stopped matching when I used getter methods instead of direct member access.)

I'm less confident about the object itself. I have the struct `Element100d6b4c` in there now. This is basically just `std::pair<const char*, LegoCacheSound*>` except for that destructor. Might be missing something obvious here.

Other small items:
* ncc was complaining about the variable name with an underscore in `MxAtomId` so that's fixed. Coincidentally, the `MxAtomIdCounterSet` class is another instance where it seems like we should have a map, but a set fit better for _reasons_. The node size is 0x18, which means two pointers (i.e. key and value) instead of just one, and that would also suggest a map.
* The table in the reccmp HTML output was getting broken due to the very long symbol names without a space character.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,6 +270,7 @@ add_library(lego1 SHARED
   LEGO1/lego/legoomni/src/audio/legocachesound.cpp
   LEGO1/lego/legoomni/src/audio/legoloadcachesoundpresenter.cpp
   LEGO1/lego/legoomni/src/audio/legosoundmanager.cpp
+  LEGO1/lego/legoomni/src/audio/legounknown100d5778.cpp
   LEGO1/lego/legoomni/src/audio/legounknown100d6b4c.cpp
   LEGO1/lego/legoomni/src/audio/mxbackgroundaudiomanager.cpp
   LEGO1/lego/legoomni/src/build/buildingentity.cpp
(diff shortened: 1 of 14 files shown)


foxtacles:
Looks good!

The `Element100d6b4c` destructor just lacked the test for the value of `m_name`. I've also declared the comparator struct a friend of `Element100d6b4c`, to enable that ~100% match that you mentioned is enabled by direct member access through the comparison function. I feel like this is something they could have done, although of course there may be many other inlining patterns that could maybe yield the same, but I think this makes enough sense.

disinvite:
Yeah, that makes sense. Thanks for fixing that Tickle function that was very hackily put together.

PR #561, opened 2024-02-13
GasStation destructor, enable, create

ItzSwirlz:
I'm still a little confused with some things. There is some global variable, and to be honest I'm new to dealing with globals and statics in decompilation in general, but I'm off a little bit with that global data. The destructor doesn't match because of this.

In the Create function, there is some check about the GasStationState's m_unk0x14 that I just cannot seem to get it match. Ghidra says it sets the value to 4, and then checks if the previous value is not 4, but I already tried doing that and it wasn't matching. Something weird is happening there.

### DIFF
--- a/LEGO1/lego/legoomni/src/gasstation/gasstation.cpp
+++ b/LEGO1/lego/legoomni/src/gasstation/gasstation.cpp
@@ -7,13 +7,14 @@
 #include "mxnotificationmanager.h"
 #include "mxticklemanager.h"
 
+// GLOBAL: LEGO1 0x100f0160
 undefined4 g_unk0x100f0160;
 
 // FUNCTION: LEGO1 0x100046a0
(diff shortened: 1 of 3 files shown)


LEGO1/lego/legoomni/src/gasstation/gasstation.cpp
```diff
@@ -1,12 +1,19 @@
 #include "gasstation.h"
 
+#include "legocontrolmanager.h"
+#include "legogamestate.h"
+#include "legoinputmanager.h"
+#include "legoomni.h"
 #include "mxnotificationmanager.h"
+#include "mxticklemanager.h"
+
+undefined4 g_unk0x100f0160;
```
disinvite, line 10: Adding `// GLOBAL: LEGO1 0x100f0160` here will help reccmp match this variable.

PR #564, opened 2024-02-15
PoliceState: Create script enum + implement FUN_1005ea40

ItzSwirlz:
85% match for the function. Registry swaps.


foxtacles:
> Registry swaps.

No, the code was incorrect. Also, it's a register: https://en.wikipedia.org/wiki/Processor_register

ItzSwirlz:
> > Registry swaps.
> 
> 
> 
> No, the code was incorrect. Also, it's a register: https://en.wikipedia.org/wiki/Processor_register

My Minecraft modding got me in a trance clearly

PR #579, opened 2024-02-19
Implement HistoryBook::ReadyWorld

tahg:
Not a great match yet, as most of the local variables are at the wrong offsets. I don't think it's miles away otherwise, but there are some sections that don't match well.


tahg:
No idea why Format is failing, as the tool just errors out at the end of the file without finding whatever it's looking for.  Maybe it's the nesting, because I can't find anything else unusual about it.

foxtacles:
Fixed most of the inaccuracies and a few bugs. The bulk of the inaccuracy % left comes from `MxBitmap::GetStart`, which is the same inaccuracy we have in all functions that use it, so that one is probably worth taking another look at.

PR #581, opened 2024-02-19
reccmp: HTML refactor and diff address display

disinvite:
Our diff output uses `unified_diff` from the Python `difflib` library to produce text like this:

```diff
---
+++
@@ -1,6 +1,6 @@
mov eax, dword ptr [esp + 4]
test eax, eax
je 0x6
push eax
-call <OFFSET1>
+call _MemFreePtr@4 (FUNCTION)
ret
```

To help with the display, we can add some color to lines that begin with `-` or `+`. Markdown  just did the same above.

Many functions are long and the parts that differ are small, so we show up to 10 lines of "context" before and after each diff rather than the entire thing. The `@@ -1,6 +1,6 @@` blurb at the top tells where we are in the textual output. This means we are looking at lines 1 through 6 of both the original and recomp assembly output.

Question: how can we show the virtual addresses instead?

We can't put the addresses in the actual text output because they are not the same between original and recomp. The `difflib.SequenceMatcher` class has an option to return "opcodes". These are a list of instructions that describe what sections of the two text samples are alike, and which sections need to be added, removed, or replaced. `unified_diff` uses this to produce the text output.

My solution to this problem was to pair up instruction addresses with the assembly text, diff just the text component, and then use the opcodes to assemble diff output with instructions included. We just need to change how the diffs are printed, so now you get this from `reccmp` verbose mode:

```diff
0x10086260 : mov eax, dword ptr [esp + 4]
0x10086264 : test eax, eax
0x10086266 : je 0x6
0x10086268 : push eax
0x10086269 : -call <OFFSET1>
0x10061af9 : +call _MemFreePtr@4 (FUNCTION)
0x1008626e : ret

??3@YAXPAX@Z is only 83.33% similar to the original, diff above
```

You can provide the `--print-rec-addr` option to show both original and recomp addresses if you want.

We also use this library to show vtable diffs, but these also have a sort of address/text pairing, so I modified that comparison code to produce similar output. This required a change to the `vtable` test tool and the impact to the feature set is that "verbose mode" is now always enabled. I think this is an acceptable tradeoff for now; we are showing more information, not less, and the tool is not really used unless there is a problem.

The other big thing to figure out is how to display the new diffs in the HTML report, and I took the opportunity to refactor the javascript code we had been using. To avoid pulling in any dependencies, I used the [Web Component API](https://developer.mozilla.org/en-US/docs/Web/API/Web_components) to take a React-like approach to the code with lifecycle methods.

One new feature is that you can now use the search bar to look at the assembly output, either including or excluding the matching "context" lines. For example, you can search "*4" to see which functions have a jump table instruction that looks like `jmp dword ptr [eax*4 + <offset>]`.

Feedback is welcome! This design is definitely not set in stone.

### DIFF
--- a/tools/isledecomp/isledecomp/compare/asm/parse.py
+++ b/tools/isledecomp/isledecomp/compare/asm/parse.py
@@ -192,11 +192,13 @@ class ParseAsm:
     def parse_asm(self, data: bytes, start_addr: Optional[int] = 0) -> List[str]:
         asm = []
 
-        for inst in disassembler.disasm_lite(data, start_addr):
+        for raw_inst in disassembler.disasm_lite(data, start_addr):
             # Use heuristics to disregard some differences that aren't representative
             # of the accuracy of a function (e.g. global offsets)
-            result = self.sanitize(DisasmLiteInst(*inst))
+            inst = DisasmLiteInst(*raw_inst)
+            result = self.sanitize(inst)
+
             # mnemonic + " " + op_str
-            asm.append(" ".join(result))
+            asm.append((hex(inst.address), " ".join(result)))
 
         return asm
--- a/tools/isledecomp/isledecomp/compare/core.py
+++ b/tools/isledecomp/isledecomp/compare/core.py
@@ -12,6 +12,7 @@ from isledecomp.dir import walk_source_dir
 from isledecomp.types import SymbolType
 from isledecomp.compare.asm import ParseAsm, can_resolve_register_differences
 from .db import CompareDb, MatchInfo
+from .diff import combined_diff
 from .lines import LinesDb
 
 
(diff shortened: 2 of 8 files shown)


foxtacles:
Looks great! Would it be possible to also show these lines:

```---
+++
@@ -1,67 +1,67 @@```

in the `reccmp` console output for an individual function?

foxtacles:
Looks great! Good to merge for me, but will wait for feedback from others for a bit.

disinvite:
One thing I forgot to mention: I took out the context indicator (the `@@ -1,6 +1,6 @@` thing.) Do we need that if we have the addresses instead? I put a screenshot on Matrix chat that had the address range in there instead of line numbers, but I think even that is redundant. You can just read the addresses to tell where you are. I can add it back in if people want it, though.

foxtacles:
> One thing I forgot to mention: I took out the context indicator (the `@@ -1,6 +1,6 @@` thing.) Do we need that if we have the addresses instead? I put a screenshot on Matrix chat that had the address range in there instead of line numbers, but I think even that is redundant. You can just read the addresses to tell where you are. I can add it back in if people want it, though.

I think it might still be useful, just so it's a little easier to quickly visually recognize different contexts.

tahg:
I agree that some sort of break indicator would be good. If the context indicator is added back in, it would be nice if it was colored blue, like other diff coloring tools do. And while non-standard, I'd like the @@ to show the old/new binary offsets, as I think that'd be more useful than line numbers.

PR #584, opened 2024-02-21
CustomizeAnimFile MxVariable

disinvite:
Another child of `MxVariable`, with the key string "CUSTOMIZE_ANIM_FILE". The value is sent to static functions from `LegoUnkSaveDataWriter`, `LegoPlantManager` and `LegoBuildingManager`. They all have the same code to set a static member for the respective class.

The constructor for `LegoUnkSaveDataWriter` should inline almost all of the constructor for the new variable, but it's not happening. If not for that then it should match. The function at `0x10085aa0` is a wrapper for the `MxVariable` constructor except that it sets a second vtable.

The STL map from `LegoUnkSaveDataWriter` is also referenced in the constructor so I looked into that too. I feel pretty confident about the value type being the pair of  `LegoROI*` and `MxU32/MxS32`; I left it as its own struct for the moment. `FUN_10083500` wraps around the `_Insert` function of the map. If the value struct didn't already exist, it gets set up around `0x100835f4`. If it did exist, the counter is incremented at `0x100835aa`. `FUN_10084030` does something to the `LegoROI*` parameter before setting it in the struct.

### DIFF
--- a/LEGO1/lego/legoomni/include/legobuildingmanager.h
+++ b/LEGO1/lego/legoomni/include/legobuildingmanager.h
@@ -21,6 +21,7 @@ public:
 	}
 
 	static void configureLegoBuildingManager(MxS32);
+	static void SetCustomizeAnimFile(const char* p_value);
 
 	void FUN_1002fa00();
 	void FUN_1002fb30();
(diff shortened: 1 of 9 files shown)


foxtacles:
The variable class and `LegoUnkSaveDataWriter` work a bit differently. I've pushed the changes. It all matches 100% now

PR #618, opened 2024-03-03
Data comparison tool

disinvite:
As we get further into meaningfully starting the game and debugging the decomp, it's now time to compare variables.

The main challenge when doing this is complex types. We could just compare the binary blob for a non-primitive type, but if the structure has any pointers then these values will register as a diff, even if they point at the same thing in the respective binaries. Also, if there _is_ a diff, it would be nice to know which member has a problem and display it according to its type instead of a hex dump.

To do that, we are now parsing the `TYPES` section from cvdump output. This section has comprehensive type information for not just variable data types, but also things like function argument lists. We're not interested in those right now, but we can parse all the information required to compare variables in the current pdb. This new parser reads all the text data from cvdump and provides four methods that walk the dependency tree to return useful information:

- The `get()` method gives you generic "type information" like the total size and list of members for a complex type. If we wanted to cross reference our member offset comments with the actual offsets, we could use this list.
- `get_scalars()` is the main method that drives the comparison. This recursively expands a complex type until we have a list of primitive values.
- `get_format_string()` combines the two previous methods into a text string to be passed into Python's `struct.unpack()` function. We can read the data from the original binary and display the real values according to data type and signed vs. unsigned.
- `get_by_name()` is unimplemented, but this will allow you to send in a class or struct name and get the type information back. When finished, we can use this to supplement (or replace) the `DECOMP_SIZE_ASSERT` statements.

I added `datacmp.py` to do the actual comparison. Argument list is the same as `reccmp`, with `original-binary`, `recompiled-binary`, `recompiled-pdb`, and `decomp-dir` as the ordinal parameters. The default view shows only variables that don't match. `--verbose` mode shows everything. We also support the `--print-rec-addr` and `--no-color` options.

If we have no type information at all -- which is the case for some SmartHeap variables and the handful of statics that we've added -- then all we can do is compare the raw bytes. This may not go so well because there is a chance we will pull in more data than actually belongs to the given variable. I'm marking diffs in these situations as a "warning" for further review.

As usual with new tooling, I fixed a few things in `CONFIG` and `LEGO1` where a problem is now apparent. Of the three that are left in `LEGO1`:
- `g_saveDataInit` needs more reversing on the members, specifically with `m_pFrameData`. When that's done, this tool will help with populating the many fields.
- I'm not sure on the syntax for `g_extraPaths` but it seems to be 29 totally empty structs.
- `g_saveData3` should be uninitialized and it has no value in the file. For some reason, MSVC set an address where _some_ of the bytes are in the initialized section of `.data`. This causes the tool to assume it has an initialized value when it does not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -297,7 +297,6 @@ add_library(lego1 SHARED
   LEGO1/lego/legoomni/src/common/legoplantmanager.cpp
   LEGO1/lego/legoomni/src/common/legostate.cpp
   LEGO1/lego/legoomni/src/common/legotextureinfo.cpp
-  LEGO1/lego/legoomni/src/common/legounksavedatawriter.cpp
   LEGO1/lego/legoomni/src/common/legoutil.cpp
   LEGO1/lego/legoomni/src/common/legovariables.cpp
   LEGO1/lego/legoomni/src/common/mxcompositemediapresenter.cpp
@@ -353,6 +352,7 @@ add_library(lego1 SHARED
   LEGO1/lego/legoomni/src/isle/radiostate.cpp
   LEGO1/lego/legoomni/src/isle/skateboard.cpp
   LEGO1/lego/legoomni/src/main/legoomni.cpp
+  LEGO1/lego/legoomni/src/common/legounksavedatawriter.cpp
   LEGO1/lego/legoomni/src/notify/legoeventnotificationparam.cpp
   LEGO1/lego/legoomni/src/paths/legopathcontroller.cpp
   LEGO1/lego/legoomni/src/paths/legopathpresenter.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -297,6 +297,7 @@ add_library(lego1 SHARED
   LEGO1/lego/legoomni/src/common/legoplantmanager.cpp
   LEGO1/lego/legoomni/src/common/legostate.cpp
   LEGO1/lego/legoomni/src/common/legotextureinfo.cpp
+  LEGO1/lego/legoomni/src/common/legounksavedatawriter.cpp
   LEGO1/lego/legoomni/src/common/legoutil.cpp
   LEGO1/lego/legoomni/src/common/legovariables.cpp
   LEGO1/lego/legoomni/src/common/mxcompositemediapresenter.cpp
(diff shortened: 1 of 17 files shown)


foxtacles:
As always, thank you for this very valuable contribution. This goes a long way to further ensure correctness of the decompilation!

In C and C++, both global and static variables should be zero-initialized, which means that `g_extraPaths` for instance should already be all zero. I'm not sure why it's not working in this case. I've declared it `static` and now `datacmp` seems to accept it. In Ghidra, it still seems to be uninitialized memory though? Maybe I'm missing something.

disinvite:
Glad to help!

I think I've figured this out. `g_extraPaths` runs off the end of the `.data` section. It is 1508 bytes in total size and should be uninitialized. Despite that fact, `LEGO1.dll` reserves initialized data (all zero) for the first 216 bytes.

Practically speaking, it doesn't matter that the variable is "half-initialized" like this. The uninitialized addresses are zeroed out when Windows loads the image file, according to the PE standard:

> The size of the section (for object files) or the size of the initialized data on disk (for image files). For executable images, this must be a multiple of FileAlignment from the optional header. **If this is less than VirtualSize, the remainder of the section is zero-filled.** Because the SizeOfRawData field is rounded but the VirtualSize field is not, it is possible for SizeOfRawData to be greater than VirtualSize as well. When a section contains only uninitialized data, this field should be zero.
>
>https://learn.microsoft.com/en-us/windows/win32/debug/pe-format

So the effect is that reading any of the virtual addresses for this variable would read a zero. I'm not sure why _this_ variable is the one left on the boundary. Maybe a combination of the variable being large enough and the "rounding" mentioned for the section's SizeOfRawData value.

What I saw before is that it is instead the `g_saveData3` variable that falls on the boundary for `.data`. If I move `legounksavedatawriter` after `legoomni` in CMake, things line up better. `roadmap` suggested order also supports doing this.

This now reveals a new bug that I'm still working on: I hadn't anticipated the situation where you would need to read past the end of the raw data (in IsleBin) so the code cuts off the buffer rather than padding with zero bytes. This causes `struct.unpack` to blow up; your buffer isn't big enough for what it expects to unpack.

I guess the best situation here is to emulate a read from the fully loaded image: the uninitialized data is zeroed out, and you can spill over into other sections if you read enough data.

disinvite:
Rather than change binfile reading, I think what will work is to just assume that a truncated read means the variable is uninitialized, so treat it as such.

I tried setting `g_extraPaths` to a char array of all zeroes, then all zeroes except the first byte. They both initialize the entire array. Dangling the variable off the end of the section does not seem to occur _unless_ the variable is uninitialized.

foxtacles:
Interesting, thanks for looking into this!

I've reverted the order change of `legounksavedatawriter` since it seems out of place to re-order one unit while all the rest is still ordered alphabetically. It would be great if we can establish order for all the units in one fell swoop rather than piecemeal, at some point.

PR #626, opened 2024-03-04
fix bug in MxPresenter::CreateEntity

MishaProductions:
The game crashed as MxPresenter::CreateEntity returned null, and LegoWorldPresenter expects it not to be null. Fixed an issue where the default name would be overwritten with the extra data size. After this fix, the game now crashes in LegoModelPresenter as the names of LegoROI aren't being read yet.


foxtacles:
There doesn't seem to be any functional difference according to `reccmp`?

PR #649, opened 2024-03-10
refactor script enum structure

Ramen2X:
This PR completely refactors the way we structure `Script` enums, moving from ones we've manually defined by hand to using enums from headers that were automatically generated from 1.1 English Interleaf files using the [actionheadergen](https://github.com/isledecomp/actionheadergen) tool.

While these manually written names are definitely nice, there is a glaring labor problem with continuing to use this structure going forward. When you consider that some Interleaf files (such as ISLE.SI) can have upwards of *4000* unique actions, the problem becomes extremely apparent. As we'd like to avoid using magic numbers in the codebase as much as we possibly can, this is the best compromise.

The use of this structure also drastically increases portability of the decompilation to other versions of LEGO Island as well, as where most builds drastically differ is in their action IDs.

I've secluded all of these headers to an `actions` folder in `legoomni/include`, so they aren't in the way of handwritten code.

These automatically generated enums are scoped in classes to avoid conflicts, as some Interleaf files share action names.

I've reworked all of the game logic I could find that used the old enums to use the new ones. There's a lot of code in this repository now, so I may have missed something. A thorough review is greatly appreciated.


Ramen2X:
I'm not exactly sure how we're going to get ncc to ignore these. Ruffling through their documentation a little bit, it seems like excluding entire directories isn't currently a supported feature; although some of the work needed for it is there, just commented out.

Ramen2X:
Alright, I've made some changes. The generated headers now scope their enums in namespaces instead of classes (not yet reflected in the generation tool, but will push the relevant changes soon). I've also made a change to the `naming.yml` workflow to dynamically build a list of headers in the `legoomni/include/actions` folder and store it in a variable to be passed on later to ncc's ignore list, which should spare us from the future-proofing issues with hardcoding and the labor involved with it.

foxtacles:
Looks great! Thanks a lot, this is obviously far superior to manually maintaining the IDs.

PR #656, opened 2024-03-11
KeyValueStringParse closer to beta version

disinvite:
Changed some code in the public KeyValueStringParse function from `MxUtilities` to more closely match Beta 1.0. Two parameter names and a variable name were revealed in the asserts and I changed `p_output` to be shorter to match.

No effect on match percentage for the release `LEGO1` except marginal compiler entropy in three places.

My main goal with this is to get some discussion going about what things from the beta we should add to our codebase.

- Do we want the asserts even though they are no-ops in non-debug builds? (Maybe that _is_ the reason to add them)
- Any consideration for line numbers of the asserts? I commented the line numbers for the three here, but ours will probably never match exactly with the addition of the decomp assertions and ifdefs for building on newer compilers.
- Do we want decomp assertions for the beta build to compare against? What should the module name be? `BETA10`?
- I was looking at `MxPresenter::ParseExtra` and that led me to this function. The `MxAutoLocker` call is a bit different. The 96 source shows that they used the macro `AUTOLOCK` in place of `MxAutoLocker lock(&m_criticalSection)`. I think the reason is that in the debug version they add two parameters: `__FILE__` and `__LINE__`, similar to an assert. Do we want to do the same for our debug builds?

*Maybe we can find a better name for this function too*

### DIFF
--- a/LEGO1/omni/src/common/mxutilities.cpp
+++ b/LEGO1/omni/src/common/mxutilities.cpp
@@ -83,12 +83,12 @@ void MakeSourceName(char* p_output, const char* p_input)
 MxBool KeyValueStringParse(char* p_output, const char* p_command, const char* p_string)
 {
 	MxBool didMatch = FALSE;
-	assert(p_string);  // 90
-	assert(p_command); // 91
+	assert(p_string);
+	assert(p_command);
 
 	MxS16 len = strlen(p_string);
 	char* string = new char[len + 1];
-	assert(string); // 95
+	assert(string);
 	strcpy(string, p_string);
 
 	for (char* token = strtok(string, ", \t\r\n:"); token; token = strtok(NULL, ", \t\r\n:")) {
--- a/LEGO1/omni/src/common/mxutilities.cpp
+++ b/LEGO1/omni/src/common/mxutilities.cpp
@@ -9,6 +9,8 @@
 #include "mxpresenterlist.h"
 #include "mxrect32.h"
 
+#include <assert.h>
+
 // GLOBAL: LEGO1 0x101020e8
 void (*g_omniUserMessage)(const char*, int) = NULL;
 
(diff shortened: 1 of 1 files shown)


foxtacles:
> No effect on match percentage for the release `LEGO1` except marginal compiler entropy in three places.

Can it be shown that it is compiler randomness? I've seen many cases that looked like it but were permanently resolved with some changes. Considering the function has matched 100% before the changes in this PR we'd need some evidence to show for sure that it is compiler randomness (generally adding/changing "stuff" outside the function should get it back to 100%).

In general I'd like to remind/stress that the main focus should remain on matching the code to the retail assembly as much as possible. The beta debug build can aid with that but we have to remember it's still very possible that code has changed between that version and the retail version.

> Do we want the asserts even though they are no-ops in non-debug builds? (Maybe that is the reason to add them)

For preservation purposes we can add them the way we find them in the beta debug build. I think this is more of a stretch goal than anything though.

> Any consideration for line numbers of the asserts?

It's practically impossible for us to get the same line numbers as in the original (or even the beta debug build). There is an endless amount of possible variations on the source code that will yield the same assembly but change line numbers, even disregarding our decomp additions - so I don't think there is any value in annotating the line numbers.

> Do we want decomp assertions for the beta build to compare against? What should the module name be? BETA10?

I think annotating the address of a function from the beta debug build *could* be helpful to quickly point us to the code in order to facilitate the decompilation of the retail version, however I don't see much of a point in comparing the code. We don't actually want to match the code of the beta build, it may just be coincidence that we can get it to match as well, but it's not a goal of this project.

For that reason, if we add annotations for the beta, I'd probably not add them in the same fashion we use our current annotations, but create a separate/distinct place for them, maybe similar to the `library_*.h` files.

> The 96 source shows that they used the macro AUTOLOCK in place of MxAutoLocker lock(&m_criticalSection).

Yes this should be changed to use that `AUTOLOCK` macro, assuming of course it won't impair the match, but iirc it shouldn't

tahg:
Line numbers might be useful for relative ordering of code within a file, or as a rough indication of which classes are in the file, although we have enough address hints to mostly determine that already.

disinvite:
Thanks, all good thoughts. My reason for comparing against the debug build is just literally to compare. We already have all this code so it's interesting to see where things are close. I arrived at this function sort of randomly and I could not make any headway on the caller `MxPresenter::ParseExtra`.

I see now I should have wrote three *functions* instead of "three places". Sorry to mislead. *This* function generates the same assembly before and after the changes. The three diffs are these:

```
Increased (2):
0x100bd760 - DecodeChunks (66.67% -> 67.34%)
0x100bfb30 - DeserializeDSObjectDispatch (92.34% -> 92.69%)

Decreased (1):
0x100b6e10 - GetRectIntersection (7.56% -> 7.26%)
```

The two that increased only differ in the jump table at the bottom of the functions. We have slightly different addresses in the table so we generate fewer junk instructions and it matches "better". (I still need to finish that new feature that will sanitize these.)

`GetRectIntersection` did change, but the function is so far from matching that it's hard to say one way or the other.

FWIW, I can't get the key value function back up to 100% with the usual tricks.

foxtacles:
> FWIW, I can't get the key value function back up to 100% with the usual tricks.

Then I'd say the code should be changed so that it matches 100% again. Of course we can keep the asserts, updated variable names etc. but otherwise it seems like a regression to me - or am I missing something?

disinvite:
Oh, huh. The CI diff doesn't agree with what I've got in the branch. Maybe _I'm_ missing something from master.

disinvite:
Something weird's going on here. On my local I have the same 91.51% diff building from master without this change, but the CI job shows 100% match. I'll table this until I figure that out.

disinvite:
It's the asserts. This gets it back to 100%:

```cpp
// FUNCTION: LEGO1 0x100b7050
MxBool KeyValueStringParse(char* p_output, const char* p_command, const char* p_string)
{
    #define assert /**/
    MxBool didMatch = FALSE;
    assert(p_string);
    assert(p_command);
```

So where does that leave us?

(The problem with my local compare is that `ISLE_USE_SMARTHEAP` was still set to off from when I was building target=debug.)

foxtacles:
Hmm, I see. I guess `((void)0)` triggers the compiler randomness then. That is annoying. I'm not sure what's the best course of action here to be honest; I suppose it is sufficient proof that the regression is "just" compiler entropy and we can accept it.

If we determine that asserts keep impairing match percentages in other places that aren't present without them, we can still macro expand it to nothing I guess.

foxtacles:
image

This also gets the function back to 100%. So I think we are good.

PR #660, opened 2024-03-12
Use AUTOLOCK macro

disinvite:
This unifies all calls to `MxAutoLocker` to use the macro `AUTOLOCK`, like this: `AUTOLOCK(m_criticalSection)`. Should be no diffs.

The macro comes from the 96 source. Debug builds call the `MxAutoLocker` constructor with the special `__FILE__` and `__LINE__` macros to produce the code filename and line number. `MxCriticalSection::Enter` will print those to the deadlock file in the event of a problem. I didn't add the debug-only changes but we can now do it easily.

It's not clear which header file had this macro so I put it in mxautolocker.h. That's already included wherever this is used.

Notes:
- I took out `this->` where it was used.
- `MxOmni` named the member `m_criticalsection` without camel case. I changed that to match the others.
- `MxNotificationManager` calls the member `m_lock` but I left that alone.
- `MxDiskStreamController` has two critical section members, so they can't both be `m_criticalSection`. No name change here either.


foxtacles:
I think - iirc - the class was originally called `MxAutoLock` and not `MxAutoLocker`. Maybe you can double check that with the 1996 source and/or the beta debug build asserts. In case it's different we should rename it as well

disinvite:
Yes, there is a reference to MxAutoLock.h in one of the included makefiles. I changed it here.

LEGO1/omni/include/mxautolock.h
```diff
@@ -0,0 +1,17 @@
+#ifndef MXAUTOLOCK_H
+#define MXAUTOLOCK_H
+
+#include "mxcriticalsection.h"
+
+#define AUTOLOCK(CS) MxAutoLock lock(&CS);
```
madebr, line 6: Skip the semicolon, so users need to add it.
```suggestion
#define AUTOLOCK(CS) MxAutoLock lock(&CS)
```

PR #671, opened 2024-03-13
Implement MxMemoryPool

disinvite:
Another insight from the beta. The MxStreamer class has two nearly identical members with a constructor that does some odd things:
```cpp
    inline MxStreamerSubClass1(undefined4 p_size)
    {
        m_buffer = NULL;
        m_size = p_size;
        undefined4* ptr = &m_unk0x08;
        for (int i = 0; i >= 0; i--) {
            ptr[i] = 0;
        }
    }
```

They are referenced in MxDSBuffer where there is equally strange bit twiddling in AllocateBuffer and the destructor.

The beta showed that the two members of MxStreamer are MxMemoryPool objects, wrappers around an MxBitset datatype. STL `bitset` provides a convenient way to access and flip individual bits while using a minimal memory footprint. MxBitset is just a reimplementation of that, and I copied most of the structure from the STL file, as I assume Mindscape did. This is almost entirely inlined in the retail LEGO1, but I confirmed that they did _not_ switch to using the STL version because doing so adds more code.

I made the required changes to MxDSBuffer and MxStreamer and the result looks very good. The functions that decreased in MxDSBuffer are definite entropy; they go back to normal when I add one of the weirdo code blocks. e.g.: (credit @foxtacles )
```cpp
class X {
  int y;
  inline void QWE() {}
};
```

### DIFF
--- a/LEGO1/omni/include/mxbitset.h
+++ b/LEGO1/omni/include/mxbitset.h
@@ -11,83 +11,84 @@
 template <size_t N>
 class MxBitset {
 public:
-	MxBitset() { _Tidy(); }
+	MxBitset() { Tidy(); }
 
-	class reference {
+	class Reference {
 		friend class MxBitset<N>;
 
 	public:
-		reference& flip()
+		Reference& Flip()
 		{
-			m_bitset->flip(m_offset);
+			m_bitset->Flip(m_offset);
 			return (*this);
 		}
-		bool operator~() const { return (!m_bitset->test(m_offset)); }
-		operator bool() const { return (m_bitset->test(m_offset)); }
+		bool operator~() const { return (!m_bitset->Test(m_offset)); }
+		operator bool() const { return (m_bitset->Test(m_offset)); }
 
 	private:
-		reference(MxBitset<N>& _X, size_t _P) : m_bitset(&_X), m_offset(_P) {}
+		Reference(MxBitset<N>& p_bitset, size_t p_offset) : m_bitset(&p_bitset), m_offset(p_offset) {}
 		MxBitset<N>* m_bitset;
 		size_t m_offset;
 	};
 
-	reference operator[](size_t p_bit) { return (reference(*this, p_bit)); }
+	Reference operator[](size_t p_bit) { return (Reference(*this, p_bit)); }
 
-	MxBitset<N>& flip(size_t p_bit)
+	MxBitset<N>& Flip(size_t p_bit)
 	{
 		if (N <= p_bit) {
-			_Xran();
+			Xran();
 		}
-		m_blocks[p_bit / _bits_per_block] ^= 1 << p_bit % _bits_per_block;
+		m_blocks[p_bit / e_bitsPerBlock] ^= 1 << p_bit % e_bitsPerBlock;
 		return (*this);
 	}
 
-	size_t count()
+	size_t Count()
 	{
 		// debug only
 		return 0;
 	}
 
-	bool test(MxU32 p_bit)
+	bool Test(MxU32 p_bit)
 	{
 		if (p_bit >= N) {
-			_Xran();
+			Xran();
 		}
 
-		return (m_blocks[p_bit / _bits_per_block] & (1 << p_bit % _bits_per_block)) != 0;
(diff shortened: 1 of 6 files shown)


LEGO1/omni/include/mxmemorypool.h
```diff
@@ -0,0 +1,86 @@
...
+#include "decomp.h"
+#include "mxbitset.h"
+#include "mxtypes.h"
+
+#include <assert.h>
+
+template <size_t BS, size_t NB>
+class MxMemoryPool {
+public:
+	MxMemoryPool() : m_pool(NULL), m_blockSize(BS){};
```
foxtacles, line 13: Shouldn't have a semicolon at end of line

LEGO1/omni/include/mxmemorypool.h
```diff
@@ -0,0 +1,86 @@
+#ifndef MXMEMORYPOOL_H
```
foxtacles, line 1: Can we add annotations where possible, and decomp size asserts? Thank you!

foxtacles:
Very nice, glad to see this finally figured out!

LEGO1/omni/include/mxmemorypool.h
disinvite: The closest logical place for annotations would be mxstreamer.cpp. I had to typedef the memory pools because I guess the assert didn't like the double template. This look ok naming-wise?
```cpp
DECOMP_SIZE_ASSERT(MxMemoryPool64, 12);
DECOMP_SIZE_ASSERT(MxMemoryPool128, 12);
DECOMP_SIZE_ASSERT(MxBitset<22>, 4);
DECOMP_SIZE_ASSERT(MxBitset<2>, 4);
```

What annotations should I add? `// SIZE`? The two classes are always 12 and 4 bytes where used, but they both have variable size. The only two functions there to annotate are the two destructors on the memory pools.

foxtacles: Member offsets should be mostly possible. Also the Reference. Anything that is template dependent we can just ignore

LEGO1/omni/include/mxbitset.h
```diff
@@ -0,0 +1,94 @@
...
+
+	void Xran() { assert("invalid MxBitset<N> position" == NULL); }
+
+	// Not a real enum. This is how STL BITSET defines these constants.
+	enum {
+		e_bitsPerBlock = CHAR_BIT * sizeof(MxU32),
+		e_blocksRequired = N == 0 ? 0 : (N - 1) / e_bitsPerBlock
+	};
+
+	MxU32 m_blocks[e_blocksRequired + 1]; // 0x0
```
foxtacles, line 91: minor nit but should be 0x00

LEGO1/omni/include/mxmemorypool.h
foxtacles: naming looks good, for the size values we should use hex instead of decimal

disinvite: Ah yes, I missed those

LEGO1/omni/src/stream/mxstreamer.cpp
```diff
@@ -8,6 +8,10 @@
 #include <algorithm>
 
 DECOMP_SIZE_ASSERT(MxStreamer, 0x2c);
+DECOMP_SIZE_ASSERT(MxMemoryPool64, 0xc);
+DECOMP_SIZE_ASSERT(MxMemoryPool128, 0xc);
```
foxtacles, line 12: generally want to pad the hex values to 2 digits

LEGO1/omni/include/mxbitset.h
```diff
@@ -13,6 +13,7 @@ class MxBitset {
 public:
 	MxBitset() { Tidy(); }
 
+	// SIZE 0x8
```
foxtacles, line 16: 0x08

disinvite: haha

disinvite: My next script should be the tool that catches all these things

PR #673, opened 2024-03-14
Implement methods to load dta files.

tahg:
Start to implement LegoAnimationManager.  LoadScriptInfo  is more correct than it seems, but a large block is out of place.


LEGO1/lego/legoomni/include/animstate.h
```diff
@@ -3,6 +3,33 @@
...
+};
+
+struct AnimInfo {
+	char* m_animName;       // 0x00
+	undefined4 m_unk0x4;    // 0x04
+	MxS16 m_unk0x8;         // 0x08
+	MxU8 m_unk0xa;          // 0x0a
+	MxU8 m_unk0xb;          // 0x0b
+	MxU8 m_unk0xc;          // 0x0c
+	MxU8 m_unk0xd;          // 0x0d
```
Ramen2X, line 23: Small styling nitpick: We traditionally pad singular hex values with a `0`

LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
```diff
@@ -1,7 +1,17 @@
...
+#include "legoomni.h"
+#include "misc.h"
+#include "mxutilities.h"
+
+#include <io.h>
+
 DECOMP_SIZE_ASSERT(LegoAnimationManager, 0x500)
 
+// GLOBAL: LEGO1 0x100f7048
+Character g_characters[0x2f];
```
Ramen2X, line 13: Might be cleaner to represent the declaration of the size in decimal?

LEGO1/lego/legoomni/include/legoanimationmanager.h
```diff
@@ -1,9 +1,69 @@
...
+	char* m_name;        // 0x00
+	undefined4 m_unk0x4; // 0x04
+};
+
+void C100d8ca8_Handler();
+void C100d8cd8_Handler();
+
+// VTABLE: LEGO1 0x100d8ca8
+// SIZE 0x10
+class C100d8ca8 : public MxCore {
```
Ramen2X, line 27: Unknown classes and structs are typically formatted as `Unk` and then the file offset; or if they're only used a member, `Unknown` and then their relative member offset.

foxtacles: Omni-based global unknown classes should be named `LegoUnknown1000dead`

LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
tahg: Ultimately it's a preinitialized array, but I changed it for now.

foxtacles:
The entirety of those unknown classes and related assets are a list based on `MxPtrList`.

I suggest making a copy of one of the previously created lists as a template, such as: https://github.com/isledecomp/isle/blob/master/LEGO1/lego/legoomni/include/legoroilist.h then removing all the addresses and filling them with the right ones of the new list. Be careful with all the annotations, make sure you get the right addresses for the right dtors etc.

Of course the prerequisite here would be to find out what the list element is. Since it's `MxPtrList` based it's obviously a pointer, but you can probably figure out the element type and then fully build the list based on that.

PR #701, opened 2024-03-19
More features for reccmp HTML UI

disinvite:
I've been working on adding more stuff to the `reccmp` HTML display.

- The first and most fundamental change is pagination. We had been showing the many thousands of records for `LEGO1` on a single page, but the browser really starts chugging when you need to sort these rows or redraw the full list after clearing your search filter. I set the page size to 200 and this seems to perform well on my ~potato~ computer. To mitigate the loss of how convenient it is to have everything on the page at once, the page select dropdown acts as a kind of index for the full dataset based on the sorting column.

- We now have the option to display the recomp address alongside the original. (attn @tahg)

- The `onclick` handlers for each function row work a bit differently now. I've found myself wanting to copy the address of a particular function to jump there in Ghidra, but clicking anywhere in the row opens or closes the diff display drawer below it. To stop that happening, I changed the row so that clicking the original or recomp address will copy it to your clipboard instead. Clicking the _name_ of the function (or anywhere near it) toggles the diff display.

- There is a new column that shows the number of diffs for the given function. You can't sort the table by this column yet but that's not difficult to add. I'm not entirely convinced of how useful this is but it does demonstrate that a (long) function close to matching can still differ in a lot of small ways.

- All clickables have the `cursor: pointer;` CSS directive. Items like column headers and function addresses and names have a `:hover` styling to underline the text to show that you can click here where it is not necessarily obvious.

- I removed the column borders in the listing table and it seems to flow better visually.

Everything seems to work ok on Chrome/Firefox/Safari.

### DIFF
--- a/tools/reccmp/reccmp.js
+++ b/tools/reccmp/reccmp.js
@@ -61,6 +61,46 @@ function formatAsm(entries, addrOption) {
   return output;
 }
 
+// Special internal values to ensure this sort order for matching column:
+// 1. Stub
+// 2. Any match percentage [0.0, 1.0)
+// 3. Effective match
+// 4. Actual 100% match
+function matchingColAdjustment(row) {
+  if ('stub' in row) {
+    return -1;
+  }
+
+  if ('effective' in row) {
+    return 1.0;
+  }
+
+  if (row.matching === 1.0) {
+    return 1000;
+  }
+
+  return row.matching;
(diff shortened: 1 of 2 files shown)


foxtacles:
Looks good to me!

@tahg - if you don't have any feedback I think this is good to merge.

PR #704, opened 2024-03-20
implement Hospital::HandleEndAction()

Ramen2X:
Matching except for the jump table and the `movsx` instructions the compiler generated in the original function when doing comparisons against `m_hospitalState`'s MxS16 members. It might be something super simple I'm missing, but I couldn't figure out how to generate the sign extension without a super ugly/convoluted cast on every check that I'm almost certain the developers didn't write.

https://github.com/isledecomp/isle/assets/64166386/d5ecba91-a0a1-4cc5-843f-232aa68da3d0

With this function implemented, most of the hospital's mission introduction sequence logic is in place, and the flickering LEDs animate.

### DIFF
--- a/LEGO1/lego/legoomni/src/worlds/hospital.cpp
+++ b/LEGO1/lego/legoomni/src/worlds/hospital.cpp
@@ -1,5 +1,6 @@
 #include "hospital.h"
 
+#include "act1state.h"
 #include "hospital_actions.h"
 #include "islepathactor.h"
 #include "jukebox.h"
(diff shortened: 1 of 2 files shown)


LEGO1/lego/legoomni/src/worlds/hospital.cpp
```diff
@@ -228,11 +230,149 @@ MxLong Hospital::HandleKeyPress(MxS8 p_key)
...
+	case 7:
+	case 10:
+		m_hospitalState->m_unk0x08.m_unk0x00 = 8;
+		m_unk0x100 = 1;
+		m_time = Timer()->GetTime();
+		break;
+	case 11:
+		switch (m_currentActorId) {
+		case 1:
+			if (0 > m_hospitalState->m_unk0x0e || 1 >= m_hospitalState->m_unk0x0e) {
```
foxtacles, line 269: The conditions in all of those branches are incorrect. If you look closely at the assembly you will notice the jump offsets don't match the original

foxtacles (approved):

PR #712, opened 2024-03-23
Animmanager

tahg:
Implement/match StartEntityAction

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
+++ b/LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
@@ -358,18 +358,23 @@ MxResult LegoAnimationManager::StartEntityAction(MxDSAction& p_dsAction, LegoEnt
 {
 	MxResult result = FAILURE;
 	LegoROI* roi = p_entity->GetROI();
+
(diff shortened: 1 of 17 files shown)


LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
```diff
@@ -1,13 +1,22 @@
...
 
 #include <io.h>
 
 DECOMP_SIZE_ASSERT(LegoAnimationManager, 0x500)
+DECOMP_SIZE_ASSERT(Character, 0x18)
+DECOMP_SIZE_ASSERT(Vehicle, 0x8)
+DECOMP_SIZE_ASSERT(Unknown0x3c, 0x18)
+
+// GLOBAL: LEGO1 0x100f6d10
+Vehicle g_vehicles[] = {"bikebd", 0,        FALSE, "bikepg", 0,        FALSE, "bikerd", 0,       FALSE, "bikesy", 0,
```
foxtacles, line 18: The global doesn't check out:

image

tahg:
Still a diff in g_characters, but I know about that. You should add datacmp.py to the workflow, I didn't know we even had that.

PR #732, opened 2024-03-26
reccmp: Identify jump tables / data tables

disinvite:
A non-trivial `switch` statement will produce a jump table, or a list of addresses for each switch case. This instruction will jump to an offset into the table; the index is the case value being checked:

```asm
jmp dword ptr [eax*0x4 + <jump_table_addr>]
```

If the case values are non-sequential, you might also see a data table used to fold the range of possible case values into only the specific switch cases that are present:

```asm
mov al, byte ptr [ecx + <data_table_addr>]
```

A good example of both is the function `Score::Notify` at `0x10001410` in LEGO1.

The problem for us is that our `capstone` library assumes everything in a given function is part of an instruction, so we get a lot of "junk" instructions where we have tried to disassemble part of an address or data value. This change adds a "pre-parse" step to our code sanitization that will identify these jump or data tables and handle them separately.

This has increased match percentage across the board because we are filtering out the noise from these junk instructions. (One function in `ISLE` has dropped because of a missing annotation on some data from SmartHeap.)

The next step is to add labels from both the switch cases and each local jump inside a function. This will make things a lot easier to read and allow us to meaningfully compare the switch statements. I left that out of this PR because deciding on how to name these labels (so that we have appropriate diffs) will be its own project.

_Most of the time_ the jump or data tables are pushed to the end of the function, after all instructions. This makes it easy to just read instructions and stop at the correct place. This is not the case in our Beta 1.0 build, where data can appear between two blocks of code. I had previewed this change on Matrix chat maybe a month ago, and what held this up until now is dealing with that situation.

An extreme edge case is the function `_SmackDoFrameToBuffer` from the smacker library at `0x100cda83`. Here the jump into the table of addresses appears _after_ the table itself, so it is too late for us to identify it. I'm leaving this one unsolved for now because comparing the smacker stuff isn't a priority and this situation is hopefully rare.


foxtacles:
Thank you, this will make working with the affected functions much easier!

PR #735, opened 2024-03-26
Implement ExtraActor vftable and passthrough methods

tahg:
100% vtable, function match.


LEGO1/lego/legoomni/src/paths/legoextraactor.cpp
```diff
@@ -7,10 +7,77 @@ LegoExtraActor::LegoExtraActor()
...
+}
+
+// STUB: LEGO1 0x1002b290
+void LegoExtraActor::VTable0x9c()
+{
+	// TODO
+}
+
+// STUB: LEGO1 0x1002b440
+void LegoExtraActor::VTable0x70(float p_float)
```
foxtacles, line 47: params shouldn't have a name unless used

PR #739, opened 2024-03-27
Implement LegoExtraActor::VTable0x90

tahg:
Implemented but not matched. Despite reccmp's really low rating of 45% it's not that bad. Some sort of unused stack variable that is in the original but not present in my version is throwing off offsets. Afaict it should exist between the two vectors. Unless it's a PACK issue that requires Vector3 to be on an 8 byte boundary?  There's some issues with inlining of vector functions too which is causing a few additional differences. The inlined matrix code seems to match exactly, barring the shifted stack pointer.

### DIFF
--- a/LEGO1/lego/legoomni/include/legocarraceactor.h
+++ b/LEGO1/lego/legoomni/include/legocarraceactor.h
@@ -27,12 +27,12 @@ public:
 		return !strcmp(p_name, LegoCarRaceActor::ClassName()) || LegoRaceActor::IsA(p_name);
 	}
 
-	void VTable0x6c() override;              // vtable+0x6c
-	void VTable0x70(float p_float) override; // vtable+0x70
-	MxS32 VTable0x90() override;             // vtable+0x90
-	MxS32 VTable0x94() override;             // vtable+0x94
-	void VTable0x98() override;              // vtable+0x98
-	void VTable0x9c() override;              // vtable+0x9c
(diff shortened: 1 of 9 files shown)


foxtacles:
Is there a beta version of this function?

tahg:
Yep, it's at 10080b4c

foxtacles:
I've fixed some of the matrix construction and calls. Now the code is equivalent except for the stack allocation issue, which I also don't have a solution for right now. I do have a suspicion though that it is because of this call:

https://github.com/isledecomp/isle/pull/739/files#diff-8de4a56fc8e3a31a0197d44c6ee42726c50f8f41fd2165dbaa1456c747e89d59R82

Normally you'd expect that it does always inline the call to `SubImpl`, but it doesn't. We have the same issue in several other places:

https://github.com/isledecomp/isle/blob/1cf60ccda65efeecb9abe895c2961a23849759ca/LEGO1/viewmanager/viewmanager.cpp#L397
https://github.com/isledecomp/isle/blob/master/LEGO1/lego/sources/roi/legoroi.cpp#L500

In all those cases you can coerce the compiler to not inline `Sub` by the reference cast as seen in those examples, however that is obviously unlikely to be original code. Whatever we are missing related to this could also be the cause for the stack allocation issue.

foxtacles:
Nevermind, found the issue. See latest commit above. Match at ~90% now. The issue with the call to `Sub` is still a thing though.

PR #743, opened 2024-03-28
Define g_characters

disinvite:
My trick for doing this was to use the `datacmp` utility after setting the value to this:

```cpp
// GLOBAL: LEGO1 0x100f7048
Character g_characters[47] = {0};
```

Then the tool reports what each struct offset should be, in the correct datatype, for any value that is non-zero.


Ramen2X (approved):

tahg:
Part of the reason this wasn't done yet, is the underlying structure is incomplete. That undefined[12] should be MxU8, MxU8, <align>, MxU32, MxU32

LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
```diff
@@ -19,7 +19,55 @@ Vehicle g_vehicles[] = {"bikebd", 0,        FALSE, "bikepg", 0,        FALSE, "b
 						FALSE,    "motoni", 0,     FALSE,    "motola", 0,     FALSE,    "board", 0,     FALSE};
 
 // GLOBAL: LEGO1 0x100f7048
-Character g_characters[47]; // TODO: Initialize this
+Character g_characters[47] = {
+	{"pepper", 0, 6, 0, FALSE, {0, 1, 0, 0, 220, 5, 0, 0, 32, 78, 0, 0}, FALSE},
```
ItzSwirlz, line 23: Trying to understand the values here.

This may be 3 colors expressed in terms of RGB:
`0, 1, 0,' is black
'0, 220, 5` is green
`0, 0, 32` is an extremely dark blue
`78, 0, 0` is a maroon like red

Ramen2X: This is incorrect; please read tahg's comment. He explains what is known about this structure so far. This data has nothing to do with colors; character colors are stored as textures in WORLD.WDB.

ItzSwirlz: Yeah, I spent about ten minutes looking to see if the values meant anything but it's definitely not color. So ignore what I said

PR #745, opened 2024-03-28
implement/match ReadNamedTexture()

Ramen2X:
100% match


LEGO1/lego/legoomni/src/common/legoutils.cpp
```diff
@@ -377,10 +377,30 @@ void SetLightPosition(MxU32)
...
+	texture = new LegoTexture();
+	if (texture != NULL) {
+		if (texture->Read(p_file, 0) != SUCCESS) {
+			delete texture;
+
+			return namedTexture;
+		}
+
+		namedTexture = new LegoNamedTexture(string.GetData(), texture);
+		if (namedTexture == NULL && texture != NULL) {
```
foxtacles, line 398: The `texture != NULL` here is not necessary

foxtacles (approved):

PR #748, opened 2024-03-29
implement/match LegoAnimationManager::FindVehicle()

Ramen2X:
100% match

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
+++ b/LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
@@ -294,10 +294,15 @@ done:
 	return result;
 }
 
-// STUB: LEGO1 0x10060140
-MxBool LegoAnimationManager::FUN_10060140(char* p_name, MxU32& p_index)
+// FUNCTION: LEGO1 0x10060140
+MxBool LegoAnimationManager::IsNotVehicle(char* p_name, MxU32& p_index)
 {
-	// TODO
+	for (MxS32 i = 0; i < sizeof(g_vehicles); i++) {
+		if (!strcmpi(p_name, g_vehicles[i].m_name)) {
+			p_index = i;
+			return TRUE;
+		}
+	}
 	return FALSE;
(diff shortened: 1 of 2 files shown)


LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
```diff
@@ -294,10 +294,15 @@ MxResult LegoAnimationManager::LoadScriptInfo(MxS32 p_scriptIndex)
...
+// FUNCTION: LEGO1 0x10060140
+MxBool LegoAnimationManager::IsNotVehicle(char* p_name, MxU32& p_index)
 {
-	// TODO
+	for (MxS32 i = 0; i < sizeof(g_vehicles); i++) {
+		if (!strcmpi(p_name, g_vehicles[i].m_name)) {
+			p_index = i;
+			return TRUE;
+		}
+	}
```
foxtacles, line 305: A blank line before the return statement would be good

LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
```diff
@@ -294,10 +294,15 @@ MxResult LegoAnimationManager::LoadScriptInfo(MxS32 p_scriptIndex)
 	return result;
 }
 
-// STUB: LEGO1 0x10060140
-MxBool LegoAnimationManager::FUN_10060140(char* p_name, MxU32& p_index)
+// FUNCTION: LEGO1 0x10060140
+MxBool LegoAnimationManager::IsNotVehicle(char* p_name, MxU32& p_index)
 {
-	// TODO
+	for (MxS32 i = 0; i < sizeof(g_vehicles); i++) {
```
foxtacles, line 300: `sizeof` is incorrect. It should be `_countof`. The fact that this matches to 100% despite that should be looked into, I'm assuming it has to do with the offset comparison past the structure end. @disinvite could you check why it passes?

foxtacles (approved):

PR #759, opened 2024-03-30
Implement/match LegoExtraActor::VTable0x94

tahg:
Build failing, but with workflow issues I can't fix.  VTable0x94 is close to matching, but one part is not inlining properly. (Part of vector's copy construction.)  I think all the jump offset differences are due to that but it's hard to say for sure.  A signature change on the function affected quite a few files.

The inlined function above is only called from VTable0x94, but beta says it was a separate function for some reason. Names are from there also.


LEGO1/lego/legoomni/src/paths/legoextraactor.cpp
```diff
@@ -1,7 +1,24 @@
...
 DECOMP_SIZE_ASSERT(LegoExtraActor, 0x1dc)
 
+// GLOBAL: LEGO1 0x100f31d0
+LegoWorld* g_unk0x100f31d0;
+
+// GLOBAL: LEGO1 0x100f31d4
+LegoAnimPresenter* m_assAnimP;
+
+// GLOBAL: LEGO1 0x100f31d8
+LegoAnimPresenter* m_disAnimP;
```
disinvite, line 17: These three variables just need to be initialized to NULL to pass the variable check

LEGO1/lego/legoomni/src/paths/legoextraactor.cpp
```diff
@@ -132,10 +149,106 @@ MxResult LegoExtraActor::FUN_1002aae0()
...
+				LegoROI* roi = m_roi;
+				SoundManager()->GetUnknown0x40()->FUN_1003dae0("crash5", (char*) m_roi->GetName(), FALSE);
+				VTable0xc4();
+				m_state = 0x102;
+				Mx3DPointFloat dir = p_actor->GetWorldDirection();
+				MxMatrix matrix3 = MxMatrix(roi->GetLocal2World());
+				Vector3 positionRef(matrix3[3]);
+				((Vector3&) positionRef).Add(&g_unk0x10104c18);
+				roi->FUN_100a58f0(matrix3);
+				float dot = dir.Dot(&dir, &Mx3DPointFloat(1.0f, 0, 0));
```
disinvite, line 236: Someone else can correct me but I think these two need something like this to pass the modern compiler step. (It doesn't like how we're using the temporary `Mx3DPointFloat` object.)

```cpp
#ifdef COMPAT_MODE
    float dot;
    {
        Mx3DPointFloat tmp(1.0f, 0, 0);
        dot = dir.Dot(&dir, &tmp);
    }
#else
    float dot = dir.Dot(&dir, &Mx3DPointFloat(1.0f, 0, 0));
#endif
```

And then a similar thing for `double dot2`.

PR #765, opened 2024-04-02
GasStation::ReadyWorld

ItzSwirlz:
Mainly long inlined fun with sticks poking out at random spots. I haven't been able to get it past 2/3%.


ItzSwirlz:
I've decided to take this PR and mark it ready for review. There are some ways I can increase the matching percentage, but.some of them are illogical, like creating an action without setting the atom id in some cases. It just doesn't feel logical. In order to keep progress moving, I think its best to barrel along.

PR #787, opened 2024-04-10
Implement LegoPathActor::VTable0x88

tahg:
More problems with vectors, sigh.  I'm not sure what's different about this function that it calls the Vector2 constructor. (but only for 4 of the 5?) Aside from that, the main thing affecting the match is there's an extra stack variable somewhere.  I'm pretty sure the compiler reused a stack offset in the original code that it's not doing for some reason now.  Other minor issues:

the return FAILURE is wrong, but does match how it is in beta (function at 0x100ae9da), so not sure what changed.

the call at line 113 should be EqualsCross at 0x80 (same as line 106) and then inlined to 0x74 according to the beta, but I couldn't figure out what cast would call that. The inline declaration at mxgeomtry3d.h line 41 seems to be masking the parent class method, even though the parameter types are different.  The direct call I have now does make the code match, but it's hacky.

Clean up LegoPathBoundary some and implement AddActor, 100% match.

### DIFF
--- a/LEGO1/lego/sources/geom/legoedge.cpp
+++ b/LEGO1/lego/sources/geom/legoedge.cpp
@@ -4,14 +4,6 @@
 
 DECOMP_SIZE_ASSERT(LegoEdge, 0x24)
 
-// TODO Based on the offset, this should be in the header, but as a stub it's getting inlined when there...
-// STUB: LEGO1 0x1002ddc0
-LegoResult LegoEdge::FUN_1002ddc0(LegoWEEdge* p_face, Vector3& p_point)
-{
-	// TODO
-	return SUCCESS;
-}
-
 // FUNCTION: LEGO1 0x1009a470
 LegoEdge::LegoEdge()
 {
--- a/LEGO1/lego/sources/geom/legoedge.cpp
+++ b/LEGO1/lego/sources/geom/legoedge.cpp
@@ -4,6 +4,14 @@
 
 DECOMP_SIZE_ASSERT(LegoEdge, 0x24)
 
+// TODO Based on the offset, this should be in the header, but as a stub it's getting inlined when there...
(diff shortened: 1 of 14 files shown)


LEGO1/lego/sources/geom/legoedge.cpp
```diff
@@ -4,6 +4,14 @@
 
 DECOMP_SIZE_ASSERT(LegoEdge, 0x24)
 
+// TODO Based on the offset, this should be in the header, but as a stub it's getting inlined when there...
+// STUB: LEGO1 0x1002ddc0
+LegoResult LegoEdge::FUN_1002ddc0(LegoWEEdge* p_face, Vector3& p_point)
```
foxtacles, line 9: In cases like this it's important to implement the inline function since it very often has a major effect on the code generation of the caller, to the point it's impossible to match the caller (`VTable0x88` in this case) without the right implementation. Creating a non-inline stub has a completely different impact on the code generation than a compiler generated materialization of an inline function

tahg: Implemented now, in the class header. This has not unexpectedly made the match on 0x88 worse.

foxtacles: That's fine - we do know for certain it is an inline function, the match will be figured out eventually. We should not implement things incorrectly just for the sake of the match.

foxtacles: Yeah, that is fine - since we know for certain it is an inline function it's better to implement it correctly now and figure out the match later if necessary. I'll take a look at it tomorrow and see if I can make it line up

foxtacles:
It matches ~94% now. The inlining of the concrete vectors at the beginning isn't quite right yet, I've seen this in several other places like `OrientableROI::OrientableROI`, but it's close enough.

PR #799, opened 2024-04-13
implement FUN_1003e050

MishaProductions:
This function is responsible for setting the camera position according to animation data. Something is wrong with it as the camera position is set in the middle of nowhere. Not sure what is wrong with it. Other functions match


foxtacles:
Matched to 100%. The issue you are describing might be because `LegoAnimNodeData::CreateLocalTransform` is incomplete, it still needs the implementation for `GetKeys`

foxtacles:
(and `Matrix4::FromQuaternion` might also be necessary)

PR #804, opened 2024-04-14
More effective match strategies

disinvite:
I made these changes to our effective match validation:

- Everything is now under the `find_effective_match` method so we can limit changes to `core.py`.
- I tried a new approach to register swapping. The way we have been doing this is through a "rolling patch" where we check each token of each instruction string in both original and recomp. If we find a diff and both tokens are registers, patch the current and all future occurrences of that register. This continues and patches new diffs that are created along the way. As it turns out, this algorithm is the same as replacing all registers with a placeholder string (of the correct size for the register used). Doing it that way passed a handful of new functions, too. I'm not sure why the previous approach didn't allow those because they looked like effective matches to me.
- Some functions require two different kinds of patching. For example, corrections to `cmp` operand order _and_ register swapping. To support this, each patch function now returns a set of the array indices in recomp that can be corrected. When we are done, after combining all the indices via set union, if we have resolved each line from recomp that is in conflict, we declare an effective match.
- On Matrix chat I mentioned that the register swap algorithm gave the green light to some diffs that are not effective matches. `0x10010f10` is the function in question; the problem is that two `push` instructions had a swapped register, but this is caused by a bug in the function, not MSVC compiler noise. We are now checking for this specific condition and overruling the register swap for those instructions.
- We are now handling a new type of compiler noise: rearranged instructions. Many instructions can be moved to an arbitrary location, and this largely depends on whether the registers or memory values in question are modified between the start and end positions. Right now, for an instruction relocation to be permitted, we are checking whether any of its registers are the first operand in any of the intermediate instructions. For `call` instructions, we assume `eax` will be modified, though we have no way to confirm (from just the text) whether this would occur.

### DIFF
--- a/tools/isledecomp/isledecomp/compare/asm/fixes.py
+++ b/tools/isledecomp/isledecomp/compare/asm/fixes.py
@@ -1,6 +1,9 @@
-from difflib import SequenceMatcher
-from typing import List
+import re
+from typing import List, Tuple, Set
 
+DiffOpcode = Tuple[str, int, int, int, int]
+
+REG_FIND = re.compile(r"(?: |\[)(e?[a-d]x|e?[s,d]i|[a-d][l,h]|e?[b,s]p)")
 
 ALLOWED_JUMP_SWAPS = (
     ("ja", "jb"),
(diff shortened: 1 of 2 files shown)


foxtacles:
@disinvite would it be possible to add a switch to get the asm diff on the console using `--verbose` without the effective match logic?

PR #817, opened 2024-04-17
Beta matching MxOmniCreateFlags

disinvite:
The beta shows that this class uses the bitmask struct `FlagBitfield` from MxTypes.h. Similar to other flag-type classes.

I removed the extra setters because they don't appear to be used.

### DIFF
--- a/LEGO1/omni/include/mxomnicreateflags.h
+++ b/LEGO1/omni/include/mxomnicreateflags.h
@@ -7,36 +7,47 @@ class MxOmniCreateFlags {
 public:
 	MxOmniCreateFlags();
 
+	// FUNCTION: BETA10 0x10092b50
+	inline void CreateObjectFactory(MxBool p_enable) { this->m_flags1.m_bit0 = p_enable; }
+
+	// FUNCTION: BETA10 0x10092b90
+	inline void CreateTickleManager(MxBool p_enable) { this->m_flags1.m_bit2 = p_enable; }
+
+	// FUNCTION: BETA10 0x10092bd0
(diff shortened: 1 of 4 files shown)


LEGO1/omni/include/mxomnicreateflags.h
```diff
@@ -5,82 +5,42 @@
...
-	inline void CreateTimer(MxBool p_enable)
-	{
-		this->m_flags2 = (p_enable ? this->m_flags2 | c_createTimer : this->m_flags2 & ~c_createTimer);
-	}
-	inline void CreateStreamer(MxBool p_enable)
-	{
-		this->m_flags2 = (p_enable ? this->m_flags2 | c_createStreamer : this->m_flags2 & ~c_createStreamer);
-	}
+	// FUNCTION: BETA10 0x10130cd0
+	inline const MxBool CreateObjectFactory() const { return this->m_flags1.m_bit0; }
```
foxtacles, line 11: Can you put a blank line before the comment and after the implementation? This is how it's been done in most other places so far in terms of style. Thank you!

disinvite: Just realized that the setters probably come first according to the addresses, so I moved those up

LEGO1/omni/include/mxomnicreateparam.h
```diff
@@ -18,13 +19,15 @@ class MxOmniCreateParam : public MxParam {
 		MxOmniCreateFlags p_flags
 	);
 
+	// FUNCTION: BETA10 0x10092cb0
```
foxtacles, line 22: Same thing here

PR #855, opened 2024-04-26
Score class cleanup

foxtacles:
Fixes various bugs and improves matches. The cube is now rendered correctly and no longer crashes the game.


theOneTrueFunniBoi:
No idea why but the score cube still crashes for me. Using the generated pdb files and visual studio, I was able to view the following error:
`Exception thrown at 0x100627FA (LEGO1.DLL) in TMA_LAUNCHER.EXE: 0xC0000005: Access violation reading location 0x00000018.`
This is the farthest I've gotten because I've got no idea how to trace the memory address to actual code, or use a proper debugger that actually works for that matter.

Interestingly enough, it does this when no save is loaded, or my dev test save is loaded. If I create a new save, the score cube room works flawlessly, and I've got no idea why.

foxtacles:
The reason for this is that the state classes have not been implemented yet. In particular, the current crash happens because `PizzaMissionState` isn't implemented. It needs to initialize its members in the constructor at `0x10039030` but none of this has been decompiled yet.

It works when the state object hasn't been created at all; in this case there is a fallback in the Score cube code that makes it work, but in all other cases it will crash.

PR #865, opened 2024-04-28
Refactor: Replace magic values by enum values

jonschz:
I found some magic values and replaced them by enums. Ideally, I wanted to have a version of `LegoWorld::Find()` that only accepts enums, but I found no elegant way to do so. The best I could come up with was
```c++
// legoworld.h
	inline MxCore* Find(const MxAtomId& p_atom, IsleScript::Script p_entityId) { return Find(p_atom, (MxS32) p_entityId); }
	inline MxCore* Find(const MxAtomId& p_atom, CopterScript::Script p_entityId) { return Find(p_atom, (MxS32) p_entityId); }
	inline MxCore* Find(const MxAtomId& p_atom, DunecarScript::Script p_entityId) { return Find(p_atom, (MxS32) p_entityId); }
	inline MxCore* Find(const MxAtomId& p_atom, JetskiScript::Script p_entityId) { return Find(p_atom, (MxS32) p_entityId); }
	inline MxCore* Find(const MxAtomId& p_atom, RacecarScript::Script p_entityId) { return Find(p_atom, (MxS32) p_entityId); }
	inline MxCore* Find(const MxAtomId& p_atom, InfomainScript::Script p_entityId) { return Find(p_atom, (MxS32) p_entityId); }

	// Only use this one when calling from other functions
	inline MxCore* FindById(const MxAtomId& p_atom, MxS32 p_entityId) { return Find(p_atom, p_entityId); }
private:
	MxCore* Find(const MxAtomId& p_atom, MxS32 p_entityId);
```
but that had a surprisingly wide range of effects (accuracy increased for 57 functions, decreased for 53 functions, overall reduction from 97.03 % to 96.97 %). For now, I think we should apply this fix and nothing else.

### DIFF
--- a/LEGO1/lego/legoomni/src/actors/radio.cpp
+++ b/LEGO1/lego/legoomni/src/actors/radio.cpp
@@ -80,7 +80,7 @@ void Radio::Stop()
 	if (m_state->IsActive()) {
 		LegoWorld* world = CurrentWorld();
 
-		MxControlPresenter* presenter = (MxControlPresenter*) world->Find(world->GetAtom(), 18);
+		MxControlPresenter* presenter = (MxControlPresenter*) world->Find(world->GetAtom(), IsleScript::c_Radio_Ctl);
 
 		if (presenter) {
 			presenter->VTable0x6c(0);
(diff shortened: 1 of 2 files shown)


foxtacles:
Unfortunately the code generation of the compiler is sensitive to extra (or missing) inline functions so we aren't free to do things like that. There is a beta debug build of `LEGO1.DLL` that we reference to figure out which inline functions the developers used in the original>\.

PR #873, opened 2024-05-01
Implement the rest of the SkateBoard class

jonschz:
Still a draft, showing my progress here so we avoid working on the same code. @foxtacles thank you for the contributions! There wasn't a lot of overlap between your progress and mine just yet. I'm also open to collaborate if anyone wants to.

Current diff:
```
New (1):
0x10010270 - SkateBoard::FUN_10010270 (new -> 96.06%)

Increased (3):
0x100100e0 - SkateBoard::VTable0xcc (stub -> 94.84%)
0x100104f0 - SkateBoard::VTable0xd0 (stub -> 76.92%)
0x10010510 - SkateBoard::FUN_10010510 (stub -> 100.00%)

Decreased (2):
0x10030c10 - Isle::Notify (100.00% -> 71.76%)
0x10031820 - Isle::Enable (96.96% -> 95.74%)
```


foxtacles:
Hi! Looks good, I've made some changes:

* The "state" classes are a notable exception when it comes to getters/setters. It seems that most classes access most of the members of the state classes directly. This has been confirmed by the beta debug version of `LEGO1.DLL`. So in this case making `SkateBoard` a friend of `Act1State` is correct. There are still instances in the code where getters/setters are used; this code is probably wrong and needs to be updated eventually.
* If possible we should be using forward declarations, for example in `skateboard.h`. I've forward declared `Act1State` to avoid pulling in the header as a dependency.
* `FUN_10010270` takes a `MxBool`. It passes this parameter to `Enable` (vtable+0x54) of `c_SkatePizza_Bitmap`. A bitmap is generally of type `MxStillPresenter`.
* `m_unk0x160` declared as `MxBool`. In 99% of cases in this code base, when a single byte is only ever tested for zero or non-zero, it can be considered a boolean.
* The signatures of `FUN_10064670` and `FUN_10064740` were indeed incorrect. This is fairly common with unimplemented functions so it can be changed when necessary.
* I've added some beta annotations to functions I've found in the beta build.
* Replaced some magic numbers
* Added spacing to make code more readable

The functions should all be matching completely now.

One important thing to learn about this project is that the code generation of the MSVC 4.20 compiler is somewhat erratic. We call that "compiler randomness" or entropy. What it comes down to is essentially that changes in the code base, for instance in a header, can pseudo-randomly affect the assembly code generation of functions in compilation units that include this header, even if the change(s) are completely unrelated to those functions. For example, just by adding an extra (unused) inline function or another enum declaration in a header the assembly in some functions may wind up different and `reccmp` will report a reduced (or increased) accuracy for those.

This is the case for `Isle::Notify` and `Isle::Enable` and possibly others in this PR. Unfortunately, not much can be done about this since we don't know the exact mechanisms the compiler uses here.

Ultimately we have to ignore "regressions" due to this phenomenon as much as possible and only verify we don't have regressions in functions which may actually have been impacted by the changes we make.

jonschz:
Hey, thanks for the improvements and for merging! I wasn't expecting someone to pick it up this quickly, otherwise I might have rebased and cleaned it up a bit myself.

Your explanations are also very helpful, thank you! Maybe it'd be a good idea to copy part of your post to the README?

foxtacles:
Good Point, I'll be updating the contribution guide soon and explain the compiler situation there too. Thanks!

PR #878, opened 2024-05-02
Beta matching MxDSObject/MxDSAction

disinvite:
Some more corrections from the beta. The secondary goal here is to get more annotations down so it's easier to parse through the file and compare/match other stuff.

It seems like they used getters more consistently than setters. I had to change `MxDSObject` from private to protected so we can set `m_type` and other members directly. The constructors for all the MxDS* child classes do this, but I only marked the address for this PR. I wanted to limit the overall changes so we wouldn't have a messy diff, and that seems to have worked out. The only degrade is `DeserializeDSObjectDispatch` which is compiler noise.

Dead code: copy constructors for `MxDSObject` and `MxDSAction`. (maybe all of them?)

In the `MxDSAction` constructor, I have the calls to `Vector3::Fill` splayed out to match the beta, but if there's a better way to do this then we should take it.

We aren't checking for this (and probably shouldn't) but `MxDSAction::~MxDSAction` breaks the function order for the `BETA10` annotations. We are still correct for `LEGO1`. Maybe a hint that some of these functions belong in the header instead?

### DIFF
--- a/LEGO1/omni/include/mxdsaction.h
+++ b/LEGO1/omni/include/mxdsaction.h
@@ -8,6 +8,7 @@
 class MxOmni;
 
 // VTABLE: LEGO1 0x100dc098
+// VTABLE: BETA10 0x101c1b68
 // SIZE 0x94
 class MxDSAction : public MxDSObject {
 public:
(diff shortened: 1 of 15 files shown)


foxtacles:
Looks good, I've made just a few changes:

> In the MxDSAction constructor, I have the calls to Vector3::Fill splayed out to match the beta, but if there's a better way to do this then we should take it.

The actual signature used by `EqualsScalar` (and several others) has been found to take a const reference. This is as an odd design choice since it's pretty inefficient but it does yield the correct code (forcing constant values onto the stack so that a reference can be passed). I've renamed `EqualsScalar` to `Fill` since that's the name we found in the alpha versions.

Since we touched the classes, I've also removed the `this->` keyword everywhere since we don't do that anymore.

disinvite:
Thanks! Looks good in the beta. I'll start taking out `this->` as I go through these other modules.

PR #885, opened 2024-05-05
Start work on building save data

stravant:
* Add LegoBuildingData struct to LegoBuildingManager, naming mirrors LegoCharacterData in LegoCharacterManager.

* 100% match of many methods in LegoBuildingManager using the struct.

* Note: Some of the functions only show up as a 98% match because the end of `g_buildingData` coincides with the start of `g_characterData` (they are directly back to back in the original DLL) confusing the assembly diff checker into thinking the wrong label is being used. This could be fixed by restructuring the `LegoXManager` files but this would be best done after they're all implemented.


DimaLeon2000:
LGTM!

PR #890, opened 2024-05-09
LegoBuildingManager Round 2

stravant:
* 100% match most of the remaining methods, only Tickle and FUN_10030630 remain.

* The interesting finding is that the curious members of the LegoBuildingManager form a short embedded fixed-length array used to store info about current animation of buildings.

* I saw that you removed the SetY which I had added to MxMatrix. Agree that this method doesn't make sense on MxMatrix, however I've added it back to Matrix4. I see the pattern of setting / getting the Y component used in enough places that I doubt they were just hoping they remembered the subscript correctly every time. Let me know if you agree or still don't think it makes sense to include.

### DIFF
--- a/LEGO1/lego/legoomni/include/legobuildingmanager.h
+++ b/LEGO1/lego/legoomni/include/legobuildingmanager.h
@@ -21,26 +21,35 @@ struct LegoBuildingInfo {
 		c_bit4 = 0x08
 	};
 
-	LegoEntity* m_entity;             // 0x00
-	const char* m_hausName;           // 0x04
-	MxU32 m_cycle1;                   // 0x08
-	MxU32 m_cycle2;                   // 0x0c
-	MxU8 m_cycle3;                    // 0x10
-	MxS8 m_unk0x11;                   // 0x11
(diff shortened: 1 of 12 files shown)


foxtacles:
> * I saw that you removed the SetY which I had added to MxMatrix. Agree that this method doesn't make sense on MxMatrix, however I've added it back to Matrix4. I see the pattern of setting / getting the Y component used in enough places that I doubt they were just hoping they remembered the subscript correctly every time. Let me know if you agree or still don't think it makes sense to include.

While it certainly makes sense from an engineering standpoint, I'm certain they have not used such a function. This is essentially confirmed by the Beta build which reveals all the inline functions that existed. Only the subscript operator[] exists and it is used everywhere, so we need to follow what they did.

The last bit is actually critically important because apart from the authenticity we strive for, using the right inline functions/patterns is crucial to getting matching code (in medium to large sized functions especially). A good example of this is the TGL library. It essentially needs to be re-done because it is impossible to match many functions within it without adding the inline template indirections they used that lead to the implementation.

PR #898, opened 2024-05-09
Beta addrs for ViewLODList

disinvite:
Implemented `ViewLODListManager::Destroy`. It would return an `MxBool` if we used the Mx types here, so I have it as `char`.

I moved global `g_unk0x10101064` to a static member `ViewLODListManager::g_ROINameUID`. It is only ever incremented and only in `ViewLODListManager::Create` so it just seems to be a unique id for the ROI name.

There are a few _alpha_ addresses here along with the usual beta ones. The reason is to verify which asserts should be there.

The STL function `erase()` on the map of `ViewLODList` is called two different ways; one with a single iterator and one with two (i.e. start, end). The problem is that both functions have the name `map<char const *,ViewLODList *,ROINameComparator,allocator<ViewLODList *> >::erase` in the PDB output. There is no mangled symbol name to use that would distinguish the two, so we just match whatever one comes first (in recomp). The first one to be matched is removed from consideration when we try to match the second one, which is why this works at all. You could argue that we don't actually need all these functions to be marked, especially for the beta.

`ViewLODList::Dump` is not the usual case of dead code because this is a virtual method. But it is implemented nonetheless. I moved it out of the header file so we would not have to import `viewlod.h` over there. The spacing of the assert line numbers does kinda support the fact that something about this size belongs in the .cpp file.

### DIFF
--- a/LEGO1/viewmanager/viewlodlist.cpp
+++ b/LEGO1/viewmanager/viewlodlist.cpp
@@ -11,14 +11,31 @@ DECOMP_SIZE_ASSERT(LODList<ViewLOD>, 0x10)
 DECOMP_SIZE_ASSERT(ViewLODList, 0x18)
 
 // GLOBAL: LEGO1 0x10101064
-int g_unk0x10101064 = 0;
+// GLOBAL: BETA10 0x10205d08
+int ViewLODListManager::g_ROINameUID = 0;
+
+#ifdef _DEBUG
+// FUNCTION: BETA10 0x10178310
+inline void ViewLODList::Dump(void (*pTracer)(const char*, ...)) const
+{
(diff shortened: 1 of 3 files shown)


LEGO1/viewmanager/viewlodlist.cpp
```diff
@@ -106,7 +123,30 @@ ViewLODList* ViewLODListManager::Lookup(const ROIName& p_roiName) const
 	return pLODList;
 }
 
-// STUB: LEGO1 0x100a7680
-void ViewLODListManager::Destroy(ViewLODList* lodList)
+// FUNCTION: LEGO1 0x100a7680
+// FUNCTION: BETA10 0x1017886b
+char ViewLODListManager::Destroy(ViewLODList* lodList)
```
foxtacles, line 128: That would need to be `unsigned char` if modelled like `MxBool`

PR #905, opened 2024-05-11
Beta addrs for TglSurface and related

disinvite:
A few minor changes, no new diffs on the build. Just tying this off before I start looking at the other tgl stuff that _is_ different.


foxtacles:
Just a heads up, most of the TGL library needs to be refactored/rewritten to use the template pattern architecture found in the alpha code, this is actually important to get matches for retail. The beta should help with that a lot, but it will be a major refactor.

PR #909, opened 2024-05-11
Add Ghidra function import script

jonschz:
Hi everyone,

as discussed on Matrix, here is a work-in-progress for a script that auto-imports the function signatures from the source code into Ghidra. So far, it works with > 1500 functions, which I have imported on a copy of `LEGO1.DLL` in the shared Ghidra repository under _LEGO1.DLL (experimental new version)_. There are most likely still some issues to iron out, but the first results look very promising.

Here are some things it cannot do (yet):
- Create or modify structs / classes
- Do anything with stubs
- Anything other than LEGO1.DLL

I decided to go with Python 2 since that is what Ghidra supports out-of-the-box, but I am not sure if that was a good idea. The linter naturally complains about a number of things that could be done more elegantly in Python 3.

I am leaving this as a draft pull request for now so I can iron out some more issues. Feel free to try the code and give feedback, though!


foxtacles:
Nice! Maybe some code from the `isledecomp` library can be re-used, for the parsing/ identification of functions possibly - what do you think @disinvite?

disinvite:
This looks great! A lot more sophisticated than the `ImportSymbolsScript.py` that comes with Ghidra. The `isledecomp` package is for python 3, so if that's ultimately a blocker for integrating with Ghidra then we could always just serialize to a file. The `datacmp.py` script is probably the simplest demonstration for how to set up the `IsleCompare` object.

In a shared server does the script still run locally? We need the recomp binary and pdb to get most of the matches. The parser can sort of read the function name underneath the `// FUNCTION` marker but we throw this way to use the (better) name in the pdb. We also have the decorated/mangled names from the pdb that includes return type and parameters. If Ghidra can use those to populate the function args then that would be really useful, especially for the decompiler.

We can read type information from the pdb to populate structs/datatypes, but that's not integrated into the main database as the script will show.

foxtacles:
Yeah, makes sense to use the PDB / `isledecomp` as the primary source for the data. Once this is production ready it would probably also be a good idea to run the script after every push to master, so that the Ghidra repo stays updated automatically.

jonschz:
So the plan would be to import function signatures from the pdb and (if possible) structs / classes from the source code? Or can the latter be found in the pdb as well?

foxtacles:
> So the plan would be to import function signatures from the pdb and (if possible) structs / classes from the source code? Or can the latter be found in the pdb as well?

I'd say this would be the best approach. The `isledecomp` library already parses a lot of information from the PDB, so if we can use that it will make things easier. For example, you can use the library to get all the functions in the code base:

https://github.com/isledecomp/isle/blob/119ff93461126585da9bdc0990159c2d30dbafd0/tools/isledecomp/isledecomp/compare/core.py#L725

Type information is also present in the PDB, but I don't think it's parsed yet. But this can probably be added. Maybe this is something @disinvite and you can collaborate on over time.

disinvite:
We are parsing type info, but it's not as accessible as the other stuff for two reasons:
- it's outside of the main "compare database", as you see from `datacmp.py`
- there isn't a "get all types" method, but I'm working on that

I would guess that bootstrapping will be an issue for a struct that refers to others or to itself. Maybe the simplest thing is to do a first pass that creates any new structures (with just the name) so they at least exist in Ghidra's database, then add in the actual members for everything.

Also: Ghidra doesn't seem to like spaces in its labels (as in ``ClassName:`scalar deleting destructor'``) so we may want to decide on some scheme for reformatting the names so they are consistent. (I haven't read the code here top to bottom so maybe you have that solved already)

jonschz:
Thank you for your insights!

I think the easiest solution would be to
1. go over all types,
2. create those that only depend on types that are already present in Ghidra, and increment a counter on each creation,
3. repeat until the counter stagnates.

For that, it would be suffice to be able to read the name of each referenced struct. Only self-references need some extra care, but we can simply first create and then fill each struct to handle that.

This is not the most efficient solution, but it should be good enough - after all, performance is not our (main) concern here.

I will also look at Python 3 in Ghidra when I find the time (there is a compatibility layer which I have not tried yet).

jonschz:
Just to keep you in the loop (and to avoid duplication), I have made some progress on the PDB parsing, where a lot of information is indeed available but not yet parsed. If all goes well, I can upload some code the next few days.

jonschz:
I have just pushed my progress. Parsing and importing function signatures from the PDB mostly works. Note that some changes to other existing code was necessary. I have not run the test suite yet, but I have not seen any obvious regressions in `reccmp`.

There is one class of issues which I do not fully understand (all the functions mentioned below have 100 % match in the recompilation):

- `Score::ClassName` behaves as expected. It has the following symbols in the cvdump:
  ```
  (00010C) S_GPROC32: [0001:00055F80], Cb: 00000006, Type:             0x1077, Score::ClassName
  [...]
  (000144)  S_REGISTER: ecx, Type:             0x1076, this
  ```
  and the type `0x1077` looks like
  ```
  0x1077 : Length = 26, Leaf = 0x1009 LF_MFUNCTION
	  Return type = T_32PRCHAR(0470), Class type = 0x1072, This type = 0x1076,
	  Call type = ThisCall, Func attr = none
	  Parms = 0, Arg list type = 0x1023, This adjust = 0
  ```
- On the other hand, `MxDSMultiAction::SetAtomId` has the following symbols:
  ```
  (000F24) S_GPROC32: [0001:000887C0], Cb: 0000010A, Type:             0x23C0, MxDSMultiAction::SetAtomId
  [...]
  (000FE0)  S_BPREL32: [FFFFFFDC], Type:             0x23A5, this
  (000FF4)  S_BPREL32: [00000008], Type:             0x1210, p_atomId
  ```
  and the type `0x23C0` looks like
  ```
  0x23c0 : Length = 26, Leaf = 0x1009 LF_MFUNCTION
	  Return type = T_VOID(0003), Class type = 0x23A4, This type = 0x23A5,
	  Call type = ThisCall, Func attr = none
	  Parms = 1, Arg list type = 0x11f7, This adjust = 0
  ```

In the latter case I would have expected
- that the `this` parameter is in the `ecx` register, not on the stack,
- that the argument `p_atomId` is located at the stack offset `4`, not `8`.

Generating the function
```
void __thiscall MxDSMultiAction::SetAtomId(MxDSMultiAction *this,MxAtomId param0)
```
in Ghidra yields exactly the layout I expect. The main issue is that the names of the variables can only be found in the symbols, so matching the expected stack layout to the PDB symbols is rather important. This issue affects about 1 out of every 10 functions we have.

Thanks in advance for any input!

foxtacles:
Great work! I'll check it out in the coming days. @disinvite would you have time to review the changes as well?

disinvite:
Yes, definitely. But not much time to dig in and actually run it until the end of the week. I skimmed through it quickly earlier today.

One thing about the type hints: I don't know if we have any control over which python version runs in the github CI, but IIRC it's 3.9, so we don't have [PEP 604](https://peps.python.org/pep-0604/) that added the `int | None` syntax. I think you can get those to work by adding `from __future__ import annotations` but I opted to just use the older `Optional[int]` syntax. (We also need a `mypy` CI action at some point, but there's a lot of little problems to fix so that's a job for another day)

disinvite:
Without trying any of the new stuff (yet), here's why the current unit tests fail:

- Line 366 should be `return self.get(normalize_type_id(underlying_type))`. The test `test_enum` expects the type alias to be shortened to just "T_INT4" rather than "T_INT4(0074)". The new code now grabs the enum base type from the cvdump output rather than assuming int4.

- Line 359 was `size=obj.get("size")` to allow for the `LF_FIELDLIST` "type" to have a null size. The field list could be assumed to have the same size as the parent class/struct, but we'd have to match those together ahead of time. Does any of the new code need to look at the fieldlists independently from the parent?

If you could set up some simple tests like the ones in `test_cvdump_types.py` for the new `LF_` varieties that we're checking, that would be great!

jonschz:
Thanks to some help over at Ghidrathon I was able to do some refactoring and (hopefully) fixed the Python 3.9 support. Will look at new unit test cases and importing classes/structs when I find the time.

jonschz:
I have added basic functionality to import structs and classes. Basically, the code goes over all functions and imports the structs / classes recursively, overwriting existing ones.

This code is still nowhere near production ready, but if someone wants to tinker with it, feel free to. Most features should be implemented, and I get single-digit errors except for the one I explained above.

Speaking of this issue here: https://github.com/isledecomp/isle/pull/909#issuecomment-2120960643

> There is one class of issues which I do not fully understand (all the functions mentioned below have 100 % match in the recompilation):


Does anyone have any idea what is going on here? I'd appreciate someone else taking a look, because I don't have a good idea where to start. At the moment, the script can import ~2600 functions, with 273 more failing due to the aforementioned error.

foxtacles:
> Does anyone have any idea what is going on here? I'd appreciate someone else taking a look, because I don't have a good idea where to start. At the moment, the script can import ~2600 functions, with 273 more failing due to the aforementioned error.

I checked it out but can't make sense of it myself. We'll probably need someone who knows the PDB file format well enough and can explain how to interpret the data

jonschz:
I would consider the import mostly complete now, apart from the unsolved stack issue (which affects only a fraction of all functions). Happy to hear your thoughts. I have also applied the latest version to _LEGO1.DLL (experimental new version)_ in the shared repository.

tools/ghidra_scripts/lego_util/pdb_extraction.py
```diff
@@ -0,0 +1,161 @@
...
+            arglist=arg_list_pdb_types,
+            return_type=function_type["return_type"],
+            class_type=class_type,
+            stack_symbols=stack_symbols,
+        )
+
+    def get_function_list(self) -> list[tuple[MatchInfo, FunctionSignature]]:
+        handled = (
+            self.handle_matched_function(match)
+            for match in self.compare.db.get_matches_by_type(SymbolType.FUNCTION)
```
disinvite, line 122: Could use `self.compare.get_functions()` here

tools/ghidra_scripts/lego_util/pdb_extraction.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+        return [signature for signature in handled if signature is not None]
+
+    def handle_matched_function(
+        self, match_info: MatchInfo
+    ) -> Optional[tuple[MatchInfo, FunctionSignature]]:
+        assert match_info.orig_addr is not None
+        match_options = self.compare.db.get_match_options(match_info.orig_addr)
```
disinvite, line 130: I think it would be better to add another method (`should_skip(orig_addr)`?) to the compare API rather than expose the DB. Also: we can still get _something_ out of a stub method. It has an incomplete or non-existent implementation but we expect the original address to be correct.

tools/isledecomp/tests/test_cvdump_types.py
```diff
@@ -374,7 +374,7 @@ def test_2d_array(parser):
 def test_enum(parser):
     """LF_ENUM should equal 4-byte int"""
     assert parser.get("0x3cc2").size == 4
-    assert parser.get_scalars("0x3cc2") == [(0, None, "T_INT4")]
+    assert parser.get_scalars("0x3cc2") == [(0, None, "T_INT4(0074)")]
```
disinvite, line 377: My objection here would be that it's inconsistent with the other parsed items where we drop the number in parens

tools/isledecomp/isledecomp/types.py
```diff
@@ -1,4 +1,5 @@
 """Types shared by other modules"""
+
```
disinvite, line 2: We should see whether pylint can check for newlines after the docstring. I'm sure I/we were not consistent about this, but FWIW [PEP 257](https://peps.python.org/pep-0257/) rules against a blank line.

disinvite:
If I build `mxdsmultiaction.cpp.s`, it shows this for the stack:

```asm
;	COMDAT ?SetAtomId@MxDSMultiAction@@UAEXVMxAtomId@@@Z
_TEXT	SEGMENT
$T14029 = -16
__$EHRec$ = -12
_p_atomId$ = 8
_this$ = -36
_cursor$ = -32
_action$ = -16
?SetAtomId@MxDSMultiAction@@UAEXVMxAtomId@@@Z PROC NEAR	; MxDSMultiAction::SetAtomId, COMDAT
```

So maybe the explanation is that "this" points at the local copy unless only `ecx` is used?

For the 10% of functions that have the issue, are they all functions that use the exception handling? Maybe an extra `push` onto the stack explains the offset?

tools/isledecomp/isledecomp/types.py
jonschz: Quite the contrary, `black` actually adds that empty line back, at least the version `24.4.2` on Python 3.12 that I am using. To me, it looks like PEP 257 specifically discusses docstrings in functions.

Furthermore, `black` actually wants to add a lot of these blank lines to files I have not touched. I wanted to do submit this in a separate PR because this one is extremely large already.

LEGO1/lego/legoomni/include/legoworldlist.h
```diff
@@ -65,7 +65,7 @@ class LegoWorldListCursor : public MxPtrListCursor<LegoWorld> {
 // TEMPLATE: LEGO1 0x10059900
 // MxCollection<LegoWorld *>::~MxCollection<LegoWorld *>
 
-// TEMPLATE: LEGO1 0x10059950
+// TEMPLATE: LEGO1 0x10059947
```
jonschz, line 68: I would like to have a second opinion on this. I am a little puzzled by this address that is not divisible by 16, but the code looks like this should be the correct location.

tools/isledecomp/isledecomp/types.py
jonschz: On reviewing this again, it seems like I have commited this change by accident. Will revert it since this it belongs into the other PR.

tools/ghidra_scripts/lego_util/pdb_extraction.py
jonschz: I decided to forward `get_match_options()` in `Compare` since that allows us to distinguish `stub` and `skip` in the future. Let me know if that works for you. 

Stub handling sound like an interesting idea. While I would not trust the stub's signature, we could (at least) sync their addresses and names. Should we open a follow-up ticket for stubs? I'd prefer to get the code merged at some point instead of succumbing to eternal feature creep.

jonschz:
> So maybe the explanation is that "this" points at the local copy unless only `ecx` is used?
> 
> For the 10% of functions that have the issue, are they all functions that use the exception handling? Maybe an extra `push` onto the stack explains the offset?

I looked into it and unfortunately, I have not seen a simple pattern. What I can say for sure is:
- Ghidra and MSVC seem to have a different opinion on where the top of the stack is
- By Ghidra's conventions, the variable must be at stack offset `0x04`. Manually setting Ghidra's function layout to resemble the MSVC output is nonsensical. Consider the even simpler example `MxDSObject::SetAtomId()`.
  - This is the decompile with Ghidra's auto-generated stack layout: 
grafik
  - This is the decompile with `param0` set to `0x08` manually: 
grafik
     It is quite apparent that no stack parameter at 0x08 is read, but the one at 0x04 is passed to another function.

In order not to clutter this thread even further, should we create a follow-up ticket for this issue? I have a bunch of hypotheses and I'd like to further investigate possible destructors / exception handlers.

LEGO1/lego/legoomni/include/legoworldlist.h
disinvite: `0x10059950` is correct. It's a function with only a `ret` instruction, and it is referenced in the `MxCollection` constructor for that template type.

`0x10059947` is an exception handler unwind from the function above (starting at `0x10059900`). IIRC the shared ghidra had all the unwinds marked. Maybe someone had a script to do that? They are easy to find by reading all the [funcinfo](https://www.openrce.org/articles/full_view/21) structs

jonschz:
I might have solved the stack layout issue. A thorough investigation showed that
- all the affected functions had a `Flags: Frame Ptr Present` in their `S_GPROC32` block
- all of the functions that did *not* have a `Flags: Frame Ptr Present` worked correctly,
- all the working functions with `Flags: Frame Ptr Present` had no arguments.
I now added some code to simply shift the stack for all the functions with this flag by `-4`, which seems to have done the trick - I no longer get any mismatch errors at all.

foxtacles:
Awesome! I think once the failing CI test is fixed this should be good to merge. What do you think @disinvite?

tools/ghidra_scripts/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Script Directory
+- In Ghidra, _Open Window -> Script Manager_.
+- Click the _Manage Script Directories_ button on the top right.
+- Click the _Add_ (Plus icon) button and select this file's parent directory.
+- Close the window and click the _Refresh_ button.
+- This script should now be available under the folder _LEGO1_.
+
+## Development
+- Type hints for Ghira (optional): Download a recent release from https://github.com/VDOO-Connected-Trust/ghidra-pyi-generator,
```
disinvite, line 18: typo: Ghira -> Ghi**d**ra

disinvite:
I just got around to actually trying out the plugin. Very impressive!

Two things tripped me up, but just take these as observations if you prefer:
- The `virtualenv` requirement. I know it's python best practice but I don't always use it. Any way to skip the line `add_python_path(".venv/Lib/site-packages")` if we are not inside a venv?
- Having retail files in the `isle` root. Maybe I'm the oddball but I have these off in their own directory. Since Ghidra is the one driving `isledecomp` I don't know how you would specify where you want to load the files from. UI popup? How sophisticated is the Ghidra API?

I think we are good with the changes to the `isledecomp` module. Thanks for fixing those few things.

jonschz:
Thanks for the feedback!

> - The virtualenv requirement. I know it's python best practice but I don't always use it. Any way to skip the line add_python_path(".venv/Lib/site-packages") if we are not inside a venv?

The core issue is that we don't have a lot of control what Ghidra's Python environment does. In particular, we can't `pip install` anything within Ghidra's Python. My goal was to make Ghidra work somehow with the existing package structure and its external dependencies.

The line will not cause any harm if you are not using a `virtualenv`, since the directory will just be ignored if there is nothing there. However, I doubt that Ghidra will find external dependencies that have been installed globally, so I don't expect the code to work in that scenario (not tested, though). One could probably evaluate the `PYTHONPATH` environment variable, hard-code a Python version to use, and add an import for that directory as well - something like (pseudocode) `add_python_path("$PYTHONPATH/Python312/Lib")`. While I am fine with just saying "If you want to use the Ghidra script, you have to use the `virtualenv`", I also have no objections against putting this feature onto the wishlist.

> - Having retail files in the isle root. Maybe I'm the oddball but I have these off in their own directory. Since Ghidra is the one driving isledecomp I don't know how you would specify where you want to load the files from. UI popup? How sophisticated is the Ghidra API?

This is a good point. The `.gitignore` made me believe that this is their canonical location. I see a few ways out:
- Follow the _Convention over Configuration_ dogma and specify a canonical location for the retail files inside the repository. I saw that at least one CI script uses `/legobin`. You should be able to use file links to your local installation if you don't want to copy your files.
- Let's say you don't want to place the files into the repository at all (even if `.gitignore`d or linked). I think that Ghidra Script does support an _Open File_ dialog, so one could prompt the user for the location on first startup and then save this setting in a local `.gitignore`d file. To me, this does not feel that different from the first solution plus using file links, but let me know if that is preferable to you.

foxtacles:
Merging, - seems like additional changes/improvements would be better suited in their own PRs from now on. Thank you!

PR #961, opened 2024-05-28
Refactor MxBitmap (again)

disinvite:
Using the beta as our template for where functions should be inlined. This is a widely used class, so the minimal changes to the header created a big diff. The change to `GetDataSize` alone is responsible for a lot of that, though it is functionally equivalent.

- Removed `this->` to match other files where we've stopped using it.
- Refactored if-block "pyramids" to use `goto`. The beta supports doing this.
- We had been confused by the `Abs` function used here that does not precisely match the other `Abs` used everywhere else. The beta tells all: it is only used in combination with bitmap height, so it does appear to be a dedicated function just for this purpose.
- References to `sizeof(MxBITMAPINFO)` instead call a function that just returns the constant of 0x428 bytes. This could mean that there is _not_ a special struct for this and they just used the regular `BITMAPINFO`.

I left `BitBlt` and `BitBltTransparent` alone for now. The main reason is that we need to expand a lot of the convenience functions like `GetAdjustedStride`. I could not find a reference for this in the beta, but that doesn't mean it's definitely not there. Other files use these functions too, so we should take those on all at once and in a new PR.

### DIFF
--- a/LEGO1/omni/include/mxbitmap.h
+++ b/LEGO1/omni/include/mxbitmap.h
@@ -111,12 +111,8 @@ public:
 	inline MxBITMAPINFO* GetBitmapInfo() const { return m_info; }
 
 	// FUNCTION: BETA10 0x100982b0
-	inline MxLong GetDataSize() const
-	{
-		MxLong absHeight = GetBmiHeightAbs();
-		MxLong alignedWidth = AlignToFourByte(m_bmiHeader->biWidth);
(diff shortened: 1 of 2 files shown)


foxtacles:
Looks good! A few months ago I've also looked at `MxBitmap` - can confirm `GetAdjustedStride` doesn't seem to exist anywhere as a function in the BETA, but it still seems like it could have been a macro instead.

PR #990, opened 2024-06-04
Add more annotations to ISLE

foxtacles:
Also minor refactor of some code.

@disinvite there's only a few mismatches left in the MSVC library, mainly addresses as constants that are different and string constants. Not sure if there is a way to get them fixed. Also datacmp on `ISLE.EXE` will report a few diffs which seem to be due to stored pointers that have different addresses, but also this error which I'm not sure about:

```
?_pnhHeap@@3P6AHI@ZA (0x4125f8) ... DIFF
      0 (value)                        (initialized) : (uninitialized)
```

Also not sure why it can't find those two symbols:

```
[ERROR] Failed to find function symbol with name: _write_multi_char
[ERROR] Failed to find function symbol with name: _write_string
```


disinvite:
`_write_multi_char` and `_write_string` aren't named in the pdb, so not much we can do at the moment. They fall between these two symbols that _are_ named:

```
S_PUB32: [0001:00009580], Flags: 00000000, __output
S_PUB32: [0001:0000A040], Flags: 00000000, _raise
```

And the `SECTION CONTRIBUTIONS` data shows what comes between, with names added by me:

```
  sect.id  seg:offset      size     flags
  005D  0001:00009580  000009A2  60503020   ; __output
  005D  0001:00009F30  00000041  60503020   ; _write_char
  005D  0001:00009F80  00000031  60503020   ; _write_multi_char
  005D  0001:00009FC0  00000037  60503020   ; _write_string
  005D  0001:0000A000  0000000F  60503020   ; _get_int_arg
  005D  0001:0000A010  00000015  60503020   ; _get_int64_arg
  005D  0001:0000A030  00000010  60503020   ; _get_short_arg
  005C  0001:0000A040  000001DD  60503020   ; _raise
```

`_pnhHeap` is a real edge case. We have the bad luck of this being a variable set to 0 and addressed at the very end of the `.data` section. In orig, the four zero bytes are actually there in the file, but recomp tends to put this variable either partly or entirely on the virtual side of `.data`, so we say that this is uninitialized. In practice, a read for the virtual address in either case would come back as four zeroes.

I don't know if we have any control over where this variable is placed. It could be a byproduct of not having the exact same smartheap version as retail, or just that we have less data altogether.

In `datacmp` we can potentially detect the situation with `_pnhHeap` because the variable is right at the end of the section. It's just a matter of distinguishing between something that should be set to 0 or left uninitialized.

For all the other weird comparisons, we do it that way because we have no type information. We could add it via an annotation: maybe something similar to the [format characters](https://docs.python.org/3/library/struct.html#format-characters) for the python `struct` module.

foxtacles:
Okay thank you, makes sense! I will comment out `_pnhHeap` for now so that the build passes.

PR #1006, opened 2024-06-09
Ghidra import script - Ideas and future features

jonschz:
Using a PR since no other features are enabled on this repo. Based on #909.

# Feature wish list for the PDB -> Ghidra import script

These are roughly sorted from more to less important

- [ ] Run the import in GitHub Actions when new code is merged to `master` if possible (https://github.com/isledecomp/isle/pull/909#issuecomment-2106055112)
- [ ] Import names and addresses for stubs instead of ignoring them (https://github.com/isledecomp/isle/pull/909#discussion_r1621321918)
- [ ] Discuss: Support global installation of dependent Python packages (currently `virtualenv` is required - see https://github.com/isledecomp/isle/pull/909#issuecomment-2156281224)
- [ ] Fix: Some functions are incorrectly detected as having been changed (this is not a major issue, but it's still a little confusing: You see `Changed functions: 130` when `0` would be correct)
  - [ ] Reuse and verify all kinds of types instead of re-recreating them every time (e.g. Enums)
  - [ ] Correct difference detection for functions whose return types is larger than 4 bytes
    - These are _imported_ correctly, but Ghidra modifies them on creation, so the delta detection fails on the next run


foxtacles:
I've enabled issues for this repository. Could you transfer this information to individual tickets? thank you!

jonschz:
Done. Thanks!

PR #1047, opened 2024-06-27
Create MxDirectXInfo module

disinvite:
The beta shows a file `MxDirectXInfo.cpp` used for the video device enumeration/auto-detect functions. I moved these out of `mxdirect3d.cpp` and `mxdirectdraw.cpp` into their own file and improved the accuracy.

There is a subclass of `MxDeviceEnumerate` that I have renamed to `LegoDeviceEnumerate`. The beta asserts point to a similar file `LegoDXInfo.cpp`, hence the name. I didn't create that file for this PR, and here's why: the `CONFIG` app uses both classes. I'm not sure on how best to set up the linking for this because it would disturb the Mx/LegoOmni separation of concerns. Maybe a second static library just for `LegoDeviceEnumerate`?


foxtacles:
There's a few examples where they broke the separation of concerns before, such as https://github.com/isledecomp/isle/blob/master/LEGO1/lego/sources/misc/legocontainer.h - I have a feeling these were "last minute" additions. I wouldn't be surprised if that's the case for the CONFIG application as well, so I think it's ok if we intentionally break this as well (with a comment)

PR #1050, opened 2024-06-30
Start implementing `LegoRaceCar`

jonschz:
The implementation here is quite the struggle due to the virtual inheritance used.

I get an address mismatch at `0x10012fb1`. This is because the compiler replaces the loop over `i` by a loop over the pointer into the static array, and checks whether we have reached the array's end. Of course, the static array has a different address in the recompile, so the mismatch is expected. Do we have some kind of marker / comment for this situation that `reccmp` would understand?

Furthermore, I am not entirely sure that I have annotated the strings correctly. A double check would be appreciated.

### DIFF
--- a/LEGO1/lego/legoomni/include/legoracecar.h
+++ b/LEGO1/lego/legoomni/include/legoracecar.h
@@ -4,37 +4,10 @@
 #include "legocarraceactor.h"
 #include "legoracemap.h"
 
+// SIZE 0x08
 struct EdgeReference {
-	char* m_name;
-	LegoPathBoundary* m_data;
-};
-
-// GLOBAL: LEGO1 0x100f0a20
-static EdgeReference g_edgeReferences[] = {
-	{// STRING: LEGO1 0x100f0a10
-	 "EDG03_772",
-	 NULL
-	},
-	{// STRING: LEGO1 0x100f0a04
-	 "EDG03_773",
-	 NULL
-	},
-	{// STRING: LEGO1 0x100f09f8
-	 "EDG03_774",
-	 NULL
-	},
-	{// STRING: LEGO1 0x100f09ec
-	 "EDG03_775",
-	 NULL
-	},
-	{// STRING: LEGO1 0x100f09e0
-	 "EDG03_776",
-	 NULL
-	},
-	{// STRING: LEGO1 0x100f09d4
-	 "EDG03_777",
-	 NULL
-	}
+	const char* m_name;       // 0x00
+	LegoPathBoundary* m_data; // 0x04
 };
 
 // VTABLE: LEGO1 0x100d58a0 LegoRaceActor
@@ -94,6 +67,12 @@ private:
 	LegoAnimActorStruct* m_unk0x74; // 0x74
 	LegoPathBoundary* m_unk0x78;    // 0x78
 	LegoPathBoundary* m_unk0x7c;    // 0x7c
+
+	// GLOBAL: LEGO1 0x100f0a20
+	static EdgeReference g_edgeReferences[];
+
+	// GLOBAL: LEGO1 0x100f0a50
+	static const EdgeReference* g_pEdgeReferences;
 };
 
 #endif // LEGORACECAR_H
--- a/LEGO1/lego/legoomni/include/legoracecar.h
+++ b/LEGO1/lego/legoomni/include/legoracecar.h
@@ -87,13 +87,13 @@ public:
 	// LegoRaceCar::`scalar deleting destructor'
 
 private:
-	undefined m_unk0x54;         // 0x54
-	undefined4 m_unk0x58;        // 0x58
-	Mx3DPointFloat m_unk0x5c;    // 0x5c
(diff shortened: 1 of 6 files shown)


LEGO1/lego/legoomni/include/legoracecar.h
```diff
@@ -4,6 +4,39 @@
 #include "legocarraceactor.h"
 #include "legoracemap.h"
 
+struct EdgeReference {
+	char* m_name;
+	LegoPathBoundary* m_data;
+};
+
+// GLOBAL: LEGO1 0x100f0a20
+static EdgeReference g_edgeReferences[] = {
```
ItzSwirlz, line 13: EDG03_772: This address 0x100f0a20 specifically points to 0x100f0a10, which contains EDG03_772 which I believe is correct.

EDG03_773 has another pointer at 0x100f0a28

LEGO1/lego/legoomni/include/legoracecar.h
```diff
@@ -4,6 +4,39 @@
 #include "legocarraceactor.h"
 #include "legoracemap.h"
 
+struct EdgeReference {
```
foxtacles, line 7: Please add a `// SIZE` comment and a `DECOMP_SIZE_ASSERT` in the `cpp`. Thank you!

LEGO1/lego/legoomni/include/legoracecar.h
```diff
@@ -4,6 +4,39 @@
...
+	char* m_name;
```
foxtacles, line 8: This should be `const char*`

foxtacles, line 8: Please add the relative offsets of the members

LEGO1/lego/legoomni/include/legoracecar.h
foxtacles, line 13: Data should not be defined in a header. Please move this to the compilation unit (`.cpp` file). You can add the declaration as a private member of the class.

LEGO1/lego/legoomni/src/race/legoracecar.cpp
```diff
@@ -18,42 +22,78 @@ LegoRaceCar::LegoRaceCar()
...
+// FUNCTION: LEGO1 0x10012e60
 void LegoRaceCar::SetWorldSpeed(MxFloat p_worldSpeed)
 {
-	// TODO
+	if (!m_userNavFlag) {
+		if (!LegoCarRaceActor::m_unk0x0c) {
+			m_maxLinearVel = p_worldSpeed;
+		}
+		LegoAnimActor::SetWorldSpeed(p_worldSpeed);
+		return;
```
foxtacles, line 45: It would be better to turn this branch into a `if..else` structure than using an early return here

foxtacles:
> Do we have some kind of marker / comment for this situation that `reccmp` would understand?

The only way to fix this is to define the data that immediately follows after `g_edgeReferences` in the original. Then `reccmp` can correlate the "past-the-end" pointer to whatever data we have defined there and verify that it is in the same position.

> Furthermore, I am not entirely sure that I have annotated the strings correctly. A double check would be appreciated.

Looks good to me!

LEGO1/lego/legoomni/include/legoracecar.h
jonschz: I did notice those other pointers. Could you make a suggestion how these should be annotated?

jonschz: Good point - I first thought that this is a global variable, but it looks like it is only accessed from `LegoRaceCar`.

jonschz:
> The only way to fix this is to define the data that immediately follows after `g_edgeReferences` in the original. Then `reccmp` can correlate the "past-the-end" pointer to whatever data we have defined there and verify that it is in the same position.

Unfortunately, that did not work - while `reccmp` did show me the name of the constant I declared instead of a memory address, it did not understand the pattern and match them up. I think it would only work if I could force the declared variable that came after the original array to be placed at the same relative location in the re-compile, which I could not do easily.

LEGO1/lego/legoomni/include/legoracecar.h
```diff
@@ -46,21 +52,27 @@ class LegoRaceCar : public LegoCarRaceActor, public LegoRaceMap {
...
-	undefined4 m_unk0x7c;     // 0x7c
+	undefined m_unk0x54;            // 0x54
+	undefined4 m_unk0x58;           // 0x58
+	Mx3DPointFloat m_unk0x5c;       // 0x5c
+	LegoAnimActorStruct* m_unk0x70; // 0x70
+	LegoAnimActorStruct* m_unk0x74; // 0x74
+	LegoPathBoundary* m_unk0x78;    // 0x78
+	LegoPathBoundary* m_unk0x7c;    // 0x7c
+
+	// GLOBAL: LEGO1 0x100f0a20
```
foxtacles, line 71: The annotations should be at the site of definition (in the .cpp file)

PR #1057, opened 2024-07-04
Refactor surrounding `MxBitmap::GetAdjustedStride`

disinvite:
The stride of a bitmap is the number of bytes to seek ahead to get to the next row of pixels. Each row of the bitmap is dword-aligned, so there are between 0 and 3 padding bytes added to the visible image width on each row.

The `MxBitmap::GetAdjustedStride` function takes into account that the image can be stored with its rows in reverse order:

>There are two varieties of [Device-Independent Bitmaps]:
>
>A bottom-up DIB, in which the origin lies at the lower-left corner.
>A top-down DIB, in which the origin lies at the upper-left corner.
>If the height of a DIB, as indicated by the Height member of the bitmap information header structure, is a positive value, it is a bottom-up DIB; if the height is a negative value, it is a top-down DIB. Top-down DIBs cannot be compressed.
>
> https://learn.microsoft.com/en-us/windows/win32/gdi/device-independent-bitmaps

Until now we have been using an inline function for `GetAdjustedStride`, but there is no match for this in the beta. The calculation always refers to the `IsTopDown` function that gets used 13 times: 6 in `MxBitmap` and 7 in `MxDisplaySurface`. Of these, only one has a slightly different order of the functions, so it made sense to use a `GetAdjustedStride` macro in place of the function call.

The affected functions are:
- `MxBitmap::BitBlt`
- `MxBitmap::BitBltTransparent`
- `MxDisplaySurface::FUN_100bb500` (changed to non-static)
- `MxDisplaySurface::VTable0x28`
- `MxDisplaySurface::VTable0x30`
- `MxDisplaySurface::VTable0x44`

The main refactoring was in the `MxDisplaySurface` functions. These are stubborn to match even in the beta, but as best I can tell we are doing the same work in each. I'm not sure what exactly to test in the game to verify this, but I had problems with both the opening lego movie and the Pepper "dude with the food" movie while working on this. Both are playing fine now.

### DIFF
--- a/LEGO1/omni/include/mxbitmap.h
+++ b/LEGO1/omni/include/mxbitmap.h
@@ -94,6 +94,7 @@ public:
 	// FUNCTION: BETA10 0x1002c690
 	static MxLong HeightAbs(MxLong p_value) { return p_value > 0 ? p_value : -p_value; }
 
+	// FUNCTION: BETA10 0x10142030
 	inline BITMAPINFOHEADER* GetBmiHeader() const { return m_bmiHeader; }
 
 	// FUNCTION: BETA10 0x1002c440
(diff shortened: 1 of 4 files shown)


foxtacles:
Looks good! Since we touched the file, could you remove the this-> keyword? Thank you!

PR #1063, opened 2024-07-06
Implement `LegoRaceCar::FUN_10012ff0()`, refactoring

jonschz:
I implemented `LegoRaceCar::FUN_10012ff0()` and added a number of annotations based on `BETA10`.

Some open questions / discussable points:
* Some local variables in `BETA10` have bad names like `a` and `r`. I have opted to rename them and add their original name as a comment.
* There is evidence that field `0x14` in `LegoAnim` is called `m_camAnim` (its getter is called `GetCamAnim()`, see BETA10 at `0x101f629c`).
  * This could suggest that the class `LegoAnimScene` should be called `LegoCameraAnim` or `LegoCameraAnimation` instead.
* I have added a few assertions from BETA10 that appear not to affect the build in its release config at all.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -373,7 +373,7 @@ add_library(lego1 SHARED
   LEGO1/lego/legoomni/src/race/jetskirace.cpp
   LEGO1/lego/legoomni/src/race/legorace.cpp
   LEGO1/lego/legoomni/src/race/legoraceactor.cpp
-  LEGO1/lego/legoomni/src/race/legoracecar.cpp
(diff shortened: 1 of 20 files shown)


jonschz:
(will fix the CI errors later today or tomorrow)

foxtacles:
> * Some local variables in `BETA10` have bad names like `a` and `r`. I have opted to rename them and add their original name as a comment.

Since we are striving for maximum accuracy we should be using the original names, even in cases they are objectively bad. Rest looks good!

PR #1064, opened 2024-07-07
Refactor file names based on BETA10

jonschz:
I looked over the `.cpp` and `.h` files references in BETA10. Here is the list of files that did not exist in our repository yet:

* Done:
  * `./Lego/LegoOmni/Src/Actors/Act3Actors.cpp`: Simple rename from `act3actor.cpp`
  * `./LegoDXInfo.cpp`: This one is harder. I moved two functions over there, but I don't believe the structure is correct with the implementation of `MxDeviceEnumerate` being . I guess there is some child class of `MxDeviceEnumerate` that does not overwrite any virtual functions and hence shares `MxDeviceEnumerate`'s vtable?
* Not done:
  * `./Lego/LegoOmni/Src/Actors/Act3Ammo.cpp` - I could not easily identify any of the BETA10 functions implemented in that file, though I did not search exhaustively. 
  * `./Lego/LegoOmni/Src/Race/LegoRaceSpecial.cpp` - I could not easily identify any of the BETA10 functions implemented in that file, though I did not search exhaustively. 
  * `./Lego/LegoOmni/Src/Worlds/Act2sup.cpp` - The functions I checked are not even stubbed, so I am leaning towards "none of the functions there are implemented". I have not checked exhaustively, though.
  * `./Lego/LegoOmni/Src/Video/LegoPhonemeRef.cpp` - There is only one function in this file, and it looks orphaned to me - I did not find a single location where that function is called, and it is not exported either.


disinvite:
For `LegoDXInfo.cpp`, the subclass is `LegoDeviceEnumerate` (arbitrary name). It's in the main `mxdirectxinfo.h` file now, but it should get moved to its own `LegoDXInfo.h` file. It's hard to know which functions belong to the subclass and which to `MxDeviceEnumerate`. There isn't any solid evidence except for where the Lego subclass constructor gets called and which functions are called using the object. Based on that, my plan in #1047 was to take everything from `ParseDeviceName` to the end and move it into the Lego subclass, but that's just an educated guess

jonschz:
I have decided to migrate the bare minimum to make everything consistent (i.e. the two functions with solid evidence are relocated and everything compiles). Feel free to commit further refactorings onto this branch.

foxtacles:
> I have decided to migrate the bare minimum to make everything consistent (i.e. the two functions with solid evidence are relocated and everything compiles). Feel free to commit further refactorings onto this branch.

The problem with the current approach is that it will be impossible to reconcile the locations of the functions in both compilation units with their relative locations in the retail binary. So basically we can't just extract some functions from the middle of a unit and put them in their own, because there's no way to get them back into their relative location. We have to slice all compilation units so that none of them interleave with respect to their function addresses. Like @disinvite suggested, we'd have to take *everything* from *ParseDeviceName* until the end for this to work

jonschz:
Thanks for the thorough explanation. I understand that there is a contiguous block of code in multiple sources which would no longer be contiguous in the recompile if we migrate some part from the middle, which also highly suggests that this entire block came from the same `.cpp` file. I copied the block over and tried to make sure that the order is correct now (a double check by a reviewer would not hurt, though).

I traced the block back in `LEGO1.DLL` and found the first interruption above `0x1009c4c0` at `MxDeviceEnumerate::BuildErrorString()`, so one could speculate that `LegoDeviceEnumerate` starts at that point (I did not test whether that refactor would compile, though). For now I think the current solution is good enough to merge.

PR #1065, opened 2024-07-09
Implement `LegoRaceCar::HandleSkeletonKicks` and dependents

jonschz:
Looking forward to your feedback as always!

### DIFF
--- a/LEGO1/lego/legoomni/include/legoracers.h
+++ b/LEGO1/lego/legoomni/include/legoracers.h
@@ -16,10 +16,10 @@ struct EdgeReference {
 
 // SIZE 0x10
 struct SkeletonKickPhase {
-	EdgeReference* m_edgeRef;
-	float m_lower;
-	float m_upper;
-	unsigned char m_userState;
+	EdgeReference* m_edgeRef;  // 0x00
+	float m_lower;             // 0x04
+	float m_upper;             // 0x08
+	unsigned char m_userState; // 0x0c
 };
 
 // VTABLE: LEGO1 0x100d58a0 LegoRaceActor
@@ -53,7 +53,7 @@ public:
 		return !strcmp(p_name, LegoRaceCar::ClassName()) || LegoCarRaceActor::IsA(p_name);
 	}
 
-	void ParseAction(char*) override;                  // vtable+0x20
+	void ParseAction(char* p_extra) override;          // vtable+0x20
 	void SetWorldSpeed(MxFloat p_worldSpeed) override; // vtable+0x30
 	MxU32 VTable0x6c(
 		LegoPathBoundary* p_boundary,
@@ -70,8 +70,8 @@ public:
 	MxResult VTable0x9c() override; // vtable+0x9c
 
 	virtual void SetMaxLinearVelocity(float p_maxLinearVelocity);
-	virtual void FUN_10012ff0(float);
-	virtual MxS32 HandleSkeletonKicks(float);
+	virtual void FUN_10012ff0(float p_param);
+	virtual MxS32 HandleSkeletonKicks(float p_param1);
 
 	// SYNTHETIC: LEGO1 0x10014240
 	// LegoRaceCar::`scalar deleting destructor'
--- a/LEGO1/lego/legoomni/include/legoracers.h
+++ b/LEGO1/lego/legoomni/include/legoracers.h
@@ -4,10 +4,22 @@
 #include "legocarraceactor.h"
 #include "legoracemap.h"
 
+#define LEGORACECAR_UNKNOWN_STATE 0
+#define LEGORACECAR_KICK1 2 // name guessed
+#define LEGORACECAR_KICK2 4 // name validated by BETA10 0x100cb659
+
 // SIZE 0x08
 struct EdgeReference {
 	const char* m_name;       // 0x00
 	LegoPathBoundary* m_data; // 0x04
+};
+
+// SIZE 0x10
+struct SkeletonKickPhase {
+	EdgeReference* m_edgeRef;
+	float m_lower;
+	float m_upper;
+	unsigned char m_userState;
 };
 
 // VTABLE: LEGO1 0x100d58a0 LegoRaceActor
@@ -59,7 +71,7 @@ public:
 
 	virtual void SetMaxLinearVelocity(float p_maxLinearVelocity);
 	virtual void FUN_10012ff0(float);
-	virtual MxBool FUN_10013130(float);
+	virtual MxS32 HandleSkeletonKicks(float);
 
 	// SYNTHETIC: LEGO1 0x10014240
 	// LegoRaceCar::`scalar deleting destructor'
@@ -74,7 +86,9 @@ private:
 	LegoPathBoundary* m_unk0x7c;    // 0x7c
 
 	static EdgeReference g_edgeReferences[];
-	static const EdgeReference* g_pEdgeReferences;
+	static const SkeletonKickPhase g_skeletonKickPhases[]; // TODO: better name
+
+	static const char* g_soundSke13;
 };
 
 #endif // LEGORACERS_H
--- a/LEGO1/lego/legoomni/include/legoracers.h
+++ b/LEGO1/lego/legoomni/include/legoracers.h
@@ -16,10 +16,10 @@ struct EdgeReference {
 
 // SIZE 0x10
 struct SkeletonKickPhase {
-	EdgeReference* m_edgeRef;  // 0x00
-	float m_lower;             // 0x04
-	float m_upper;             // 0x08
-	unsigned char m_userState; // 0x0c
+	EdgeReference* m_edgeRef; // 0x00
+	float m_lower;            // 0x04
+	float m_upper;            // 0x08
+	MxU8 m_userState;         // 0x0c
 };
 
 // VTABLE: LEGO1 0x100d58a0 LegoRaceActor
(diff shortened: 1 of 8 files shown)


jonschz:
I seem to be unable to get the array of structs accepted by `datacmp.py`. Is there something wrong with my annotations or is `datacmp` unable to handle this structure? According to a quick search, the source code on `master` contains arrays of strings, but no other type of relevant constant array.

disinvite:
> I seem to be unable to get the array of structs accepted by `datacmp.py`. Is there something wrong with my annotations or is `datacmp` unable to handle this structure? According to a quick search, the source code on `master` contains arrays of strings, but no other type of relevant constant array.

The annotations look correct to me. I think we just haven't run into this situation yet: a pointer into an array whose elements are not individually identified. Usually we can just check whether the pointers (in orig and recomp) point to the same thing and call it a day, but here there isn't a way to annotate each struct of `g_edgeReferences[]`. Maybe you could splay them out into individual variables, but I'm not certain it would work and it's not good code style.

We might have to do the comparison like this:
1. If the value is a pointer and we have it identified in both binaries, check whether they point at the same thing (i.e. current behavior)
2. If we fail to meet any of that criteria, compare the pointed-at-variable as its own data item.

The only pitfall with adding no. 2 is that you could wind up with redundant data, but `datacmp` isn't (currently) the best tool to handle that anyway.

jonschz:
I tried something different and added matches for all elements of a matched array. The code is more of a spike and not ready to be merged, but I'd still appreciate feedback if you think this idea is worth pursuing further.

jonschz:
I went ahead and cleaned up the code a little bit. Also, to see that it works correctly, here is the output when disabling this line: https://github.com/isledecomp/isle/blob/f6ce99fbe4693fe6e2040b17a9e858962448d218/tools/isledecomp/isledecomp/compare/core.py#L268
This causes the recomp addresses to be added but not matched to the original:
```
     16 [1].m_edgeRef                  Unknown pointer 0x100f0a28 : Pointer to LegoRaceCar::g_edgeReferences[1].m_name (DATA)
     32 [2].m_edgeRef                  Unknown pointer 0x100f0a30 : Pointer to LegoRaceCar::g_edgeReferences[2].m_name (DATA)
     48 [3].m_edgeRef                  Unknown pointer 0x100f0a30 : Pointer to LegoRaceCar::g_edgeReferences[2].m_name (DATA)
     64 [4].m_edgeRef                  Unknown pointer 0x100f0a28 : Pointer to LegoRaceCar::g_edgeReferences[1].m_name (DATA)
     96 [6].m_edgeRef                  Unknown pointer 0x100f0a38 : Pointer to LegoRaceCar::g_edgeReferences[3].m_name (DATA)
    112 [7].m_edgeRef                  Unknown pointer 0x100f0a40 : Pointer to LegoRaceCar::g_edgeReferences[4].m_name (DATA)
    128 [8].m_edgeRef                  Unknown pointer 0x100f0a48 : Pointer to LegoRaceCar::g_edgeReferences[5].m_name (DATA)
    144 [9].m_edgeRef                  Unknown pointer 0x100f0a48 : Pointer to LegoRaceCar::g_edgeReferences[5].m_name (DATA)
    160 [10].m_edgeRef                 Unknown pointer 0x100f0a40 : Pointer to LegoRaceCar::g_edgeReferences[4].m_name (DATA)
    176 [11].m_edgeRef                 Unknown pointer 0x100f0a38 : Pointer to LegoRaceCar::g_edgeReferences[3].m_name (DATA)
```

DimaLeon2000:
What is the match percentage?

jonschz:
back up to 100 % for data after the fix, and the newly implemented functions have 100 % as well.

DimaLeon2000:
LGTM!

foxtacles:
@disinvite could you take a look at the reccmp changes? Thank you!

disinvite:
Looks good! But what if we streamline `_check_if_array_and_match_elements` like this:
```python
# preceding code stays the same, but remove the helper function

variable_key = node.data_type.key.lower()

if not variable_key.startswith("0x"):
    # scalar type, so clearly not an array
    return

data_type = self.cv.types.get(variable_key)
if data_type is None:
    return

if data_type.members is None:
    return

# Ignore LF_STRUCTURE
if self.cv.types.keys[variable_key]["type"] != "LF_ARRAY":
    return

for member in data_type.members:
    self._db.set_recomp_symbol(
        recomp_addr + member.offset,
        None,
        f"{name}{member.name}",
        None,
        0,
    )
    self._db.set_pair(orig_addr + member.offset, recomp_addr + member.offset)
```

If the goal is just to set up the orig/recomp address pair for array elements, we can just use `None` for the symbol type and set size to 0.

jonschz:
Thanks for the review!
* The hard-coded size of 4 in the helper function is a leftover from an older version and incorrect indeed, good catch! Hard-coding to 0 sounds like a good idea, I'll try that. The actual size should also be accessible if needed.
* I'll double check your suggested code later, which would be a great simplificattion if it works correctly. Right now I have some doubts whether the `members` of an array of structs contains an entry for every field of every element. If it doesn't, the current example would only work by accident (because the referenced string is the first field in the struct).

jonschz:
As expected, I could not get rid of the helper function in its entirety. Nevertheless I could simplify the code significantly based on your suggestion.

LEGO1/lego/legoomni/include/legoracers.h
```diff
@@ -4,12 +4,24 @@
...
+
 // SIZE 0x08
 struct EdgeReference {
 	const char* m_name;       // 0x00
 	LegoPathBoundary* m_data; // 0x04
 };
 
+// SIZE 0x10
+struct SkeletonKickPhase {
+	EdgeReference* m_edgeRef;
```
foxtacles, line 19: Could you add relative offset annotations for the members? Thank you!

LEGO1/lego/legoomni/include/legoracers.h
```diff
@@ -59,7 +71,7 @@ class LegoRaceCar : public LegoCarRaceActor, public LegoRaceMap {
 
 	virtual void SetMaxLinearVelocity(float p_maxLinearVelocity);
 	virtual void FUN_10012ff0(float);
-	virtual MxBool FUN_10013130(float);
+	virtual MxS32 HandleSkeletonKicks(float);
```
foxtacles, line 74: Could you add the same parameter name as in the definition body?

LEGO1/lego/legoomni/include/raceskel.h
```diff
@@ -12,8 +12,10 @@ class RaceSkel : public LegoAnimActor {
 public:
 	RaceSkel();
 
+	void GetCurrentAnimData(float*, float*);
```
foxtacles, line 15: Could you add the same parameter name as in the definition body?

foxtacles (commented):
Looks good! @disinvite could you sign off on the reccmp changes? Thanks!

tools/isledecomp/isledecomp/compare/core.py
```diff
@@ -286,31 +287,26 @@ def _add_match_in_array(
...
+            assert node.data_type.members is not None
 
-            array_length, modulus = divmod(array_size_bytes, array_type_size)
-            assert modulus == 0
-
-            for i in range(array_length):
-                orig_element_base_addr = orig_addr + i * array_type_size
-                recomp_element_base_addr = recomp_addr + i * array_type_size
-                if array_type.members is None:
+            for i, array_element in enumerate(node.data_type.members):
```
disinvite, line 294: If this is an array, `member.name` should be `[offset]`. If so, you don't need the enumerate. You should update the f-strings too in case the name would ever get used by the sanitize code.

jonschz: Not sure I understand exactly what you mean. In the "array of scalars" case, the parameter `name` is something like `g_unk0x100f3218` and `array_element.name` is e.g. `[0]`. Do you still want to join the strings to `g_unk0x100f3218[0]`?

Similarly, if we have an array of structs, we find e.g. `name=g_actorLODs`, `array_element.name=[0]`, and `member.name=m_name`. Would `g_actorLODs[0].m_name` still be your desired outcome here? I am not sure I understand your point regarding the sanitize code.

In any case, I'll start by refactoring the code based on these suggestions.

disinvite: Yeah, both of those examples look correct to me. The connection with the sanitize code has to do with the fact that we are adding things to the program database. We have had functions that refer to an address that is not the start of a variable, and sanitize will want to find the name for whatever lives at that address. That name would now be `g_unk0x100f3218[1]` instead of `g_unk0x100f3218+4`, for example.

disinvite (approved):

LEGO1/lego/legoomni/include/legoracers.h
```diff
@@ -4,12 +4,24 @@
...
 	const char* m_name;       // 0x00
 	LegoPathBoundary* m_data; // 0x04
 };
 
+// SIZE 0x10
+struct SkeletonKickPhase {
+	EdgeReference* m_edgeRef;  // 0x00
+	float m_lower;             // 0x04
+	float m_upper;             // 0x08
+	unsigned char m_userState; // 0x0c
```
foxtacles, line 22: Should be `MxU8`

LEGO1/lego/legoomni/include/legoracers.h
```diff
@@ -58,8 +70,8 @@ class LegoRaceCar : public LegoCarRaceActor, public LegoRaceMap {
 	MxResult VTable0x9c() override; // vtable+0x9c
 
 	virtual void SetMaxLinearVelocity(float p_maxLinearVelocity);
-	virtual void FUN_10012ff0(float);
-	virtual MxBool FUN_10013130(float);
+	virtual void FUN_10012ff0(float p_param);
+	virtual MxS32 HandleSkeletonKicks(float p_param1);
```
foxtacles, line 74: For a 4-bit boolean it should `MxU32`. Should not make a difference in match

PR #1067, opened 2024-07-18
More legoracers functions

jonschz:
Matches to 100 %.

There is one minor detail where I'm unsure whether it is an issue or not:
```c++
// GLOBAL: LEGO1 0x100f0b10
// STRING: LEGO1 0x100f09cc
const char* LegoRaceCar::g_strSpeedCopy = "SPEED";
```
points at the same string as the following string in `define.cpp`:
```c++
// GLOBAL: LEGO1 0x101020b8
// STRING: LEGO1 0x100f09cc
const char* g_strSPEED = "SPEED";
```
However, I first tried
```c++
// GLOBAL: LEGO1 0x100f0b10
const char* LegoRaceCar::g_strSpeedCopy = g_strSPEED;
```
which led to an datacmp error `initialized : uninitialized`. Not sure if the latter syntax is not allowed in C++ or if datacmp does not correctly detect the latter string.

### DIFF
--- a/LEGO1/lego/legoomni/src/race/legoracers.cpp
+++ b/LEGO1/lego/legoomni/src/race/legoracers.cpp
@@ -65,7 +65,8 @@ const SkeletonKickPhase LegoRaceCar::g_skeletonKickPhases[] = {
 };
 
 // GLOBAL: LEGO1 0x100f0b10
-const char* LegoRaceCar::g_strSpeedCopy = g_strSPEED;
+// STRING: LEGO1 0x100f09cc
+const char* LegoRaceCar::g_strSpeedCopy = "SPEED";
 
 // GLOBAL: LEGO1 0x100f0b6c
 // STRING: LEGO1 0x100f08c4
@@ -81,7 +82,7 @@ int LegoRaceCar::g_unk0x100f0b88 = 0;
 
 // GLOBAL: LEGO1 0x100f0b8c
 // GLOBAL: BETA10 0x101f5f98
-MxBool LegoRaceCar::g_unk0x100f0b8c = FALSE;
+MxBool LegoRaceCar::g_unk0x100f0b8c = TRUE;
 
 // FUNCTION: LEGO1 0x10012950
 LegoRaceCar::LegoRaceCar()
--- a/LEGO1/lego/legoomni/src/race/legoracers.cpp
+++ b/LEGO1/lego/legoomni/src/race/legoracers.cpp
@@ -279,14 +279,14 @@ MxU32 LegoRaceCar::HandleSkeletonKicks(float p_param1)
 // FUNCTION: BETA10 0x100cb88a
 void LegoRaceCar::VTable0x70(float p_float)
 {
-	if (m_userNavFlag != 0 && (m_userState == LEGORACECAR_KICK1 || m_userState == LEGORACECAR_KICK2)) {
+	if (m_userNavFlag && (m_userState == LEGORACECAR_KICK1 || m_userState == LEGORACECAR_KICK2)) {
 		FUN_10012ff0(p_float);
 		return;
 	}
 
 	LegoCarRaceActor::VTable0x70(p_float);
 
-	if (m_userNavFlag != 0 && m_userState == LEGORACECAR_UNKNOWN_1) {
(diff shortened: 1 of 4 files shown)


LEGO1/lego/legoomni/src/race/legoracers.cpp
```diff
@@ -62,10 +64,26 @@ const SkeletonKickPhase LegoRaceCar::g_skeletonKickPhases[] = {
...
+// STRING: LEGO1 0x100f08c4
+const char* LegoRaceCar::g_srt001ra = "srt001ra";
+
 // GLOBAL: LEGO1 0x100f0b70
 // STRING: LEGO1 0x100f08bc
 const char* LegoRaceCar::g_soundSkel3 = "skel3";
 
+// GLOBAL: LEGO1 0x100f0b88
+// GLOBAL: BETA10 0x101f5f94
+int LegoRaceCar::g_unk0x100f0b88 = 0;
```
foxtacles, line 81: should be MxS32

LEGO1/lego/legoomni/src/race/legoracers.cpp
```diff
@@ -257,10 +275,61 @@ MxU32 LegoRaceCar::HandleSkeletonKicks(float p_param1)
 	return TRUE;
 }
 
-// STUB: LEGO1 0x100131f0
+// FUNCTION: LEGO1 0x100131f0
+// FUNCTION: BETA10 0x100cb88a
 void LegoRaceCar::VTable0x70(float p_float)
 {
-	// TODO
+	if (m_userNavFlag != 0 && (m_userState == LEGORACECAR_KICK1 || m_userState == LEGORACECAR_KICK2)) {
```
foxtacles, line 282: since this is a boolean, a boolean style check is more appropriate: `!m_userNavFlag` (same in other places)

LEGO1/lego/legoomni/src/race/legoracers.cpp
```diff
@@ -257,10 +275,61 @@ MxU32 LegoRaceCar::HandleSkeletonKicks(float p_param1)
...
+		float maximumSpeed = NavController()->GetMaxLinearVel();
+		char buffer[200];
+
+		sprintf(buffer, "%g", absoluteSpeed / maximumSpeed);
+
+		VariableTable()->SetVariable(g_strSpeedCopy, buffer);
+
+		if (m_sound) {
+			// pitches up the engine sound based on the velocity
+			if (0.83 * maximumSpeed < absoluteSpeed) {
```
foxtacles, line 313: it would be better/more readable to flip the operands and the operator so that constant values appear on the right side of the test - should not affect match

LEGO1/lego/legoomni/src/race/legoracers.cpp
```diff
@@ -257,10 +275,61 @@ MxU32 LegoRaceCar::HandleSkeletonKicks(float p_param1)
...
+				m_frequencyFactor = gearRpmFactor / 80.0 + 0.7;
+			}
+		}
+
+		if (absoluteSpeed != 0.0f) {
+			g_unk0x100f0b88 = p_float;
+			g_unk0x100f0b8c = FALSE;
+		}
+
+		if (5000.0f < p_float - g_unk0x100f0b88 && !g_unk0x100f0b8c) {
```
foxtacles, line 328: it would be better/more readable to flip the operands and the operator so that constant values appear on the right side of the test - should not affect match

LEGO1/lego/legoomni/src/race/legoracers.cpp
```diff
@@ -257,10 +275,61 @@ MxU32 LegoRaceCar::HandleSkeletonKicks(float p_param1)
...
+		VariableTable()->SetVariable(g_strSpeedCopy, buffer);
+
+		if (m_sound) {
+			// pitches up the engine sound based on the velocity
+			if (0.83 * maximumSpeed < absoluteSpeed) {
+				m_frequencyFactor = 1.9f;
+			}
+			else {
+				// this value seems to simulate RPM based on the gear
+				int gearRpmFactor = (int) (6.0 * absoluteSpeed) % 100;
```
foxtacles, line 318: should be MxS32

PR #1068, opened 2024-07-22
Implement/match `LegoRaceCar::VTable0x94`

jonschz:
Most of the remaining discrepancy is due to compiler entropy. I was unable to get the return statements to an exact match (the logic should be correct). Here is the remaining diff:
```
0x100135bd : mov ecx, dword ptr [eax + 4]
0x100135c0 : -mov eax, 0
0x100135c5 : -test dword ptr [ecx + ebp - 0x30], 0x7fffffff
0x100135cd : -jne 0x20
           : +test dword ptr [ecx + esi - 0x30], 0x7fffffff
           : +je 0xc
           : +xor eax, eax
0x100135cf : pop ebp
0x100135d0 : -mov eax, 0xffffffff
0x100135d5 : pop edi
```
As mentioned on Matrix, I encountered a static instance of an `Mx3DPointFloat`:
```
Mx3DPointFloat LegoRaceCar::g_vector020 = Mx3DPointFloat(0.0f, 2.0f, 0.0f);
```
It is being initialised by a dedicated function at `LEGO1 0x10012db0`, though I don't if there is an annotation for this case. From what I can tell, there is no stable debug symbol for this function in the recomp. Currently, it is named `$E49`:
```
(00009C) S_LPROC32: [0001:0004C3C0], Cb: 00000005, Type:             0x17E6, $E49
         Parent: 00000000, End: 000000C8, Next: 00000000
         Debug start: 00000000, Debug end: 00000005
```

Looking forward to your feedback!

### DIFF
--- a/LEGO1/lego/legoomni/include/legoracers.h
+++ b/LEGO1/lego/legoomni/include/legoracers.h
@@ -102,7 +102,7 @@ private:
 	static MxLong g_timeLastSoundPlayed;
 	static MxS32 g_unk0x100f0b88;
 	static MxBool g_unk0x100f0b8c;
-	static Mx3DPointFloat g_vector020;
+	static Mx3DPointFloat g_unk0x10102af0;
 };
 
 #endif // LEGORACERS_H
--- a/LEGO1/lego/legoomni/include/legoracers.h
+++ b/LEGO1/lego/legoomni/include/legoracers.h
@@ -102,7 +102,6 @@ private:
 	static MxLong g_timeLastSoundPlayed;
 	static MxS32 g_unk0x100f0b88;
 	static MxBool g_unk0x100f0b8c;
-	static MxLong g_unk0x100f3308;
 	static Mx3DPointFloat g_vector020;
 };
 
--- a/LEGO1/lego/legoomni/include/legoracers.h
+++ b/LEGO1/lego/legoomni/include/legoracers.h
@@ -88,11 +88,22 @@ private:
 
 	static EdgeReference g_edgeReferences[];
 	static const SkeletonKickPhase g_skeletonKickPhases[];
-	static const char* g_strSpeedCopy;
+	static const char* g_strSpeed;
+	static const char* g_srtsl18to29[];
+	static const char* g_srtsl6to10[];
+	static const char* g_emptySoundKeyList[];
+	static const char* g_srtrh[];
 	static const char* g_srt001ra;
 	static const char* g_soundSkel3;
+	static MxU32 g_srtsl18to29Index;
+	static MxU32 g_srtsl6to10Index;
+	static MxU32 g_emptySoundKeyListIndex;
+	static MxU32 g_srtrhIndex;
+	static MxLong g_timeLastSoundPlayed;
 	static MxS32 g_unk0x100f0b88;
 	static MxBool g_unk0x100f0b8c;
+	static MxLong g_unk0x100f3308;
+	static Mx3DPointFloat g_vector020;
 };
 
 #endif // LEGORACERS_H
--- a/LEGO1/lego/legoomni/src/race/legoracers.cpp
+++ b/LEGO1/lego/legoomni/src/race/legoracers.cpp
@@ -103,13 +103,13 @@ const char* LegoRaceCar::g_srt001ra = "srt001ra";
 const char* LegoRaceCar::g_soundSkel3 = "skel3";
 
 // GLOBAL: LEGO1 0x100f0b74
-undefined4 LegoRaceCar::g_srtsl18to29Index = 0;
+MxU32 LegoRaceCar::g_srtsl18to29Index = 0;
 
 // GLOBAL: LEGO1 0x100f0b78
-undefined4 LegoRaceCar::g_srtsl6to10Index = 0;
+MxU32 LegoRaceCar::g_srtsl6to10Index = 0;
 
 // GLOBAL: LEGO1 0x100f0b7c
-undefined4 LegoRaceCar::g_emptySoundKeyListIndex = 0;
+MxU32 LegoRaceCar::g_emptySoundKeyListIndex = 0;
 
 // GLOBAL: LEGO1 0x100f0b80
 MxU32 LegoRaceCar::g_srtrhIndex = 0;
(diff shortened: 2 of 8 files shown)


LEGO1/omni/include/mxtimer.h
```diff
@@ -20,10 +20,18 @@ class MxTimer : public MxCore {
 			return g_lastTimeTimerStarted;
 		}
 		else {
-			return g_lastTimeCalculated - this->m_startTime;
+			return GetTimeElapsed();
 		}
 	}
 
+	MxLong GetStartTime() { return m_startTime; }
```
foxtacles, line 27: These all appear to be unused? If yes, please remove

LEGO1/realtime/orientableroi.h
```diff
@@ -39,6 +39,7 @@ class OrientableROI : public ROI {
 	void FUN_100a58f0(const Matrix4& p_transform);
 	void FUN_100a5a30(const Vector3& p_world_velocity);
 
+	// FUNCTION: BETA10 0x1000fbf0
 	const Matrix4& GetLocal2World() const { return m_local2world; }
```
foxtacles, line 43: Generally if a function has an annotation it should have a blank line after the function body, even if it's a one-liner (several more instances in the PR)

LEGO1/lego/legoomni/include/legoracers.h
```diff
@@ -88,11 +88,21 @@ class LegoRaceCar : public LegoCarRaceActor, public LegoRaceMap {
...
 	static const char* g_srt001ra;
 	static const char* g_soundSkel3;
+	static MxU32 g_srtsl18to29Index;
+	static MxU32 g_srtsl6to10Index;
+	static MxU32 g_emptySoundKeyListIndex;
+	static MxU32 g_srtrhIndex;
+	static MxLong g_timeLastSoundPlayed;
 	static MxS32 g_unk0x100f0b88;
 	static MxBool g_unk0x100f0b8c;
+	static Mx3DPointFloat g_vector020;
```
foxtacles, line 105: I think the unk0x with address naming pattern would be a better fit for now

jonschz: According to Ghidra, this variable is not referenced anywhere else, so
1. it will always be equal to `(0.,2.,0.)`, and
2. we will not get a lot of additional insight into its purpose (other than a better understanding of what `VTable0x94` does semantically).
If you still feel like it should be changed, I can of course do that. Other suggestions for the name are also welcome.

LEGO1/omni/include/mxtimer.h
jonschz: Oh, my bad - these were left over from an earlier version where I had overlooked `GetTime()`.

LEGO1/lego/legoomni/include/legoracers.h
foxtacles: > (other than a better understanding of what VTable0x94 does semantically).

Yes, generally speaking we should wait until we have that understanding before we assign a "real" name to a variable. It seems very unlikely that the developers would have named this variable after its contents.

In the meantime, please consider that this is a collaborative project: if someone who is unfamiliar with this code runs into `g_vector020` they are not going to know what to make of that name. They are going to wonder what `vector020` means. Does it clearly and distinctively convey any information which cannot otherwise be trivially obtained by looking at its definition (which already immediately reveals its type and value)?

On the other hand, using the pattern for as-yet-to-be-named variables that we established for this project (`unk` followed by `0x...` address) is clear: it means we don't have a good name for it yet and we need to figure out more about it before we can name it.

I understand that while reverse engineering code, it can sometimes be helpful to give a variable a temporary name such as `vector020` because it makes it easier to work with it in your head while you write the code. But this is more of a "mental crutch" that is very individual to the person who is doing the reverse engineering and should not make it into committed code which will be shared with and read/interpreted by others.

LEGO1/lego/legoomni/src/race/legoracers.cpp
```diff
@@ -332,11 +376,80 @@ void LegoRaceCar::VTable0x70(float p_float)
...
+			Vector3(matr[3]).Add(g_vector020);
+			roi->FUN_100a58f0(matr);
+
+			p_actor->SetState(2);
+		}
+
+		if (m_userNavFlag) {
+			MxBool actorIsStuds = strcmpi(p_actor->GetROI()->GetName(), "studs") == 0;
+			MxBool actorIsRhoda = strcmpi(p_actor->GetROI()->GetName(), "rhoda") == 0;
+			MxTimer* timer = Timer();
```
foxtacles, line 403: This shouldn't be a variable. `Timer()->GetTime()` is how it's usually written

LEGO1/omni/include/mxtimer.h
```diff
@@ -20,14 +20,20 @@ class MxTimer : public MxCore {
...
 		}
 	}
 
 	// SYNTHETIC: LEGO1 0x100ae0d0
 	// MxTimer::`scalar deleting destructor'
 
 private:
+	// This function appears to be public in BETA10; this function may also be
+	// an older version of GetTime() instead of a private subroutine.
+	// None of this matters for the release build since these functions are inlined.
```
foxtacles, line 33: This isn't quite correct and something we have to be very careful about. The inlining patterns do matter and have an influence on the code the compiler generates. In other words, the fact that an extra inline function is now being called in `GetTime` may cause regressions in some (not necessarily all) instances where `GetTime` is called - and that's a lot of potential places since it's called in numerous functions.

If I revert `GetTime` to the way it was before, I actually receive a ~92% match on `LegoRaceCar::VTable0x94`. So I think it's more likely this function doesn't exist in retail anymore.

LEGO1/lego/legoomni/src/race/legoracers.cpp
jonschz: The version with the variable matches slightly better at `0x10013434`. I'll change it nevertheless.

LEGO1/omni/include/mxtimer.h
jonschz: Good catch! I used to have another error elsewhere that cancelled out the mismatch, so I didn't expect the incorrect stack layout to come from here. I added the BETA10 annotation. Furthermore, I added a comment that this function was likely changed between BETA10 and release.

PR #1069, opened 2024-07-27
Implement/match the rest of `LegoRaceCar`, BETA10 improvements

jonschz:
The implementation of `LegoRaceCar` should be completed with this PR.

I also added some additional BETA10 references and corrections, notably a number of variable names. Unfortunately, these re-introduced the entropy issue in `LegoRaceCar::VTable0x94` - it seems to be one of those functions that just goes back and forth with its register allocation.

### DIFF
--- a/LEGO1/lego/legoomni/src/race/legoracers.cpp
+++ b/LEGO1/lego/legoomni/src/race/legoracers.cpp
@@ -484,9 +484,9 @@ MxResult LegoRaceCar::VTable0x9c()
 
 	if (m_userNavFlag) {
 		result = LegoCarRaceActor::VTable0x9c();
-		MxS32 bVar2 = 0;
 
 		if (m_boundary) {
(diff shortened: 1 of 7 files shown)


LEGO1/lego/legoomni/src/race/legoracers.cpp
```diff
@@ -451,14 +476,45 @@ MxResult LegoRaceCar::VTable0x94(LegoPathActor* p_actor, MxBool p_bool)
...
 {
-	// TODO
-	return SUCCESS;
+	MxResult result;
+
+	if (m_userNavFlag) {
+		result = LegoCarRaceActor::VTable0x9c();
+
+		if (m_boundary) {
+			MxS32 bVar2 = 0;
```
jonschz, line 489: The compiler entropy/register allocation is fixed when moving this variable declaration outside of the `if` block, but the diff looks less correct.

PR #1070, opened 2024-07-28
Implement/match `LegoCarRaceActor::FUN_10080590`

jonschz:
The big new function only has one mismatching line that does not influence the semantics. The rest are refactors and quick wins.

I also found an assertion casting doubt on the signature of `LegoWEEdge::IsEqual` - I'd be curious to learn if anyone knows more about this.

### DIFF
--- a/LEGO1/lego/legoomni/include/legoactor.h
+++ b/LEGO1/lego/legoomni/include/legoactor.h
@@ -52,7 +52,7 @@ public:
 	virtual void VTable0x58(MxFloat p_unk0x70) { m_unk0x70 = p_unk0x70; } // vtable+0x58
 
 	// FUNCTION: LEGO1 0x10002cf0
-	virtual MxFloat GetUnk0x70() { return m_unk0x70; } // vtable+0x5c
+	virtual MxFloat VTable0x5c() { return m_unk0x70; } // vtable+0x5c
 
 	// FUNCTION: LEGO1 0x10002d00
 	virtual MxU8 GetActorId() { return m_actorId; } // vtable+0x60
--- a/LEGO1/lego/legoomni/include/legoactor.h
+++ b/LEGO1/lego/legoomni/include/legoactor.h
@@ -52,7 +52,7 @@ public:
 	virtual void VTable0x58(MxFloat p_unk0x70) { m_unk0x70 = p_unk0x70; } // vtable+0x58
 
 	// FUNCTION: LEGO1 0x10002cf0
-	virtual MxFloat VTable0x5c() { return m_unk0x70; } // vtable+0x5c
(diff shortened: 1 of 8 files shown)


LEGO1/lego/legoomni/include/legoactor.h
```diff
@@ -52,7 +52,7 @@ class LegoActor : public LegoEntity {
 	virtual void VTable0x58(MxFloat p_unk0x70) { m_unk0x70 = p_unk0x70; } // vtable+0x58
 
 	// FUNCTION: LEGO1 0x10002cf0
-	virtual MxFloat VTable0x5c() { return m_unk0x70; } // vtable+0x5c
+	virtual MxFloat GetUnk0x70() { return m_unk0x70; } // vtable+0x5c
```
foxtacles, line 55: The name should remain `VTable0x5c`. It's a different convention compared to regular functions, but for virtual functions we decided the naming pattern with the vtable offset is preferable

LEGO1/lego/legoomni/src/entity/legocarraceactor.cpp
```diff
@@ -27,20 +31,68 @@ LegoCarRaceActor::LegoCarRaceActor()
...
 }
 
-// STUB: LEGO1 0x10080590
-void LegoCarRaceActor::FUN_10080590(float)
+// FUNCTION: LEGO1 0x10080590
+// FUNCTION: BETA10 0x100cd8cf
+void LegoCarRaceActor::FUN_10080590(float p_float)
 {
+	MxFloat maxSpeed = m_maxLinearVel;
+	Mx3DPointFloat destEdgeUnknownVector = Mx3DPointFloat();
```
foxtacles, line 39: This should be written as `Mx3DPointFloat destEdgeUnknownVector;` - it's functionally equivalent but more conventional.

LEGO1/lego/legoomni/src/entity/legocarraceactor.cpp
```diff
@@ -27,20 +31,68 @@ LegoCarRaceActor::LegoCarRaceActor()
...
+	Mx3DPointFloat worldDirection = Mx3DPointFloat(m_roi->GetWorldDirection());
```
foxtacles, line 40: This should be written as `Mx3DPointFloat worldDirection(m_roi->GetWorldDirection());`

jonschz: Oddly enough, this is not equivalent - your suggestion produces a slightly worse match. Please let me know if you still want to change it. The mysteries of MSVC 4.2 ...

foxtacles: You are right, I guess the compiler doesn't optimize this code to be equivalent the way it should be in modern compilers. We'll leave it as-is then!

PR #1071, opened 2024-07-30
Ghidra Import: Support virtual inheritance

jonschz:
Here are some of the changes to improve the situation with virtual inhertiance (as best as possible, but with limitations).

For the sake of simplicity, let's say we have three classes A, B, C with `class B: virtual A` and `class C: B`.

* Multiple non-virtual inheritance is now supported (see e.g. `LegoRaceCar`).
* Parsing virtual base class information from the PDB is now supported (with a unit test).
* The reported class sizes in the PDB include "unallocated" space for the virtual base classes:
  * An instance of `B` consists of (in this order)
    * its vtable,
    * vboffset,
    * its own fields,
    * some "unallocated" space to account for an `A` (the exact offset is not specified in the PDB).
  * This causes problems for class `C`, which consists of (in this order)
    * The parts of `B` directly it inherits directly (vtable, vboffset, B's fields),
    * its own fields,
    * space to account for an `A`.
  * Hence, the entirety of `B` does not fit into `A` at offset 0, but only the "allocated" parts.
  * Therefore, a class `B_vbase_slim` is generated as the superclass of `C` which is identical to `B` but does not contain unallocated space.
* Virtual base offsets are now automatically generated whenever a type has a direct virtual superclass (class `B` in the above example).
* The virtual base offsets form their own `struct`. This allows Ghidra to _sometimes_ correctly infer type information.
  * Unfortunately, all the typed base offsets are offset by an additional `-4` — likely because the pointer to the virtual base offsets itself is located at offset `4` within class `B`, and the offsets in the virtual base offset struct are relative to that location.
  * I have used a rather unknown feature of Ghidra to fix this: It is possible to define a *typedef* of a pointer and then declare an offset on it. All of the virtual base pointers receive a default offset of `-4` which is **not** overwritten when re-running the script (so if it turns out to be wrong, it can be fixed by hand permanently).
  * I think I have also observed cases where different offsets need to be used depending on which function you are decompiling. This is another argument in favour of giving the user flexibility here.
* `This adjust` is now supported: Overriding virtual functions on virtual superclasses requires that the input argument is shifted by some value that is encoded in the `This adjust` value in the PDB dump. Functions with a non-zero value of `This adjust` now get their `this` argument type replaced by an offset typedef, similar to the case explained above.
  * This was especially annoying in the past, because Ghidra would always decompile code like `this[-1]->field_0x180`.

I will post some screenshots and examples later. Note that I have not yet commited a version of the shared repository after running this version (so we don't need to rollback in case changes are desired), but I am happy to do so if it facilitates the review.

### DIFF
--- a/tools/ghidra_scripts/lego_util/function_importer.py
+++ b/tools/ghidra_scripts/lego_util/function_importer.py
@@ -10,6 +10,12 @@ from ghidra.program.model.listing import Function, Parameter
 from ghidra.program.flatapi import FlatProgramAPI
 from ghidra.program.model.listing import ParameterImpl
 from ghidra.program.model.symbol import SourceType
+from ghidra.program.model.data import (
+    TypeDef,
+    TypedefDataType,
+    Pointer,
+    ComponentOffsetSettingsDefinition,
+)
 
 from lego_util.pdb_extraction import (
     PdbFunction,
@@ -17,12 +23,13 @@ from lego_util.pdb_extraction import (
     CppStackSymbol,
 )
 from lego_util.ghidra_helper import (
+    add_data_type_or_reuse_existing,
     get_or_add_pointer_type,
     get_ghidra_namespace,
     sanitize_name,
 )
 
-from lego_util.exceptions import StackOffsetMismatchError
+from lego_util.exceptions import StackOffsetMismatchError, Lego1Exception
 from lego_util.type_importer import PdbTypeImporter
 
 
@@ -106,19 +113,22 @@ class PdbFunctionImporter:
             )
             return_type_match = True
 
-        # match arguments: decide if thiscall or not
+        # match arguments: decide if thiscall or not, and whether the `this` type matches
         thiscall_matches = (
             self.signature.call_type == ghidra_function.getCallingConventionName()
         )
+
+        ghidra_params_without_this = list(ghidra_function.getParameters())
+
+        if thiscall_matches and self.signature.call_type == "__thiscall":
+            this_argument = ghidra_params_without_this.pop(0)
+            thiscall_matches = self._this_type_match(this_argument)
 
         if self.is_stub:
             # We do not import the argument list for stubs, so it should be excluded in matches
             args_match = True
         elif thiscall_matches:
-            if self.signature.call_type == "__thiscall":
-                args_match = self._matches_thiscall_parameters(ghidra_function)
-            else:
-                args_match = self._matches_non_thiscall_parameters(ghidra_function)
+            args_match = self._parameter_lists_match(ghidra_params_without_this)
         else:
             args_match = False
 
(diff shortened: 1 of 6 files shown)


jonschz:
### Virtual inheritance: Good example

See `0x10013130`. The decompile looks like:
```C++
  pLVar1 = (this->base).vbase_offset;
  if (((*(LegoPathBoundary **)((int)&ADJ(pLVar1->o_LegoPathActor)->m_boundary + (int)this) ==
        pSVar5->m_edgeRef->m_b) && (pSVar5->m_lower <= local_4)) &&
     (local_4 <= pSVar5->m_upper)) {
        this->m_userState = pSVar5->m_userState;
  }
```
and the actual code looks like
```C++
if (m_boundary == current->m_edgeRef->m_b && current->m_lower <= skeletonCurAnimPhase &&
	skeletonCurAnimPhase <= current->m_upper) {
	m_userState = current->m_userState;
}
```
In particular, the reference to `m_boundary` is recognised correctly, which is part of the virtual superclass `LegoPathActor`.

Another good example is `0x10012ff0`.

### This adjust: Good example

See `0x1005d310`. Before the change we find
```C++
*(int **)&this[-1].field_0x188 = piVar2;                    
...        
*(float *)&this[-1].field_0x190 = (float)dVar5;
...                            
*(float *)&this[-1].field_0x1a8 = (float)dVar5;
```

After the change, the writes to `this` look more reasonable and reference fields that actually exist:
```C++
ADJ(this)->m_unk0x14 = (float)dVar5;
...                            
ADJ(this)->m_unk0x18 = (float)dVar5;
...                            
ADJ(this)->m_unk0x30 = (uint)(float)dVar5;
```

### Virtual inheritance: Struggles

See `0x10012ea0`. The decompile does not understand that it should offset by `this` and instead goes for arithmetic on `&this->m_unk0x5c`:
```C++
    (**(code **)(*(int *)((int)(&this->m_unk0x5c + 1) +
                         (int)(&ADJ(((this->base).vbase_offset)->o_LegoPathActor).base + -1) + 8) +
                0x30))(0);
```
Curiously, changing the offset of `o_LegoPathActor` (or, more correctly, of its type `LegoPathActorPtrOffset` to the incorrect value of `-3` instead of `-4`), the decompile improves:
```C++
    (**(code **)(*(int *)((undefined *)
                          ((int)&ADJ(((this->base).vbase_offset)->o_LegoPathActor).base.base.base.
                                 base.vftable + 1) + (int)this) + 0x30))(0);
```
Now it is quite evident that this is a call to the vtable, entry `0x30`, of the virtual superclass `LegoPathActor`.

The offset can be done as follows: Right-click on the data type `LegoPathActorPtrOffset` and click _Settings_:

grafik

Then change the _Component Offset_: 

grafik

There are also a number of other instances where this does not work as desired. Playing around with the offsets can sometimes improve the decompilation, but there is no way around reading the assembly in several other cases.

### This adjust - Struggles

Overall, the _This adjust_ handling works quite well. There is a minor issue when it calls a function which it _itself_ adjusted, e.g. at `0x10012d90`:
```C++
long __thiscall LegoRaceCar::Notify(LegoRaceCarPtrOffset0x84 this,MxParam *p_param) {
...
lVar1 = LegoRaceMap::Notify((LegoRaceMapPtrOffset0x38)&ADJ(this)->m_unk0x58,(int)p_param);
```

The field `m_unk0x58` is wrong because the offset of the other function is not accounted for. If the original offset of `0x84` is reduced to `0x84 - 0x38 = 0x4c`, the decompile is correct:
```C++
  lVar1 = LegoRaceMap::Notify((LegoRaceMapPtrOffset0x38)&ADJ(this)->base_LegoRaceMap_vbase_slim,
                              (int)p_param);
```
However, that breaks the decompilation in other functions, so one should not commit this change. Alternatively, one can declare a second typedef with a different offset for this function by hand. This will not be overwritten by the import script.

foxtacles:
Looks great, thanks for the in-depth explanation!

PR #1072, opened 2024-08-01
Implement/match `LegoRaceActor`

jonschz:
These functions match up to compiler entropy / minor non-functional diffs.

Note that there are some striking similarities between several functions. I am not exactly sure what is going on here - BETA10 does not suggest that these are inlined versions of the same function (although you can never be sure). Furthermore, we can never be sure what degree of code duplication the original code had. I'm curious to hear if you have more insight.

### DIFF
--- a/LEGO1/lego/legoomni/src/race/legoraceactor.cpp
+++ b/LEGO1/lego/legoomni/src/race/legoraceactor.cpp
@@ -47,9 +47,7 @@ MxS32 LegoRaceActor::VTable0x68(Vector3& p_v1, Vector3& p_v2, Vector3& p_v3)
 // FUNCTION: BETA10 0x100c9c93
 MxU32 LegoRaceActor::VTable0x90(float p_float, Matrix4& p_transform)
 {
-	// TODO: Note the similarity to LegoExtraActor::VTable0x90.
-	// Code duplication or is there something deeper going on?
-
+	// Note: Code duplication with LegoExtraActor::VTable0x90
 	switch (m_state) {
 	case 0:
 	case 1:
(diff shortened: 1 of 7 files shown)


LEGO1/lego/legoomni/src/race/legoraceactor.cpp
```diff
@@ -1,31 +1,121 @@
...
 
-// STUB: LEGO1 0x100147f0
-MxU32 LegoRaceActor::VTable0x90(float, Matrix4&)
+// FUNCTION: LEGO1 0x100147f0
+// FUNCTION: BETA10 0x100c9c93
+MxU32 LegoRaceActor::VTable0x90(float p_float, Matrix4& p_transform)
 {
-	// TODO
+	// TODO: Note the similarity to LegoExtraActor::VTable0x90.
+	// Code duplication or is there something deeper going on?
```
jonschz, line 51: We should (ideally) understand this part and simplify if possible. If we think it is an actual code duplication, we should adapt the comment.

LEGO1/lego/legoomni/src/race/legoracers.cpp
```diff
@@ -406,6 +406,8 @@ void LegoRaceCar::VTable0x70(float p_float)
 MxResult LegoRaceCar::VTable0x94(LegoPathActor* p_actor, MxBool p_bool)
 {
 	if (!p_actor->GetUserNavFlag()) {
+		// TODO: Note how this code is almost, but not quite identical to LegoRaceActor::VTable0x94.
+		// Maybe there is some code reuse here? The use of different globals is notable.
```
jonschz, line 410: Same here - the code is almost identical, but it references different global variables.

LEGO1/lego/legoomni/src/race/legoraceactor.cpp
foxtacles: Code duplication is pretty common across the code base and I'm fairly sure in these cases too, they just copy-pasted code and then applied minor adjustments. So I think the comments aren't necessary

PR #541, opened 2024-02-08
Updating `README` to have `CONFIG.EXE` progress bar

danielhejduk:
If you have some problems with the PR, request changes.

PR #1077, opened 2024-08-04
Update README.md

995qa:
Added CONFIG.EXE to the readme. somewhat minor change but could help in matching CONFIG.EXE's instructions.


foxtacles:
Thank you for the PR. This has been proposed a few times already but given the way it is currently presented, we've decided it's not a good idea to add CONFIG:

https://github.com/isledecomp/isle/pull/541
https://github.com/isledecomp/isle/pull/540

PR #1079, opened 2024-08-07
Implement `LegoCarRaceActor::VTable0x70` and `VTable0x1c`

jonschz:
`VTable0x70` matches to 100 %.

I am somewhat stuck on `VTable0x1c` - the logic seems to be correct, but the stack layout does not quite match and I have run out of ideas; some variables are off by 4 bytes. Suggestions are greatly appreciated.

### DIFF
--- a/LEGO1/lego/legoomni/src/race/legoracespecial.cpp
+++ b/LEGO1/lego/legoomni/src/race/legoracespecial.cpp
@@ -256,24 +256,23 @@ MxResult LegoCarRaceActor::VTable0x9c()
 		m_destEdge->FUN_1002ddc0(*m_boundary, local_50);
 		m_destEdge->FUN_1002ddc0(*m_boundary, local_20);
 
-
 		local_3c.EqualsCross(&local_50, m_boundary->GetUnknown0x14());
 		local_78.EqualsCross(m_boundary->GetUnknown0x14(), &local_20);
 
 		local_3c.Unitize();
 		local_78.Unitize();
 
-
 		((Vector3*) &local_3c)->Mul(5.0f);
 		((Vector3*) &local_78)->Mul(5.0f);
 
+		MxResult res = VTable0x80(Vector3(m_roi->GetWorldPosition()), local_3c, local_64, local_78);
 
-		MxResult res = LegoPathActor::VTable0x80(Vector3(m_roi->GetWorldPosition()), local_3c, local_64, local_78);
-		// BETA10 only?
+#ifndef NDEBUG // BETA10 only
 		if (res) {
 			assert(0);
 			return -1;
 		}
+#endif
 
 		// Proceed here
 	}
--- a/LEGO1/lego/legoomni/src/race/legoracespecial.cpp
+++ b/LEGO1/lego/legoomni/src/race/legoracespecial.cpp
@@ -243,29 +243,29 @@ MxResult LegoCarRaceActor::VTable0x9c()
 		Vector3* v2 = m_destEdge->CCWVertex(*m_boundary);
 		assert(v1 && v2);
 
-		Mx3DPointFloat local_64;
-		local_64[0] = (*v1)[0] + ((*v2)[0] - (*v1)[0]) * m_unk0xe4;
-		local_64[1] = (*v1)[1] + ((*v2)[1] - (*v1)[1]) * m_unk0xe4;
-		local_64[2] = (*v1)[2] + ((*v2)[2] - (*v1)[2]) * m_unk0xe4;
+		Mx3DPointFloat point1;
+		point1[0] = (*v1)[0] + ((*v2)[0] - (*v1)[0]) * m_unk0xe4;
+		point1[1] = (*v1)[1] + ((*v2)[1] - (*v1)[1]) * m_unk0xe4;
+		point1[2] = (*v1)[2] + ((*v2)[2] - (*v1)[2]) * m_unk0xe4;
 
-		Mx3DPointFloat local_50;
-		Mx3DPointFloat local_20;
-		Mx3DPointFloat local_3c;
-		Mx3DPointFloat local_78;
+		Mx3DPointFloat point2;
+		Mx3DPointFloat point3;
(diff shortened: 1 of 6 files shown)


LEGO1/lego/legoomni/src/race/legoracespecial.cpp
```diff
@@ -200,16 +208,75 @@ void LegoCarRaceActor::SwitchBoundary(LegoPathBoundary*& p_boundary, LegoUnknown
...
+		SwitchBoundary(m_boundary, m_destEdge, m_unk0xe4);
+		assert(m_boundary && m_destEdge);
+
+		// variable names verified by BETA10
+		Vector3* v1 = m_destEdge->CWVertex(*m_boundary);
+		Vector3* v2 = m_destEdge->CCWVertex(*m_boundary);
+		assert(v1 && v2);
+
+		Mx3DPointFloat point1;
+		point1[0] = (*v1)[0] + ((*v2)[0] - (*v1)[0]) * m_unk0xe4;
```
foxtacles, line 247: This should be a call to the `LERP3` operation/macro from `vec.h`

LEGO1/lego/legoomni/src/race/legoracespecial.cpp
```diff
@@ -200,16 +208,75 @@ void LegoCarRaceActor::SwitchBoundary(LegoPathBoundary*& p_boundary, LegoUnknown
...
-// STUB: LEGO1 0x10080be0
+// FUNCTION: LEGO1 0x10080be0
+// FUNCTION: BETA10 0x100cdc54
 MxResult LegoCarRaceActor::VTable0x9c()
 {
-	// TODO
+	MxS32 iVar4 = VTable0x1c(m_boundary, m_destEdge);
+
+	if (iVar4) {
+		SwitchBoundary(m_boundary, m_destEdge, m_unk0xe4);
```
foxtacles, line 238: Before the call to `SwitchBoundary`, a copy of `m_boundary` is taken:

`LegoPathBoundary* b = m_boundary;`

LEGO1/lego/legoomni/src/race/legoracespecial.cpp
```diff
@@ -200,16 +208,75 @@ void LegoCarRaceActor::SwitchBoundary(LegoPathBoundary*& p_boundary, LegoUnknown
...
+		point1[0] = (*v1)[0] + ((*v2)[0] - (*v1)[0]) * m_unk0xe4;
+		point1[1] = (*v1)[1] + ((*v2)[1] - (*v1)[1]) * m_unk0xe4;
+		point1[2] = (*v1)[2] + ((*v2)[2] - (*v1)[2]) * m_unk0xe4;
+
+		Mx3DPointFloat point2;
+		Mx3DPointFloat point3;
+		Mx3DPointFloat point4;
+		Mx3DPointFloat point5;
+
+		m_destEdge->FUN_1002ddc0(*m_boundary, point2);
```
foxtacles, line 256: This call is incorrect - it uses the copies of `m_boundary` and `m_destEdge` taken earlier:

`d->FUN_1002ddc0(*b, point2);`

The function will then match ~95%.

LEGO1/lego/legoomni/src/race/legoracespecial.cpp
```diff
@@ -200,16 +208,75 @@ void LegoCarRaceActor::SwitchBoundary(LegoPathBoundary*& p_boundary, LegoUnknown
...
+	}
 }
 
-// STUB: LEGO1 0x10080be0
+// FUNCTION: LEGO1 0x10080be0
+// FUNCTION: BETA10 0x100cdc54
 MxResult LegoCarRaceActor::VTable0x9c()
 {
-	// TODO
+	MxS32 iVar4 = VTable0x1c(m_boundary, m_destEdge);
```
foxtacles, line 235: Before the call to `VTable0x1c`, a copy of `m_destEdge` is taken:

`LegoUnknown100db7f4* d = m_destEdge;`

LEGO1/lego/legoomni/src/race/legoracespecial.cpp
```diff
@@ -200,16 +208,75 @@ void LegoCarRaceActor::SwitchBoundary(LegoPathBoundary*& p_boundary, LegoUnknown
...
+
+	if (iVar4) {
```
foxtacles, line 237: `iVar4` can go away, just place the call directly into the if-clause

LEGO1/lego/legoomni/src/race/legoracespecial.cpp
```diff
@@ -200,16 +208,75 @@ void LegoCarRaceActor::SwitchBoundary(LegoPathBoundary*& p_boundary, LegoUnknown
...
+		point4.EqualsCross(&point2, m_boundary->GetUnknown0x14());
+		point5.EqualsCross(m_boundary->GetUnknown0x14(), &point3);
+
+		point4.Unitize();
+		point5.Unitize();
+
+		((Vector3*) &point4)->Mul(5.0f);
+		((Vector3*) &point5)->Mul(5.0f);
+
+		MxResult res = VTable0x80(Vector3(m_roi->GetWorldPosition()), point4, point1, point5);
```
foxtacles, line 268: You can remove the explicit construction of `Vector3` - it happens implicitly

LEGO1/lego/legoomni/src/race/legoracespecial.cpp
jonschz: Thanks for working this out! I didn't think that one of the function calls would use the variables while the other one would not.

LEGO1/lego/legoomni/src/race/legoracespecial.cpp
```diff
@@ -7,15 +7,24 @@
 #include "misc.h"
 #include "mxmisc.h"
 #include "mxvariabletable.h"
+#include "vec.h"
```
foxtacles, line 10: should be <vec.h>

PR #1087, opened 2024-08-17
Implement/match LegoRaceMap

jonschz:
All the new functions match to 100 %. I have left a few comments with questions regarding the best practices here. Looking forward to your review!

### DIFF
--- a/LEGO1/lego/legoomni/include/legovideomanager.h
+++ b/LEGO1/lego/legoomni/include/legovideomanager.h
@@ -34,9 +34,12 @@ public:
 	MxResult Tickle() override;                                                                       // vtable+0x08
 	void Destroy() override;                                                                          // vtable+0x18
 	MxResult Create(MxVideoParam& p_videoParam, MxU32 p_frequencyMS, MxBool p_createThread) override; // vtable+0x2c
-	MxResult RealizePalette(MxPalette*) override;                                                     // vtable+0x30
-	void UpdateView(MxU32 p_x, MxU32 p_y, MxU32 p_width, MxU32 p_height) override;                    // vtable+0x34
-	virtual MxPresenter* GetPresenterAt(MxS32 p_x, MxS32 p_y);                                        // vtable+0x38
+
+	MxPresenter* GetPresenterByActionObjectName(char* p_char);
+
+	MxResult RealizePalette(MxPalette*) override;                                  // vtable+0x30
+	void UpdateView(MxU32 p_x, MxU32 p_y, MxU32 p_width, MxU32 p_height) override; // vtable+0x34
+	virtual MxPresenter* GetPresenterAt(MxS32 p_x, MxS32 p_y);                     // vtable+0x38
 
 	// FUNCTION: LEGO1 0x1007ab10
 	virtual LegoPhonemeList* GetPhonemeList() { return m_phonemeRefList; } // vtable+0x3c
--- a/LEGO1/lego/legoomni/src/video/legovideomanager.cpp
+++ b/LEGO1/lego/legoomni/src/video/legovideomanager.cpp
@@ -409,6 +409,28 @@ MxPresenter* LegoVideoManager::GetPresenterAt(MxS32 p_x, MxS32 p_y)
 	return NULL;
 }
 
+// FUNCTION: LEGO1 0x1007c180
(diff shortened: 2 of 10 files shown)


LEGO1/lego/legoomni/include/legovideomanager.h
```diff
@@ -34,9 +34,12 @@ class LegoVideoManager : public MxVideoManager {
 	MxResult Tickle() override;                                                                       // vtable+0x08
 	void Destroy() override;                                                                          // vtable+0x18
 	MxResult Create(MxVideoParam& p_videoParam, MxU32 p_frequencyMS, MxBool p_createThread) override; // vtable+0x2c
-	MxResult RealizePalette(MxPalette*) override;                                                     // vtable+0x30
-	void UpdateView(MxU32 p_x, MxU32 p_y, MxU32 p_width, MxU32 p_height) override;                    // vtable+0x34
-	virtual MxPresenter* GetPresenterAt(MxS32 p_x, MxS32 p_y);                                        // vtable+0x38
+
+	MxPresenter* GetPresenterByActionObjectName(char* p_char);
```
jonschz, line 38: Do you prefer to keep the virtual functions together, or do you try to match the order in the binaries? I am open for suggestions for putting this decleration somewhere else.

foxtacles: We generally keep the virtual functions together in one block. The order of the functions in the binaries doesn't need to be reflected in the header declarations

LEGO1/omni/include/mxpresenterlist.h
```diff
@@ -98,4 +104,13 @@ class MxPresenterListCursor : public MxPtrListCursor<MxPresenter> {
 // TEMPLATE: LEGO1 0x100225e0
 // MxList<MxPresenter *>::DeleteEntry
 
+// TEMPLATE: BETA10 0x1007d1d0
```
jonschz, line 107: I added some template annotations for BETA10. Is this desired and in the expected format? These are inlined in LEGO1.

LEGO1/lego/legoomni/src/video/legovideomanager.cpp
```diff
@@ -409,6 +409,28 @@ MxPresenter* LegoVideoManager::GetPresenterAt(MxS32 p_x, MxS32 p_y)
...
 }
 
+// FUNCTION: LEGO1 0x1007c180
+// FUNCTION: BETA10 0x100d6df4
+MxPresenter* LegoVideoManager::GetPresenterByActionObjectName(char *p_actionObjectName) {
+
+	MxPresenterListCursor cursor(m_presenters);
+	MxPresenter* presenter;
+
+	while (true) {
```
foxtacles, line 419: should be `TRUE`

LEGO1/lego/legoomni/src/video/legovideomanager.cpp
```diff
@@ -409,6 +409,28 @@ MxPresenter* LegoVideoManager::GetPresenterAt(MxS32 p_x, MxS32 p_y)
 	return NULL;
 }
 
+// FUNCTION: LEGO1 0x1007c180
+// FUNCTION: BETA10 0x100d6df4
+MxPresenter* LegoVideoManager::GetPresenterByActionObjectName(char *p_actionObjectName) {
```
foxtacles, line 414: This should probably be a const char*

LEGO1/omni/include/mxpresenterlist.h
```diff
@@ -23,18 +23,24 @@ class MxPresenterList : public MxPtrList<MxPresenter> {
 	// MxPresenterList::`scalar deleting destructor'
 };
 
+DECOMP_SIZE_ASSERT(MxPresenterList, 0x18)
```
foxtacles, line 26: size asserts should never be in a header - add it to any one of the CUs that uses a list

LEGO1/omni/include/mxpresenterlist.h
foxtacles: yes looks good 👍

LEGO1/omni/include/mxpresenterlist.h
jonschz: my bad, these were already in a cpp file. I have removed the redundant new ones.

PR #1088, opened 2024-08-18
Implement `RaceSkel`, add BETA10 annotations

jonschz:
Finding the correct vtables for `RaceSkel` revealed two missing virtual functions, which I also added. All the new functions (including synthetic ones) match to 100 %.

I also added a bunch of BETA10 annotations that I found along the way. Note that I only search for assertions and other discrepancies for some, but not all of the functions I annotated.

Looking forward to your review!

### DIFF
--- a/LEGO1/lego/legoomni/include/legoanimactor.h
+++ b/LEGO1/lego/legoomni/include/legoanimactor.h
@@ -28,13 +28,17 @@ struct LegoAnimActorStruct {
 
 // VTABLE: LEGO1 0x100d5440 LegoPathActor
 // VTABLE: LEGO1 0x100d5510 LegoAnimActor
+// VTABLE: BETA10 0x101b81d8 LegoPathActor
+// VTABLE: BETA10 0x101b82c8 LegoAnimActor
 // SIZE 0x174
 class LegoAnimActor : public virtual LegoPathActor {
 public:
+	// FUNCTION: BETA10 0x1000f6c0
 	LegoAnimActor() { m_curAnim = -1; }
 	~LegoAnimActor() override;
 
 	// FUNCTION: LEGO1 0x1000fba0
+	// FUNCTION: BETA10 0x10012400
 	const char* ClassName() const override // vtable+0x0c
 	{
 		// STRING: LEGO1 0x100f057c
@@ -42,6 +46,7 @@ public:
 	}
 
 	// FUNCTION: LEGO1 0x1000fbc0
+	// FUNCTION: BETA10 0x10012440
 	MxBool IsA(const char* p_name) const override // vtable+0x10
 	{
 		return !strcmp(p_name, LegoAnimActor::ClassName()) || LegoPathActor::IsA(p_name);
--- a/LEGO1/lego/legoomni/include/legoanimactor.h
+++ b/LEGO1/lego/legoomni/include/legoanimactor.h
@@ -35,6 +35,7 @@ class LegoAnimActor : public virtual LegoPathActor {
 public:
 	// FUNCTION: BETA10 0x1000f6c0
 	LegoAnimActor() { m_curAnim = -1; }
+
 	~LegoAnimActor() override;
 
 	// FUNCTION: LEGO1 0x1000fba0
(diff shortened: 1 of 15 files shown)


LEGO1/lego/legoomni/src/race/raceskel.cpp
```diff
@@ -1,13 +1,24 @@
...
 RaceSkel::RaceSkel()
 {
-	// TODO
+	m_animPosition = 0.0f;
+}
+
+// FUNCTION: LEGO1 0x10071ad0
+RaceSkel::~RaceSkel()
+{
+	// Removing this empty destructor leads to a mismatch in RaceSkel::`scalar deleting destructor'
```
jonschz, line 21: I am a little unsure about this part, but I couldn't get anything close to a match with a purely synthetic destructor. Declaring an empty destructor in `raceskel.h` was affected by the same problem.

foxtacles:
Currently on the road but will check tomorrow!

LEGO1/lego/legoomni/include/legoanimactor.h
```diff
@@ -28,20 +28,25 @@ struct LegoAnimActorStruct {
 
 // VTABLE: LEGO1 0x100d5440 LegoPathActor
 // VTABLE: LEGO1 0x100d5510 LegoAnimActor
+// VTABLE: BETA10 0x101b81d8 LegoPathActor
+// VTABLE: BETA10 0x101b82c8 LegoAnimActor
 // SIZE 0x174
 class LegoAnimActor : public virtual LegoPathActor {
 public:
+	// FUNCTION: BETA10 0x1000f6c0
 	LegoAnimActor() { m_curAnim = -1; }
```
foxtacles, line 37: Blank line below the ctor missing

LEGO1/lego/legoomni/include/raceskel.h
```diff
@@ -3,19 +3,30 @@
...
+// VTABLE: LEGO1 0x100d94c8 LegoAnimActor
+// VTABLE: BETA10 0x101bf9d0 LegoPathActor
+// VTABLE: BETA10 0x101bfac0 LegoAnimActor
 // SIZE 0x178
 class RaceSkel : public LegoAnimActor {
 public:
 	RaceSkel();
+	~RaceSkel() override;
+
+	void ParseAction(char* p_extra) override;
```
foxtacles, line 16: Virtual functions (overrides or not) should have the // vtable+0x annotation

LEGO1/lego/legoomni/include/raceskel.h
```diff
@@ -3,19 +3,30 @@
...
+
+	MxResult FUN_1001c360(float p_und, Matrix4& p_transform) override;
```
foxtacles, line 18: Virtual functions (overrides or not) should have the // vtable+0x annotation

LEGO1/lego/legoomni/src/race/raceskel.cpp
foxtacles: Yes this is correct. You can remove the comment. Sometimes they just defined an empty explicit constructor in this way

foxtacles (approved):

LEGO1/lego/legoomni/include/raceskel.h
jonschz: Do we have a rule regarding multiple and virtual inheritance? It looks like the existing code is not entirely consistent about that either. For example, `LegoAnimActor` has no vtable annotations on the virtual functions it introduces. Another issue: `RaceSkel::FUN_1001c360` would have a `// vtable+0x04` but still come below `ParseAction` because it is in a different vtable.

I don't really mind how we want to do it, but the current pattern and rules aren't entirely clear to me.

foxtacles: Ah yes, I forgot about the multiple inheritance. You are right that it's not consistent. We don't have any established rules for theses cases yet - considering this is limited to a few classes we may not need to have any. Let's just leave it as-is for now

PR #1089, opened 2024-08-21
Various fixes in constructors

jonschz:
I looked for various constructors not matching perfectly and found a couple of issues I fixed here.

I will probably continue to work on `LegoCarBuild`, where I have some progress that is not yet ready to be merged.

### DIFF
--- a/LEGO1/lego/legoomni/include/legocarbuild.h
+++ b/LEGO1/lego/legoomni/include/legocarbuild.h
@@ -61,6 +61,9 @@ public:
 	LegoCarBuild();
 	~LegoCarBuild() override;
 
+	MxLong Notify(MxParam& p_param) override; // vtable+0x04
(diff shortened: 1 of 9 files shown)


LEGO1/lego/legoomni/include/legocarbuild.h
```diff
@@ -77,11 +74,19 @@ class LegoCarBuild : public LegoWorld {
 		return !strcmp(p_name, LegoCarBuild::ClassName()) || LegoWorld::IsA(p_name);
 	}
 
+	MxLong Notify(MxParam& p_param) override;         // vtable+0x04
```
foxtacles, line 77: These should remain where they were before to maintain the ascending order of virtual function offsets

jonschz: oh, you're right, I missed that the other two are virtual functions as well. Will change that back.

PR #1091, opened 2024-08-24
BETA10: reccomp support and Ghidra imports

jonschz:
Changes:
- A small issue was fixed that caused `reccmp` to crash when comparing a debug build to BETA10.
- A debug build can now be imported into Ghidra using the import script. I will upload a version to the shared Ghidra repository after the review.
- The import script received some improvements regarding thunks, which also has a positive effect on the LEGO1 repository.
- I fixed a few obviously wrong BETA10 annotations (probably made by me) that I found after running `reccmp`.

This is clearly not perfect since the codebase differs between BETA10 and LEGO1, but it has still shown to be rather helpful. In the future we could add some dedicated annotations / blacklists in case we want to maintain some BETA10/LEGO1 differences in the BETA10 Ghidra repository by hand, that shouldn't be overwritten by the import script. Vtable mismatches are also a slight issue right now (they can be shorter in BETA10).

Overall, the BETA10 got a match of 85 %, which I found surprisingly high. Matching BETA10 might also help getting a few difficult functions up to 100 % in both BETA10 and LEGO1.

If you want to use the import script, create a debug build in `./build_debug`. I added a note on creating a debug build in the tools README in my last PR.

Looking forward to your thoughts!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,11 +12,16 @@ ENV/
 VENV/
 env.bak/
 venv.bak/
-ISLE.EXE
-LEGO1.DLL
 /build/
+/build_debug/
+/legobin/
 *.swp
 LEGO1PROGRESS.*
 ISLEPROGRESS.*
 *.pyc
 tools/ghidra_scripts/import.log
+
+# By convention we put the retail binaries into ./legobin.
+# These entries are kept for now since that convention has not always been around.
+ISLE.EXE
+LEGO1.DLL
--- a/LEGO1/lego/legoomni/src/race/legoracespecial.cpp
+++ b/LEGO1/lego/legoomni/src/race/legoracespecial.cpp
@@ -209,7 +209,7 @@ void LegoCarRaceActor::SwitchBoundary(LegoPathBoundary*& p_boundary, LegoUnknown
 }
 
 // FUNCTION: LEGO1 0x10080b70
-// FUNCTION: BETA10 0x1000366b
+// FUNCTION: BETA10 0x100cdbae
 void LegoCarRaceActor::VTable0x70(float p_float)
 {
 	// m_unk0x0c is not an MxBool, there are places where it is set to 2 or higher
--- a/LEGO1/realtime/vector.h
+++ b/LEGO1/realtime/vector.h
@@ -89,7 +89,7 @@ public:
 	virtual float Dot(float* p_a, float* p_b) const { return DotImpl(p_a, p_b); } // vtable+0x3c
 
 	// FUNCTION: LEGO1 0x100020f0
-	// FUNCTION: BETA10 0x100028f6
+	// FUNCTION: BETA10 0x100108c0
 	virtual float Dot(Vector2* p_a, Vector2* p_b) const { return DotImpl(p_a->m_data, p_b->m_data); } // vtable+0x38
 
 	// FUNCTION: LEGO1 0x10002110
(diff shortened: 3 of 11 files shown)


foxtacles:
Looks good to me! @disinvite could you check it as well?

disinvite:
The code looks good. Do we need `_match_thunk_in_vtable`? I'm not sure the warning messages tell us anything that the main vtable diff doesn't. `_match_thunks` already covers all the debug thunks for already-matched functions, so I don't think anything new will go into the database here.

For example, we have these messages:
```
[WARNING] Thunk jump target mismatch at (100035d5, 100033dc)
[WARNING] Thunk jump target mismatch at (1000484f, 100019b5)
```

They show a problem with the vtable for LegoExtraActor. The address in the annotation is `0x101bc2b8`, but it should be `0x101bc2a0`. However, the diff already looks like this:

```diff
-0x100049f8 from orig not annotated.
-(0x100035d5 / 0x100039d6) : Thunk of 'LegoEntity::Notify'
-(0x1000484f / 0x10004b47) : Thunk of 'MxCore::Tickle'
-0x1000188e from orig not annotated.
+(no orig / 0x1000245f) : Thunk of 'LegoAnimActor::FUN_1001c1f0'
+(0x1000119f / 0x100033dc) : Thunk of 'LegoAnimActor::FUN_1001c360'
+(no orig / 0x100019b5) : Thunk of 'LegoAnimActor::FUN_1001c450'
+(no orig / 0x10002356) : Thunk of 'LegoAnimActor::ClearMaps'
```

I think the alert about vtable size mismatch is more helpful, but it won't be something we can fix if the beta was different.

We can supplement the bounds checking on orig vtable size by looking at the next orig address in the database. (`_compare_function` also does this.) This should cover some of the cases where the orig vtable is followed by another.

For example:

```python
next_orig = self._db.get_next_orig_addr(match.orig_addr)
if next_orig - match.orig_addr < vtable_size:
    logger.warning("Orig vtable is larger than recomp vtable for %s", match.name)
```

However, adding this doesn't currently have any effect on the beta.

jonschz:
Thanks for the review!

> Do we need `_match_thunk_in_vtable`?

Ouch. Yeah, I investigated for a bit, and I think you are right. I think in hindsight I misdiagnosed why I didn't see the thunks, and when I fixed that, my new code was already in place. I'll remove it since it's redundant.

> I think the alert about vtable size mismatch is more helpful, but it won't be something we can fix if the beta was different.

I agree. I had to implement the check anyway since an exception would be thrown otherwise, and adding in the warning was a no-brainer.

> However, adding this doesn't currently have any effect on the beta.

Where exactly did you put it? Even if it doesn't have any effect at the moment, it won't hurt either. The vtables in BETA10 have a lot of gaps in between them, maybe there's a minimum separation between them. But who knows.

disinvite:
IIRC I added it near these lines:

```python
vtable_size = 4 * (match.size // 4)
orig_table = self.orig_bin.read(match.orig_addr, vtable_size)
recomp_table = self.recomp_bin.read(match.recomp_addr, vtable_size)
```

As for it not having any effect; just wanted to set expectations 😉 But yes, it would still be good to have. We might find a case where the retail has more virtual methods than the beta.

jonschz:
I have already seen a number of such cases. So far, in each such case the BETA10 had a gap of zeros to the subsequent vtable, so it was handled by the check for a zero value.

PR #1101, opened 2024-09-08
Fix spurious reccmp warnings

jonschz:
This small change fixes a lot of spurious errors logged e.g. in `reccmp`, particularly when working with `BETA10`. I first considered not adding the `LPROC32` functions to the list of known symbols, but they do not appear to cause any harm and might come in handy at some point.


foxtacles:
@disinvite looks good?

disinvite:
Yes

PR #1102, opened 2024-09-08
Improve `UnknownMx4DPointFloat::FUN_100040a0`

jonschz:
This one was quite annoying and I didn't get it to match as well as I hoped, but I still think it is an improvement. I would greatly appreciate if someone else took a look as well.

I tried to match `FUN_100040a0` to BETA10 and found a number of corrections there, bringing it up to 91.4 %. LEGO1 might be down to compiler entropy, but I am not sure.

One extremely annoying thing I noticed is that the BETA10 compilation depends on the _names of variables_, even if they are unused.

### DIFF
--- a/LEGO1/library_msvc.h
+++ b/LEGO1/library_msvc.h
@@ -669,4 +669,10 @@
 // LIBRARY: BETA10 0x100f9060
 // _assert
 
+// LIBRARY: BETA10 0x100fa2fa
+// _acos
+
+// LIBRARY: BETA10 0x100f9a50
+// _sin
+
 #endif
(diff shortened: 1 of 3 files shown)


disinvite:
> One extremely annoying thing I noticed is that the BETA10 compilation depends on the _names of variables_, even if they are unused.

Is that what determines the stack order?

jonschz:
I didn't figure out the entire pattern so I can't say anything for certain, but I can say for sure that the mere act of renaming a variable did at least *influence* the stack order (I tried a couple of names for the variables I ultimately named `a` and `b`). I guess there are other factors on the stack order as well (like the scope of the variable, when and where are they first used, ...).

foxtacles:
Looks good to me! I'll give it a look later this week but it's an improvement for sure

PR #1103, opened 2024-09-11
Progress on `LegoCarBuildAnimPresenter`

jonschz:


LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
```diff
@@ -1,23 +1,55 @@
 #include "legocarbuildpresenter.h"
 
+#include "legoentity.h"
+#include "mxautolock.h"
+
+DECOMP_SIZE_ASSERT(LegoCarBuildAnimPresenterUnknownListEntry, 0xc)
```
foxtacles, line 6: Should be `0x0c`

LEGO1/lego/legoomni/include/legocarbuildpresenter.h
```diff
@@ -4,7 +4,15 @@
 #include "anim/legoanim.h"
 #include "legoanimpresenter.h"
 
+// SIZE 0xc
```
foxtacles, line 7: Should be `0x0c`. Also maybe this struct should be defined within LegoCarBuildAnimPresenter itself?

foxtacles (approved):
Looks great!

PR #1104, opened 2024-09-14
More functions in `LegoCarBuild`

jonschz:
I find one diff a little confusing:
```diff
 -  __FPinit (0x1008b340) is 100.00% similar to the original
+  __FPinit (0x1008b340) is 0.00% similar to the original
```

Removing it from `library_msvc.h` does not appear to affect any other functions, but I'd appreciate some feedback on that before doing so.


LEGO1/lego/legoomni/src/build/legocarbuild.cpp
```diff
@@ -6,14 +6,17 @@
...
 #include "mxbackgroundaudiomanager.h"
 #include "mxcontrolpresenter.h"
 #include "mxmisc.h"
 #include "mxnotificationmanager.h"
 #include "mxsoundpresenter.h"
 #include "mxstillpresenter.h"
 #include "mxticklemanager.h"
+#include "mxtransitionmanager.h"
 #include "scripts.h"
+#include "vec.h"
```
foxtacles, line 19: should be <vec.h>

foxtacles:
> I find one diff a little confusing:
> 
> ```diff
>  -  __FPinit (0x1008b340) is 100.00% similar to the original
> +  __FPinit (0x1008b340) is 0.00% similar to the original
> ```
> 
> Removing it from `library_msvc.h` does not appear to affect any other functions, but I'd appreciate some feedback on that before doing so.

Where do you encounter this diff, in this PR?

jonschz:
> Where do you encounter this diff, in this PR?

Yes, on the first commit of this PR, but it seems to have disappeared with the second one. Might be a fluke, probably not worth debugging unless it comes back.

PR #1106, opened 2024-09-20
some educated guesses for unknown view manager names

crtdll:
haven't ran any formatting so that'll probably fail, if this is merged i'd appreciate if you could format it @foxtacles


foxtacles:
Thank you, looks good. Changed class members to snake case because this class uses a different naming convention than most of the rest of the code base (based on the alpha/beta names). These contributions help a lot to gain better understanding of what the code does.

PR #1108, opened 2024-10-04
Implement `LegoCarBuildAnimPresenter::FUN_10079920()`

jonschz:
Matches to 100 %.

Since the purpose of the code was sufficiently clear, I gave this one a speculative name and changed some other names as well. 

`g_rotationAngleStepYAxis` is not referenced anywhere else in the code base, so we will not get any additional information on it. It might be declared as `const MxFloat` in `legocarbuild.cpp`, but I guess we will only worry about that when all functions are implemented and we try to get the order of everything in the `.dll` file correct.

### DIFF
--- a/LEGO1/lego/sources/anim/legoanim.h
+++ b/LEGO1/lego/sources/anim/legoanim.h
@@ -24,6 +24,17 @@ public:
 	LegoU32 TestBit2() { return m_flags & c_bit2; }
 	LegoU32 TestBit3() { return m_flags & c_bit3; }
 
+	// FUNCTION: BETA10 0x100739a0
+	void FUN_100739a0(MxS32 p_param)
+	{
+		if (p_param) {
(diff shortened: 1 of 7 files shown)


LEGO1/lego/sources/anim/legoanim.h
```diff
@@ -24,6 +24,17 @@ class LegoAnimKey {
...
 	LegoU32 TestBit3() { return m_flags & c_bit3; }
 
+	// FUNCTION: BETA10 0x100739a0
+	void FUN_100739a0(MxS32 p_param)
+	{
+		if (p_param) {
+			m_flags |= 0x01;
+		}
+		else {
+			m_flags &= 0xfe;
```
foxtacles, line 34: I'd introduce an enum constant for this flag in the class (see other enums for examples). Clearing the flag is then done with `&= ~c_bit1`

PR #1109, opened 2024-10-05
Implement `LegoCarBuildAnimPresenter::StreamingTickle()`

jonschz:
`LegoCarBuildAnimPresenter::StreamingTickle()` matches up to the stack layout, which I couldn't get quite right.

I also added a number of `BETA10` annotations and found a bunch of stubs.

### DIFF
--- a/LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
+++ b/LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
@@ -125,7 +125,7 @@ void LegoCarBuildAnimPresenter::StreamingTickle()
 
 	Lego3DView* lego3dview = videoManager->Get3DManager()->GetLego3DView();
 	LegoROI* videoManagerROI = videoManager->GetViewROI();
-	LegoROI* local_60 = m_unk0x140->GetROI();
+	LegoROI* local60 = m_unk0x140->GetROI();
 	LegoROI* camera = NULL;
 	MxFloat fov;
 
@@ -135,7 +135,7 @@ void LegoCarBuildAnimPresenter::StreamingTickle()
 		LegoAnimNodeData* animNodeData = (LegoAnimNodeData*) GetTreeNode(m_anim->GetRoot(), i)->GetData();
 
 		if (strnicmp(animNodeData->GetName(), "CAM", strlen("CAM")) == 0) {
-			camera = local_60->FindChildROI(animNodeData->GetName(), local_60);
+			camera = local60->FindChildROI(animNodeData->GetName(), local60);
 			fov = atof(&animNodeData->GetName()[strlen(animNodeData->GetName()) - 2]);
 			break;
 		}
@@ -143,7 +143,7 @@ void LegoCarBuildAnimPresenter::StreamingTickle()
 
 	assert(camera); // verifies variable name 'camera'
 
-	LegoROI* targetROI = local_60->FindChildROI("TARGET", local_60);
+	LegoROI* targetROI = local60->FindChildROI("TARGET", local60);
 
 	Mx3DPointFloat dirVec;
 
@@ -164,7 +164,7 @@ void LegoCarBuildAnimPresenter::StreamingTickle()
 	lego3dview->Moved(*videoManagerROI);
 	videoManager->Get3DManager()->SetFrustrum(fov, 0.1, 250.0);
 
-	m_unk0xe0 = local_60->FindChildROI("VIEW", local_60)->GetLocal2World();
+	m_unk0xe0 = local60->FindChildROI("VIEW", local60)->GetLocal2World();
 
 	m_previousTickleStates |= 1 << m_currentTickleState;
 	m_currentTickleState = e_repeating;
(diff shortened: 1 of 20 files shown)


LEGO1/realtime/vector.h
```diff
@@ -163,13 +163,18 @@ class Vector2 {
...
 
-	// SYNTHETIC: BETA10 0x10013460
-	// Vector3::operator[]
+	// There is another candidate for `Vector2::operator[]` at BETA10 0x10010890, which is called from only three
+	// functions in BETA10:
+	// - `Matrix4::FromQuaternion()`
+	// - `Matrix4::ToQuaternion()`
+	// - `UnknownMx4DPointFloat::FUN_100040a0()`
+	// Maybe there is another subclass of `Vector4` involved that has the same VTABLE but a different `operator[]`.
+	// It is also interesting that `Matrix4::operator[]` is located right above at BETA10 0x10010860.
```
jonschz, line 172: This is a weird issue - not exactly sure what is going on here. Binary comparison does not help since these functions are all identical. It would not make sense to have `0x10010890` be a `Matrix4::operator[]` since the places that call it clearly assume a `Vector4`. Maybe `operator[]` is overwritten for `Vector4` by the same code?

LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
```diff
@@ -65,11 +71,103 @@ void LegoCarBuildAnimPresenter::ReadyTickle()
 	// TODO
 }
 
-// STUB: LEGO1 0x100789e0
-// STUB: BETA10 0x10070cdd
+// FUNCTION: LEGO1 0x100789e0
+// FUNCTION: BETA10 0x10070cdd
```
jonschz, line 75: The code is mostly, but not exactly identical on BETA10. The match on BETA10 can be improved by adding `int dummy1, dummy2;` somewhere and disabling lines 90-92 as well as line 169. The highest I managed to get to was around 70 %.

LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
```diff
@@ -47,8 +53,8 @@ LegoCarBuildAnimPresenter::~LegoCarBuildAnimPresenter()
 	m_unk0xc8.GetRoot()->SetNumChildren(0);
 	*m_unk0xc8.GetRoot()->GetChildren() = NULL;
 
-	if (m_unk0x14c) {
-		delete m_unk0x14c;
+	if (m_mainSourceId) {
+		delete m_mainSourceId;
```
foxtacles, line 57: should be `delete[]`

LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
```diff
@@ -65,11 +71,103 @@ void LegoCarBuildAnimPresenter::ReadyTickle()
...
+
+	Mx3DPointFloat dirVec;
+
+	Vector3 cameraPosition(camera->GetWorldPosition());
+	Vector3 upVec(camera->GetWorldUp());
+	Vector3 targetPosition(targetROI->GetWorldPosition());
+
+	MxMatrix localTransform;
+
+	dirVec[0] = targetPosition[0] - cameraPosition[0];
```
foxtacles, line 156: this is probably the `VMV3` macro in `vec.h`

LEGO1/realtime/vector.h
foxtacles: Yeah, the `operator[]` access functions on the vector/matrix classes are a big mystery. I've been struggling with getting them right for a long time. The issue is that there are a large amount of possible implementations, where inserting (or omitting) an `operator[]` in any of the classes in the vector/matrix hierarchies doesn't make any functional difference, but will affect binary generation across the board. It's very hard to pin down.

LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
jonschz: I thought that too and tried it, but it did decrease the match for LEGO1. I could still change if if desired.

LEGO1/realtime/vector.h
jonschz: Feel free to close this thread if you're happy with the comment as it is right now. I am faily confident that at least the two `Vector2::operator[]` BETA10 addresses are correct now (up to permutation between them).

LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
jonschz: Huh, I learned something today. It does not affect the binary code at all since `LegoChar` does not have a destructor, but it's definitely cleaner that way.

LEGO1/realtime/vector.h
jonschz: Addendum: I just experimented a bit more and found that
```C++
class Vector4 {
	// [...]

	float& operator[](int idx) { return m_data[idx]; }

	// FUNCTION: BETA10 0x10010890
	const float& operator[](int idx) const { return m_data[idx]; }
}
```
appears to only cause a lot of entropy changes in LEGO1 (would need some additional verification) and overall improves the situation on BETA10.

Maybe we'll add that in another PR?

PR #1111, opened 2024-10-11
Implement `LegoCarBuildAnimPresenter::FUN_10079160()` and others

jonschz:
`FUN_10079160()` matches up to stack order, the rest matches up to entropy or to 100 %.

I have worked a bit on automatic stack comparison in order to simplify debugging stack mismatches (or at least in order to tell whether the code is correct). I'll open a PR when it's somewhat usable.

### DIFF
--- a/LEGO1/lego/legoomni/include/legocarbuildpresenter.h
+++ b/LEGO1/lego/legoomni/include/legocarbuildpresenter.h
@@ -11,14 +11,6 @@ class LegoCarBuildAnimPresenter : public LegoAnimPresenter {
 public:
 	// SIZE 0x0c
 	struct UnknownListEntry {
-		// variable name verified by BETA10 0x10071b56
-		LegoChar* m_name; // 0x00
-
-		// variable name verified by BETA10 0x100719f0
-		LegoChar* m_wiredName; // 0x04
-
-		undefined2 m_unk0x08; // 0x08
-
 		// FUNCTION: LEGO1 0x100795c0
 		// FUNCTION: BETA10 0x10073850
 		UnknownListEntry()
@@ -27,6 +19,14 @@ public:
 			m_wiredName = NULL;
 			m_unk0x08 = 0;
 		}
+
+		// variable name verified by BETA10 0x10071b56
+		LegoChar* m_name; // 0x00
+
+		// variable name verified by BETA10 0x100719f0
+		LegoChar* m_wiredName; // 0x04
+
+		undefined2 m_unk0x08; // 0x08
 	};
 
 	LegoCarBuildAnimPresenter();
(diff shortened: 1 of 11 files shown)


LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
```diff
@@ -185,23 +230,125 @@ void LegoCarBuildAnimPresenter::EndAction()
...
-	// called from LegoCarBuildAnimPresenter::StreamingTickle()
-	// TODO
+	LegoTreeNode* root;
+	LegoAnimNodeData* data2;
+	MxS16 i;
+	MxS16 totalNodes = CountTotalTreeNodes(m_anim->GetRoot());
+	LegoChar* name;
+	LegoTreeNode* destNode;
+	LegoAnimNodeData* destData;
+	LegoTreeNode** children;
```
jonschz, line 268: I couldn't get this stack layout quite right. What I observed:
* The order of declaration does not appear to matter when the variables are not initialized
* The names of these variables **do** matter, at least in BETA10. However, the pattern is still not quite clear to me
* The place where the variable is declared does not appear to matter as long as it is on the top level (it does matter whether the variable is declared in a smaller scope)
* Tests with my preliminary tooling suggest that these variables map 1:1 to the variables in BETA10, suggesting that the logic is correct.

LEGO1/lego/legoomni/include/legocarbuildpresenter.h
```diff
@@ -11,9 +11,22 @@ class LegoCarBuildAnimPresenter : public LegoAnimPresenter {
 public:
 	// SIZE 0x0c
 	struct UnknownListEntry {
-		LegoChar* m_unk0x00;    // 0x00
-		LegoChar* m_unk0x04;    // 0x04
-		undefined m_unk0x08[4]; // 0x08
+		// variable name verified by BETA10 0x10071b56
+		LegoChar* m_name; // 0x00
```
foxtacles, line 15: Members should be declared at the bottom of the class

LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
```diff
@@ -218,6 +365,55 @@ void LegoCarBuildAnimPresenter::FUN_10079680(LegoChar* p_param)
...
+		}
+
+		for (MxS32 i = 0; i < p_treeNode->GetNumChildren(); i++) {
+			node = FindNodeByName(p_treeNode->GetChildren()[i], p_name);
+
+			if (node) {
+				return node;
+			}
+		}
+	}
```
foxtacles, line 413: Needs blank line before return

foxtacles (commented):
Looks great!

LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
jonschz: I was confused for a second because Github's preview is not helpful at all here, suggesting that a different `return` statement was meant. Fixed it

PR #1112, opened 2024-10-13
Add new script to compare the stack layout

jonschz:
I have written a script that compares all references to the stack between the original and the recompiled binary. This is especially useful for BETA10, where it can be hard to tell if the order of variables on the stack is wrong or if there is an actual logic error in the recomp (the wrong variable is read or written). The script may also help understand how `msvc` decides the stack order since the analysis is much faster now. The names of the variables are pulled from the debug info.

Looking forward to your feedback - especially whether the built-in explanations are clear and/or how they could be improved.

**Example outputs**
Here is the (slightly abbreviated) output for a modified version of `LegoCarBuildAnimPresenter::FUN_10079160()` that matches up to the order of stack variables (rename `root` to `aaaroot` as declared here):
https://github.com/isledecomp/isle/blob/974cd7ce7c727d21652c8383f0c58910fce87602/LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp#L261
grafik

Here is the abbreviated output for ther same function as it is on `master` right now, with a slight structural mismatch:
grafik
grafik
Note how `ebp - 0x18` is listed under `not seen` since its only appearance is in the structural mismatch. It could also be that it only appears in the middle of a perfectly matching block which - those are always omitted from the output (like in `reccmp`).

Here is the abbreviated output for the same function with an intentional logic error - `counter++` is replaced by `i++` in the following line:
https://github.com/isledecomp/isle/blob/974cd7ce7c727d21652c8383f0c58910fce87602/LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp#L326
grafik

**Limitations**
- This script works best when the structure of the two functions matches perfectly, and only the stack layout differs. Structural mismatches are skipped since it cannot be determined automatically which instructions correspond.
- It might make sense to parse large matching blocks as well - those are not returned by `Compare.compare_address()`. This will likely require changes around the following line, such that the `context_size` can be set very high or disabled entirely: https://github.com/isledecomp/isle/blob/974cd7ce7c727d21652c8383f0c58910fce87602/tools/isledecomp/isledecomp/compare/core.py#L717
- I suspect there will be some issues with release builds, where I have not yet looked extensively for interesting cases to analyse.
  - As far as I can tell, debug builds only ever reference variables relative to `ebp`, never from `esp`. This is different for release builds, which reference the stack both from `ebp` and from `esp`.
  - I assume that the references from `esp` do not work correctly under some circumstances. It will likely be necessary to keep track of the stack size (i.e. parse at least every `sub esp, ...`, `add esp, ...`, `push`, `call`). This is not implemented right now.
  - I am not sure how well this code handles reused stack variables in LEGO1. It is possible that the `S_BLOCK32` entries in the PDB need to be parsed and each block handled separately.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -175,6 +175,8 @@ The example usages below assume that the current working directory is this repos
     * Generate an HTML report: `py -m tools.reccmp.reccmp --html output.html legobin/LEGO1.DLL build/LEGO1.DLL build/LEGO1.PDB .`
     * Create a base file for diffs: `py -m tools.reccmp.reccmp --json base.json --silent legobin/LEGO1.DLL build/LEGO1.DLL build/LEGO1.PDB .`
     * Diff against a base file: `py -m tools.reccmp.reccmp --diff base.json legobin/LEGO1.DLL build/LEGO1.DLL build/LEGO1.PDB .`
+* [`stackcmp`](/tools/stackcmp): Compares the stack layout for a given function that almost matches.
+    * e.g. `py -m tools.stackcmp.stackcmp legobin/BETA10.DLL build_debug/LEGO1.DLL build_debug/LEGO1.pdb . 0x1007165d`
 * [`roadmap`](/tools/roadmap): Compares symbol locations in an original binary with the same symbol locations of a recompiled binary
 * [`verexp`](/tools/verexp): Verifies exports by comparing the exports of the original DLL and the recompiled DLL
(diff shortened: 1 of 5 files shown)


jonschz:
I have found another issue. Will comment when I have resolved it.

foxtacles (approved):
This looks extremely useful, thanks! Code and output looks good and understandable to me. @disinvite do you have any feedback?

disinvite:
This looks great!

PR #1114, opened 2024-10-20
Complete `LegoCarBuildAnimPresenter`

jonschz:
also contains some drive-by fixes for BETA10 annotations

### DIFF
--- a/LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
+++ b/LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
@@ -65,7 +65,7 @@ LegoCarBuildAnimPresenter::~LegoCarBuildAnimPresenter()
 }
 
 // FUNCTION: BETA10 0x100733d0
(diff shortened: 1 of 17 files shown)


DimaLeon2000:
Looks (sort of) good to me.

LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
```diff
@@ -62,10 +64,75 @@ LegoCarBuildAnimPresenter::~LegoCarBuildAnimPresenter()
...
+		if (m_unk0x13c & 1) {
+			bvar5 = 1;
+			m_unk0x13c = time + 400;
+		}
+		else {
+			m_unk0x13c = time + 200;
+		}
+
+		if (bvar5) {
+			m_unk0x13c &= ~1;
```
foxtacles, line 86: we should add an enum with a constant for this flag (if it doesn't exist yet)

PR #1115, opened 2024-10-23
replace semikolons with commas in readme

BJNFNE:


foxtacles:
I don't see anything wrong with the semicolons

PR #1116, opened 2024-10-24
Use reccmp as a python requirement

madebr:
You can install all tools, and ncc's requirements, by running:
```
pip install -r tools/requirements.txt
```


foxtacles:
> You can install all tools, and ncc's requirements, by running:
> 
> ```
> python -r tools/requirements.txt
> ```

You mean `pip install -r tools/requirements.txt`, right?

Documentation looks good to me - just set up everything locally as described and it worked out of the box. @disinvite @jonschz if you could check/test as well and confirm everything looks good to you this should be ready to merge.

jonschz:
Looks great overall! I ran into one minor issue which we may need to agree on:
* At the moment, all registered original binaries need to be present for the code to execute at all (e.g. I can't run `datacmp` on `LEGO1` if `CONFIG.EXE` isn't present). At a glance, this looks like something that can (and probably should) be changed. I don't necessarily consider that to be a blocker, but I'm happy to hear your opinions as well.
* Currently, BETA10 is not part of the (commited) `reccmp-project.yml`. If we add it there before the first point is addressed, all developers (and probably the CI) need `BETA10.dll` to be present so any reccmp scripts can be run. If we don't add it to the file, developers that want to use BETA10 for local development have to change `reccmp-project.yml` locally but not commit it.
* Ghidra imports do not work yet, but since we don't really depend on them and I intend to fix them soon anyway, I don't consider that to be a blocker.

jonschz:
Thanks for the quick fix @madebr ! Maybe we want to add the following to the project yaml?
```yaml
  BETA10:
    filename: BETA10.DLL
    source-root: LEGO1
    hash:
      sha256: d91435a40fa31f405fba33b03bd3bd40dcd4ca36ccf8ef6162c6c5ca0d7190e7
```

madebr:
> Thanks for the quick fix @madebr ! Maybe we want to add the following to the project yaml?

Done.
About the warning messages, this situation will improve once we dive into https://github.com/isledecomp/reccmp/issues/8

disinvite:
Everything that I've checked looks good. I noticed that the roadmap tool still expects the original four args instead of using the project file, but it's still perfectly usable. I added isledecomp/reccmp#10 for this.

The tools other than `reccmp-reccmp` have more logging messages than they did before, but this also isn't a critical problem. The main impact would be to the progress report diff, but this is run by `reccmp`, so no change. @jonschz already put in isledecomp/reccmp#5 so we can discuss over there.

Would it be worth adding a note to the main README.md to point at the readme under `/tools`?

jonschz:
> Would it be worth adding a note to the main README.md to point at the readme under /tools?

It is linked under `CONTRIBUTING.md`, which I feel is sufficient.

However, I noticed that the README under `/tools` still has a lot of duplication with the README in the reccmp repo (e.g. the explanation for the different annotation types). I think this can be cleaned up later, though.

PR #1117, opened 2024-10-25
Implement LegoCarBuild functions, refactor BackgroundAudioManager

jonschz:
* Implement LegoCarBuild::FUN_10024c20 and dependents
* Implement LegoCarBuild::FUN_100250e0 and dependents
* Type refactor in `MxBackgroundAudioManager`: `int` to existing `enum`

### DIFF
--- a/LEGO1/lego/legoomni/include/legocarbuild.h
+++ b/LEGO1/lego/legoomni/include/legocarbuild.h
@@ -16,6 +16,14 @@ class MxActionNotificationParam;
 // SIZE 0x50
 class LegoVehicleBuildState : public LegoState {
 public:
+	enum AnimationState {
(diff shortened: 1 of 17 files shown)


LEGO1/lego/legoomni/include/legocarbuild.h
```diff
@@ -48,7 +48,9 @@ class LegoVehicleBuildState : public LegoState {
 
 	// Known States:
 	// * 1 == enter(ing) build screen
+	// * 2 == ?
 	// * 3 == cutscene/dialogue
+	// * 4 == ?
 	// * 6 == exit(ing) build screen
```
foxtacles, line 54: Maybe we should add an enum for these values?

PR #1126, opened 2024-11-01
Implement `LegoCarBuild::FUN_10024890()` and others, refactor

jonschz:
All the newly implement functions match structurally, but not all to 100 %. There are also some refactorings. I will leave some comments on the code.

### DIFF
--- a/LEGO1/lego/legoomni/include/legogamestate.h
+++ b/LEGO1/lego/legoomni/include/legogamestate.h
@@ -172,13 +172,11 @@ public:
 
 	Act GetCurrentAct() { return m_currentAct; }
 	Act GetLoadedAct() { return m_loadedAct; }
-	Area GetCurrentArea() { return m_currentArea; }
 	Area GetPreviousArea() { return m_previousArea; }
 	Area GetUnknown0x42c() { return m_unk0x42c; }
 	History* GetHistory() { return &m_history; }
 
 	void SetDirty(MxBool p_isDirty) { m_isDirty = p_isDirty; }
-	void SetCurrentArea(Area p_currentArea) { m_currentArea = p_currentArea; }
 	void SetPreviousArea(Area p_previousArea) { m_previousArea = p_previousArea; }
 	void SetActorId(MxU8 p_actorId) { m_actorId = p_actorId; }
 	void SetUnknown0x42c(Area p_unk0x42c) { m_unk0x42c = p_unk0x42c; }
--- a/LEGO1/lego/legoomni/src/build/legocarbuild.cpp
+++ b/LEGO1/lego/legoomni/src/build/legocarbuild.cpp
@@ -51,7 +51,7 @@ LegoCarBuild::LegoCarBuild()
 {
 	m_unk0x100 = 0;
 	m_unk0x110 = 0;
-	m_unk0xf8 = 0xffffffff;
+	m_unk0xf8 = c_unknownminusone;
 	m_unk0x2d4 = FALSE;
 	m_unk0x258 = 0;
 	m_ColorBook_Bitmap = NULL;
@@ -431,7 +431,7 @@ MxResult LegoCarBuild::Tickle()
 		return SUCCESS;
 	}
 
-	if (m_unk0xf8 == 8) {
+	if (m_unk0xf8 == c_unknown8) {
 		if (m_unk0xfc == 1) {
 			FUN_10024f50();
 		}
--- a/LEGO1/lego/legoomni/src/build/legocarbuild.cpp
+++ b/LEGO1/lego/legoomni/src/build/legocarbuild.cpp
@@ -793,13 +793,213 @@ MxS32 LegoCarBuild::FUN_10024850(MxLong p_x, MxLong p_y)
 	return result;
 }
 
-// STUB: LEGO1 0x10024890
-// STUB: BETA10 0x1006d512
-undefined4 LegoCarBuild::FUN_10024890(LegoEventNotificationParam* p_param)
+#ifdef NDEBUG
+
+// FUNCTION: LEGO1 0x10024890
+undefined4 LegoCarBuild::FUN_10024890(MxParam* p_param)
 {
-	// TODO
-	return 0;
+	undefined4 result = 0;
+	LegoControlManagerNotificationParam* param = (LegoControlManagerNotificationParam*) p_param;
+	assert(m_buildState);
+
+	if (param->m_unk0x28) {
+		switch (param->m_clickedObjectId) {
+		// The enum values are all identical between CopterScript, DunecarScript, JetskiScript, and RacecarScript
+		case CopterScript::c_Info_Ctl:
+			if (m_buildState->m_animationState != LegoVehicleBuildState::e_unknown4 &&
+				m_buildState->m_animationState != LegoVehicleBuildState::e_unknown2 &&
+				m_buildState->m_animationState != LegoVehicleBuildState::e_exiting &&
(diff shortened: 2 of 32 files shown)


LEGO1/lego/legoomni/include/legogamestate.h
```diff
@@ -172,13 +172,11 @@ class LegoGameState {
 
 	Act GetCurrentAct() { return m_currentAct; }
 	Act GetLoadedAct() { return m_loadedAct; }
-	Area GetCurrentArea() { return m_currentArea; }
 	Area GetPreviousArea() { return m_previousArea; }
 	Area GetUnknown0x42c() { return m_unk0x42c; }
 	History* GetHistory() { return &m_history; }
 
 	void SetDirty(MxBool p_isDirty) { m_isDirty = p_isDirty; }
-	void SetCurrentArea(Area p_currentArea) { m_currentArea = p_currentArea; }
```
jonschz, line 181: See the comment in line 211 (old file). I verified a number of BETA10 matches and none of them used a getter or setter.

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
```diff
@@ -771,14 +793,214 @@ MxS32 LegoCarBuild::FUN_10024850(MxLong p_x, MxLong p_y)
 	return result;
 }
 
-// STUB: LEGO1 0x10024890
```
jonschz, line 774: I had to decompile the whole BETA10 function in order to get LEGO1 right, which had some logic differences with LEGO1. Since I didn't want to throw that away, I added it below with an `#ifdef`. I also considered a bunch of inline `#ifdef`s, but that turned out to be quite unreadable. Please tell me if you want me to remove the BETA10 function below.

LEGO1/library_msvc.h
```diff
@@ -19,6 +19,7 @@
 // _free
 
 // LIBRARY: LEGO1 0x1008b020
+// LIBRARY: BETA10 0x100f8d20
```
jonschz, line 22: This line alone improved the BETA10 match by about 2 %.

LEGO1/realtime/vector.h
```diff
@@ -301,10 +302,14 @@ class Vector4 : public Vector3 {
...
-	// Hack: Some code initializes a Vector4 from a (most likely) const float* source.
-	// Example: LegoCarBuild::VTable0x6c
+	// Some code initializes a Vector4 from a `const float*` source.
+	// Example: `LegoCarBuild::VTable0x6c`
 	// Vector4 however is a class that can mutate its underlying source, making
 	// initialization with a const source fundamentally incompatible.
+	// BETA10 appears to have two separate constructors for Vector4 as well,
+	// supporting the theory that this decompilation is correct.
+
+	// FUNCTION: BETA10 0x100701b0
```
jonschz, line 312: I found another Vector4::Vector4 implementation in BETA10, increasing the likelihood that the two constructors is indeed what the original code used.

LEGO1/lego/legoomni/src/common/mxtransitionmanager.cpp
```diff
@@ -84,20 +84,23 @@ MxResult MxTransitionManager::Tickle()
...
 	TransitionType p_animationType,
 	MxS32 p_speed,
 	MxBool p_doCopy,
 	MxBool p_playMusicInAnim
 )
 {
-	if (this->m_transitionType == e_notTransitioning) {
+	assert(m_mode == e_idle);
+
+	if (this->m_mode == e_idle) {
```
foxtacles, line 97: we should remove the use of this -> (below too)

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
```diff
@@ -26,6 +27,12 @@
 #include <isle.h>
 #include <vec.h>
 
+// names and values verified by BETA10 0x1006d742
```
foxtacles, line 30: we should add a comment here that we also have these constants in our generated actions files, for example:

https://github.com/isledecomp/isle/blob/master/LEGO1/lego/legoomni/include/actions/racecar_actions.h#L17

The original code probably didn't have that and if you can confirm they used macros like this instead we should use that, but adding a comment for reference would be good

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
```diff
@@ -771,14 +793,214 @@ MxS32 LegoCarBuild::FUN_10024850(MxLong p_x, MxLong p_y)
 	return result;
 }
 
-// STUB: LEGO1 0x10024890
-// STUB: BETA10 0x1006d512
-undefined4 LegoCarBuild::FUN_10024890(LegoEventNotificationParam* p_param)
+#ifdef NDEBUG
+
+// FUNCTION: LEGO1 0x10024890
+undefined4 LegoCarBuild::FUN_10024890(MxParam* p_param)
```
foxtacles, line 799: maybe this is MxNotificationParam instead of MxParam?

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
```diff
@@ -771,14 +793,214 @@ MxS32 LegoCarBuild::FUN_10024850(MxLong p_x, MxLong p_y)
...
+				if (GameState()->GetCurrentAct() == LegoGameState::e_act2) {
+					FUN_100243a0();
+				}
+				else if (m_unk0x258->AllPartsPlaced() || m_buildState->m_unk0x4d) {
+					m_buildState->m_unk0x4d = TRUE;
+					InvokeAction(Extra::e_start, m_atomId, m_carId, NULL);
+
+					{
+						// This pattern is suggested by BETA10 and also improves the LEGO1 match
+						MxNotificationParam param;
```
foxtacles, line 842: Can param be directly instantiated within the function call instead of wrapping this in a scope, without worsening the match? This may not compile on modern compilers (we'd have to ifdef around that keeping this version for them) but seems more likely that they did that (have done so in other places)

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
jonschz: I cannot prove that they used macros; the assertions look like https://github.com/isledecomp/isle/blob/9ae3e78836b1336ee1777c82329cdb3cd5fcea71/LEGO1/lego/legoomni/src/build/legocarbuild.cpp#L947 and I am not sure how that would work with enums. I'll add the comment.

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
jonschz: Possible, but the beta strongly suggests that there is a typecast right at the beginning of the function, so the type here should not matter at all.

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
jonschz: I went for something a little more self-documenting, please tell me how you like it

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
jonschz: Great suggestion, this got the function to 100% - nice!

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
foxtacles: Looks good thanks!

PR #1134, opened 2024-11-04
Implement `FUN_10025720()` and others

jonschz:
`FUN_10025720` matches up to one minor detail, the rest matches to 100 %.

### DIFF
--- a/LEGO1/lego/legoomni/include/legocarbuild.h
+++ b/LEGO1/lego/legoomni/include/legocarbuild.h
@@ -74,14 +74,14 @@ typedef LegoVehicleBuildState LegoJetskiBuildState;
 class LegoCarBuild : public LegoWorld {
 public:
 	// SIZE 0x1c
-	struct LookupTableData {
-		undefined2 m_unk0x00; // 0x00
+	struct LookupTableActions {
+		undefined4 m_unk0x00; // 0x00
 		undefined4 m_unk0x04; // 0x04
 		undefined4 m_unk0x08; // 0x08
 		undefined4 m_unk0x0c; // 0x0c
 		undefined4 m_unk0x10; // 0x10
 		undefined4 m_unk0x14; // 0x14
-		MxU16 m_unk0x18;      // 0x18
+		undefined4 m_unk0x18; // 0x18
 	};
 
 	enum Unknown0xf8 {
@@ -239,7 +239,7 @@ private:
 	static MxS16 g_unk0x100f11cc;
 	static MxFloat g_unk0x100d65a4;
 	static MxFloat g_rotationAngleStepYAxis;
-	static LookupTableData g_unk0x100d65b0[];
+	static LookupTableActions g_unk0x100d65b0[];
 };
 
 #endif // LEGOCARBUILD_H
--- a/LEGO1/lego/legoomni/include/legocarbuild.h
+++ b/LEGO1/lego/legoomni/include/legocarbuild.h
@@ -73,6 +73,17 @@ typedef LegoVehicleBuildState LegoJetskiBuildState;
 // SIZE 0x34c
 class LegoCarBuild : public LegoWorld {
 public:
+	// SIZE 0x1c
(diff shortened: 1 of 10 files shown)


LEGO1/lego/legoomni/src/common/legoutils.cpp
```diff
@@ -31,6 +31,7 @@
 #include <string.h>
 #include <vec.h>
 
+
```
foxtacles, line 34: no need for this extra whitespace I guess:)

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
```diff
@@ -1276,12 +1346,179 @@ void LegoCarBuild::Enable(MxBool p_enable)
...
-undefined4 LegoCarBuild::FUN_10025720(undefined4 p_param1)
+// FUNCTION: BETA10 0x10070520
+inline MxU32 LegoCarBuild::Beta0x10070520()
 {
-	// TODO
-	return 0;
+	switch (m_carId) {
+	case Helicopter_Actor:
+		return 2;
+		break;
```
foxtacles, line 1355: those breaks after the returns should not be necessary

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
```diff
@@ -39,6 +40,42 @@
 
 DECOMP_SIZE_ASSERT(LegoCarBuild, 0x34c)
 DECOMP_SIZE_ASSERT(LegoVehicleBuildState, 0x50)
+DECOMP_SIZE_ASSERT(LegoCarBuild::LookupTableData, 0x1c);
+
+// These four structs can be matched to the vehicle types using BETA10 0x10070520
+
+// GLOBAL: LEGO1 0x100d65b0
+// GLOBAL: BETA10 0x101bb7c0
+LegoCarBuild::LookupTableData LegoCarBuild::g_unk0x100d65b0[] = {
```
foxtacles, line 49: Maybe we should call this struct something like LookupTableActions or LookupTableScripts?

LEGO1/lego/legoomni/include/legocarbuild.h
```diff
@@ -73,6 +73,17 @@ typedef LegoVehicleBuildState LegoJetskiBuildState;
 // SIZE 0x34c
 class LegoCarBuild : public LegoWorld {
 public:
+	// SIZE 0x1c
+	struct LookupTableData {
+		undefined2 m_unk0x00; // 0x00
```
foxtacles, line 78: The different sizes of the members are really odd considering they are all just script/entity IDs, but I guess it's verified to be correct?

jonschz: Your comment made me investigate a bit more and I managed to get a 100 % match. While changing the data types did improve the match a while ago, it didn't end up mattering for the final result. For example, the code is identical for `undefined2 m_unk0x00` and `undefined4 m_unk0x00`.

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
```diff
@@ -1276,12 +1346,172 @@ void LegoCarBuild::Enable(MxBool p_enable)
...
+			break;
+		case 4:
+			FUN_10025d10(g_unk0x100d65b0[Beta0x10070520()].m_unk0x04);
+			break;
+		case 5:
+			FUN_10025d10(g_unk0x100d65b0[Beta0x10070520()].m_unk0x08);
+			break;
+		case 6:
+			m_unk0x10a = g_unk0x100d65b0[Beta0x10070520()].m_unk0x18;
+			uVar6 = m_unk0x10a;
```
jonschz, line 1428: Removing `uVar6` here does decrease the match percentage, but I find its use confusing. Maybe there is an inline function that originally took an `MxS32`. Hard to tell because BETA10 does not have this part at all.

PR #1136, opened 2024-11-09
Implement Doors::VTable0x94

ItzSwirlz:
I need some help with this - specifically the "add +0x10" part. I think it's saying to use MxMatrix's GetData function at vtable+0x10, but also trying to call it will use vtable+0x14 and technically it doesn't specifically call that method.. so I'm not sure.


foxtacles:
The `add +0x10` instructions add an offset to the base pointers, which in this case means it is passing `Matrix4` objects located elsewhere. The type of `m_unk0x15c` and `m_unk0x160` might actually be different than `Matrix4*` but we don't know that until more code is decompiled. For now it matches 100%.

PR #1149, opened 2024-11-13
Implement most of `JetskiRace`, add BETA10 annotations

jonschz:
The new functions match to (effectively) 100 %.

I have left some comments on a few questionable parts.

Looking forward to your review!

### DIFF
--- a/LEGO1/lego/legoomni/src/race/jetskirace.cpp
+++ b/LEGO1/lego/legoomni/src/race/jetskirace.cpp
@@ -1,28 +1,168 @@
 #include "jetskirace.h"
 
+#include "actions/jetrace_actions.h"
+#include "actions/jukebox_actions.h"
+#include "dunebuggy.h"
+#include "isle.h"
+#include "legoanimationmanager.h"
+#include "legocontrolmanager.h"
+#include "legohideanimpresenter.h"
+#include "legomain.h"
+#include "legopathstruct.h"
+#include "legoracespecial.h"
+#include "legoutils.h"
+#include "misc.h"
+#include "mxactionnotificationparam.h"
+#include "mxbackgroundaudiomanager.h"
+#include "mxmisc.h"
+#include "mxstillpresenter.h"
+#include "mxtransitionmanager.h"
+#include "mxvariabletable.h"
+#include "scripts.h"
+
+// Defined in legopathstruct.cpp
+extern MxBool g_unk0x100f119c;
+
+// Defined in jetski.cpp
+extern const char* g_varJSFRNTY5;
+extern const char* g_varJSWNSHY5;
+
+// Defined in legoracespecial.cpp
+extern const char* g_raceState;
+extern const char* g_racing;
+
+// Defined in legopathactor.cpp
+extern const char* g_strHIT_WALL_SOUND;
+
+// GLOBAL: LEGO1 0x100f0bac
+static undefined4 g_unk0x100f0bac = 0;
+
+// GLOBAL: LEGO1 0x100f0bb0
+static undefined4 g_unk0x100f0bb0 = 0;
+
 DECOMP_SIZE_ASSERT(JetskiRace, 0x144)
 
-// STUB: LEGO1 0x100162c0
+// FUNCTION: LEGO1 0x100162c0
+// FUNCTION: BETA10 0x100c7e6f
 MxResult JetskiRace::Create(MxDSAction& p_dsAction)
 {
-	return SUCCESS;
+	MxResult result = LegoRace::Create(p_dsAction);
+
+	GameState()->m_currentArea = LegoGameState::e_jetrace;
+	GameState()->StopArea(LegoGameState::e_undefined);
+	LegoGameState* gameState = GameState();
+	RaceState* jetskiRaceState = (RaceState*) gameState->GetState("JetskiRaceState");
+
+	if (!jetskiRaceState) {
+		jetskiRaceState = (RaceState*) gameState->CreateState("JetskiRaceState");
+	}
+
+	m_raceState = jetskiRaceState;
+
+	if (!jetskiRaceState) {
+		return FAILURE;
+	}
+
+	m_raceState->m_unk0x28 = 1;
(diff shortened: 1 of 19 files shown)


LEGO1/lego/legoomni/src/race/jetskirace.cpp
```diff
@@ -1,28 +1,168 @@
...
+// Defined in jetski.cpp
+extern const char* g_varJSFRNTY5;
+extern const char* g_varJSWNSHY5;
+
+// Defined in legoracespecial.cpp
+extern const char* g_raceState;
+extern const char* g_racing;
+
+// Defined in legopathactor.cpp
+extern const char* g_strHIT_WALL_SOUND;
```
jonschz, line 35: Not sure if we want to stick to this pattern or make these available in the header file. I've seen both.

LEGO1/lego/legoomni/src/race/legoracespecial.cpp
```diff
@@ -12,6 +12,18 @@
 
 // File name verified by BETA10 0x100cedf7
 
+
+// FUNCTION: LEGO1 0x10012de0
+void LegoCarRaceActor::FUN_10012de0()
+{
+	// This function is probably in the wrong place.
+	// According to the address it should be in legoracespecial.h.
+	// However, if we put it there, it gets inlined instead of called.
```
jonschz, line 21: This is my biggest question. Any idea on how to get this function to be linked somewhere within `legoracespecial.h` without it getting inlined into calling functions?

jonschz:
Overall, the discrepancy to BETA10 is so large that I was unable to match some functions in the vtable to LEGO1.

LEGO1/lego/legoomni/src/race/legoracespecial.cpp
foxtacles: Maybe I'm missing something, but according to the address this function should be located between `LegoRaceCar::Notify` and `LegoRaceCar::SetWorldSpeed`, no? I know this would mean the function belongs to `LegoRaceCar` instead of `LegoCarRaceActor` but that makes sense.

The 3 static variables that are set inside the function should then also be moved to `legoracers.cpp`

jonschz: Yes, you are absolutely right. I somehow missed that this function is indeed located in the middle of a cpp file, I only saw the rather close functions in header files. Should be fixed now.

LEGO1/lego/legoomni/src/race/jetskirace.cpp
foxtacles: I don't have a strong opinion on either, so I think we can leave it like this for now. I think it's better practice to put them in respective header files but we can refactor this later.

LEGO1/lego/legoomni/src/race/jetskirace.cpp
```diff
@@ -1,28 +1,153 @@
...
+	m_unk0x130.SetRight(543);
+	m_unk0x130.SetBottom(333);
+	LegoRaceCar::FUN_10013670();
+	InvokeAction(
+		Extra::e_start,
+		m_atomId,
+		DuneBuggy::GetColorOffset(g_varJSFRNTY5) + (DuneBuggy::GetColorOffset(g_varJSWNSHY5) * 5 + 0xf) * 2,
+		NULL
+	);
+	InvokeAction(Extra::e_start, m_atomId, 0x61, NULL);
```
foxtacles, line 73: `0x61` should probably be `JetskiScript::c_JSFRNT_T1_Up_Bitmap`

LEGO1/lego/legoomni/src/race/jetskirace.cpp
```diff
@@ -1,28 +1,153 @@
...
+	if (((LegoControlManagerNotificationParam*) &p_param)->m_unk0x28 == 1) {
+		switch (((LegoControlManagerNotificationParam*) &p_param)->m_clickedObjectId) {
+		case JetraceScript::c_JetskiArms_Ctl:
+			m_act1State->m_unk0x018 = 0;
+			VariableTable()->SetVariable(g_raceState, "");
+			VariableTable()->SetVariable(g_strHIT_WALL_SOUND, "");
+			LegoRaceCar::FUN_10012de0();
+			m_destLocation = LegoGameState::e_jetraceExterior;
+			TransitionManager()->StartTransition(MxTransitionManager::e_mosaic, 50, FALSE, FALSE);
+			return 0;
```
foxtacles, line 136: This should be `break;` instead (for a 100% match)

LEGO1/lego/legoomni/src/race/jetskirace.cpp
```diff
@@ -1,28 +1,153 @@
...
+			m_act1State->m_unk0x018 = 0;
+			VariableTable()->SetVariable(g_raceState, "");
+			VariableTable()->SetVariable(g_strHIT_WALL_SOUND, "");
+			LegoRaceCar::FUN_10012de0();
+			m_destLocation = LegoGameState::e_infomain;
+			result = 1;
+			TransitionManager()->StartTransition(MxTransitionManager::e_mosaic, 50, FALSE, FALSE);
+			break;
+		default:
+			return 0;
```
foxtacles, line 147: This should be `break;` instead (for a 100% match)

LEGO1/lego/legoomni/src/race/jetskirace.cpp
jonschz: Maybe also `JetraceScript::c_JetskiDashboard`? I left in the `0x61` because no candidate looked really obvious.

foxtacles: Ah yes, then I'm pretty sure it's `JetraceScript::c_JetskiDashboard`

foxtacles: Ah yes, I'm pretty sure it would have to be that one. It's the same as in `Jetski::HandleClick`

PR #1159, opened 2024-11-18
Implement parts of LegoJetskiRaceActor

jonschz:
The new functions match to 100 % and also match BETA10 structurally (when available).

### DIFF
--- a/LEGO1/lego/legoomni/src/entity/legojetskiraceactor.cpp
+++ b/LEGO1/lego/legoomni/src/entity/legojetskiraceactor.cpp
@@ -2,10 +2,9 @@
 
 #include "legonavcontroller.h"
 #include "legopathcontroller.h"
+#include "misc.h"
 #include "mxmisc.h"
 #include "mxvariabletable.h"
-
-#include <misc.h>
 
 DECOMP_SIZE_ASSERT(LegoJetskiRaceActor, 0x1a8)
 
--- a/LEGO1/lego/legoomni/src/entity/legojetskiraceactor.cpp
+++ b/LEGO1/lego/legoomni/src/entity/legojetskiraceactor.cpp
@@ -1,8 +1,20 @@
 #include "legojetskiraceactor.h"
 
+#include "legonavcontroller.h"
+#include "legopathcontroller.h"
+#include "mxmisc.h"
+#include "mxvariabletable.h"
+
+#include <misc.h>
+
 DECOMP_SIZE_ASSERT(LegoJetskiRaceActor, 0x1a8)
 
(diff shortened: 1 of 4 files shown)


LEGO1/lego/legoomni/src/entity/legojetskiraceactor.cpp
```diff
@@ -11,17 +22,120 @@ LegoJetskiRaceActor::LegoJetskiRaceActor()
...
 }
 
-// STUB: LEGO1 0x10081550
+// FUNCTION: LEGO1 0x10081550
 void LegoJetskiRaceActor::VTable0x70(float p_float)
 {
-	// TODO
+	if (m_unk0x0c == 0) {
+		const LegoChar* raceState = VariableTable()->GetVariable(g_raceState);
+		if (stricmp(raceState, g_racing) == 0x0) {
```
foxtacles, line 126: I think it's clearer to just write it as `!stricmp(raceState, g_racing)` (the way it's done in other places)

LEGO1/lego/legoomni/src/entity/legojetskiraceactor.cpp
```diff
@@ -11,17 +22,120 @@ LegoJetskiRaceActor::LegoJetskiRaceActor()
...
+		if (m_state == 1) {
+			if (m_userNavFlag) {
+				m_unk0xe4 = 0.5f;
+			}
+
+			v1 = m_destEdge->CCWVertex(*m_boundary);
+			v2 = m_destEdge->CWVertex(*m_boundary);
+			assert(v1 && v2);
+
+			a[0] = (*v1)[0] + ((*v2)[0] - (*v1)[0]) * m_unk0xe4;
```
foxtacles, line 88: This looks like `LERP3` in `vec.h`

LEGO1/lego/legoomni/src/entity/legojetskiraceactor.cpp
jonschz: Funny, I personally find the comparison with zero clearer, but that's a matter of opinion (to me, `!stricmp` feels like it says "does not match" while meaning the opposite). I noticed that the rest of the repo isn't entirely consistent there either, but the exclamation mark pattern is much more prevalent. I'll change it.

LEGO1/lego/legoomni/src/entity/legojetskiraceactor.cpp
jonschz: This has had a negative effect on the stack match in BETA10, but the refactor definitely looks cleaner and LEGO1 is unaffected.

PR #1165, opened 2024-11-23
Implement `CarRace`

jonschz:
Most of the new functions match to 100 %, there are small deviations for `CarRace::Create()` and `CarRace::HandlePathStruct()` - input is appreciated. I also added a number of BETA10 annotations and improved `JetskiRace::HandlePathStruct()`.

The car race can now be started, but it isn't really playable because the doors cannot be opened yet.

### DIFF
--- a/LEGO1/lego/legoomni/include/legoinputmanager.h
+++ b/LEGO1/lego/legoomni/include/legoinputmanager.h
@@ -132,6 +132,13 @@ public:
 	void GetKeyboardState();
 	MxResult GetNavigationKeyStates(MxU32& p_keyFlags);
 
+	void Reset88and336()
+	{
+		// This function is hypothesized based on LEGO1 0x100173f5
+		m_unk0x88 = TRUE;
+		m_unk0x336 = FALSE;
+	}
+
 	// SYNTHETIC: LEGO1 0x1005b8d0
 	// LegoInputManager::`scalar deleting destructor'
 
--- a/LEGO1/lego/legoomni/include/legoinputmanager.h
+++ b/LEGO1/lego/legoomni/include/legoinputmanager.h
@@ -132,13 +132,6 @@ public:
 	void GetKeyboardState();
 	MxResult GetNavigationKeyStates(MxU32& p_keyFlags);
 
-	void Reset88and336()
-	{
-		// This function is hypothesized based on LEGO1 0x100173f5
-		m_unk0x88 = TRUE;
-		m_unk0x336 = FALSE;
-	}
-
 	// SYNTHETIC: LEGO1 0x1005b8d0
 	// LegoInputManager::`scalar deleting destructor'
 
--- a/LEGO1/lego/legoomni/src/race/carrace.cpp
+++ b/LEGO1/lego/legoomni/src/race/carrace.cpp
@@ -1,12 +1,14 @@
 #include "carrace.h"
 
 #include "actions/carrace_actions.h"
+#include "actions/jukebox_actions.h"
 #include "dunebuggy.h"
 #include "isle.h"
 #include "legoanimationmanager.h"
 #include "legobackgroundcolor.h"
 #include "legocontrolmanager.h"
 #include "legohideanimpresenter.h"
+#include "legomain.h"
 #include "legonavcontroller.h"
 #include "legopathstruct.h"
 #include "legoracers.h"
@@ -106,8 +108,7 @@ MxResult CarRace::Create(MxDSAction& p_dsAction)
 	InvokeAction(Extra::e_start, m_atomId, streamId, NULL);
 	InvokeAction(Extra::e_start, m_atomId, CarraceScript::c_RaceCarDashboard, NULL);
 
-	// TODO
-	return SUCCESS;
+	return result;
 }
 
 // FUNCTION: LEGO1 0x10016dd0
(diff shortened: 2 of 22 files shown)


LEGO1/lego/legoomni/src/race/carrace.cpp
```diff
@@ -1,60 +1,418 @@
...
+							m_unk0x14c = g_unk0x100d5d40[rand() % 3];
+						}
+					}
+					else {
+						position = 1;
+						m_unk0x148 = g_unk0x100d5d30[rand() % 4];
+						m_unk0x14c = g_unk0x100d5d50[rand() % 3];
+					}
+
+					InputManager()->Reset88and336();
```
jonschz, line 240: Something doesn't match perfectly around this area I think

reccmp-project.yml
```diff
@@ -19,3 +19,14 @@ targets:
...
+      ignore-types:
+        # these classes have been changed by hand to account for changes between LEGO1 and BETA10
+        - LegoCarBuild
+        - LegoCarBuildAnimPresenter
+        - LegoRace
+        - LegoWorld
+      ignore-functions:
+        # strcpy, strlen, ... (arguments are imported incorrectly)
+        - 0x100fa200
+        - 0x100f9780
```
jonschz, line 32: I added these now that https://github.com/isledecomp/reccmp/pull/19 is merged

jonschz:
Update: Now with `Doors::VTable0xcc()` implemented, the car race is fully playable.

LEGO1/lego/legoomni/src/race/carrace.cpp
```diff
@@ -1,60 +1,418 @@
...
+#include "mxnotificationmanager.h"
+#include "mxstillpresenter.h"
+#include "mxtransitionmanager.h"
+#include "mxvariabletable.h"
+#include "scripts.h"
 
 DECOMP_SIZE_ASSERT(CarRace, 0x154)
 
+// GLOBAL: LEGO1 0x100d5d10
+const MxS32 CarRace::g_unk0x100d5d10[] = {
```
foxtacles, line 27: I know `const` is generally the right qualifier here, but we haven't used it anywhere else so far for those global variables so I think I'd remove it simply for consistency - unless that worsens the match somewhere?

LEGO1/lego/legoomni/src/race/carrace.cpp
```diff
@@ -1,60 +1,418 @@
...
+	m_unk0x14c = -1;
+
+	LegoRaceCar::FUN_10012e00();
+
+	MxS32 streamId =
+		DuneBuggy::GetColorOffset(g_strCRCEDGEY0) + (DuneBuggy::GetColorOffset(g_strCRCFRNTY6) * 5 + 15) * 2;
+	InvokeAction(Extra::e_start, m_atomId, streamId, NULL);
+	InvokeAction(Extra::e_start, m_atomId, CarraceScript::c_RaceCarDashboard, NULL);
+
 	// TODO
```
foxtacles, line 109: To get the match to 100%, you have to `return result;` - the TODO comment can go away too I guess!

LEGO1/lego/legoomni/include/legoinputmanager.h
```diff
@@ -132,6 +132,13 @@ class LegoInputManager : public MxPresenter {
 	void GetKeyboardState();
 	MxResult GetNavigationKeyStates(MxU32& p_keyFlags);
 
+	void Reset88and336()
```
foxtacles, line 135: We already have this function: `DisableInputProcessing`

LEGO1/lego/legoomni/src/race/carrace.cpp
```diff
@@ -1,60 +1,418 @@
...
 void CarRace::ReadyWorld()
 {
-	// TODO
+	assert(m_hideAnim);
+	LegoWorld::ReadyWorld();
+	m_hideAnim->FUN_1006db40(0);
+
+	MxDSAction action;
+	action.SetAtomId(*g_jukeboxScript);
+	action.SetObjectId(16);
```
foxtacles, line 123: Should be the `JukeboxScript::c_RaceTrackRoad_Music` constant

LEGO1/lego/legoomni/src/race/carrace.cpp
```diff
@@ -1,60 +1,418 @@
...
+
+	BackgroundAudioManager()->PlayMusic(action, 5, MxPresenter::e_repeating);
+	AnimationManager()->Resume();
+	FUN_10015820(FALSE, 7);
```
foxtacles, line 127: The constant `7` should be replaced with `LegoOmni::c_disableInput | LegoOmni::c_disable3d | LegoOmni::c_clearScreen`. There's also another in `legocarbuild.cpp` where you could fix that. Thank you!

LEGO1/lego/legoomni/src/race/carrace.cpp
```diff
@@ -1,60 +1,418 @@
...
+				result = 1;
+			}
+
+			break;
+		case 11:
+			if (paramData <= m_unk0x108 || paramData >= m_unk0x108 + 5) {
+				break;
+			}
+
+			FUN_10017820(0xb, paramData);
```
foxtacles, line 267: To be consistent I'd use the decimal `11` instead of `0xb` since the function itself uses those values (same with the other instance of that call)

foxtacles:
Great work, added some comments. Not sure about `HandlePathStruct` myself but looks good enough to me.

LEGO1/lego/legoomni/include/legoinputmanager.h
jonschz: Can't believe I missed that - thanks!

LEGO1/lego/legoomni/src/race/carrace.cpp
jonschz: I removed the `const`s and I didn't see any (significant) regressions, so it's likely fine. Thanks!

PR #1168, opened 2024-11-25
Implement `LegoAct2::Tickle()` and others

jonschz:
The new functions match to 100 % effectively. I also refactored the `GetState()` / `SetState()` function calls a bit.

### DIFF
--- a/LEGO1/lego/legoomni/include/legoanimationmanager.h
+++ b/LEGO1/lego/legoomni/include/legoanimationmanager.h
@@ -155,7 +155,7 @@ public:
 	MxBool FindVehicle(const char* p_name, MxU32& p_index);
 	MxResult ReadAnimInfo(LegoFile* p_file, AnimInfo* p_info);
 	MxResult ReadModelInfo(LegoFile* p_file, ModelInfo* p_info);
-	void FUN_10060480(LegoChar* p_param1[], int p_param2);
+	void FUN_10060480(LegoChar* p_param1[], undefined4 p_param2);
 	void FUN_100604d0(MxBool p_unk0x08);
 	void FUN_100604f0(MxS32* p_param1, undefined4 p_param2);
 	void FUN_10060540(MxBool p_unk0x29);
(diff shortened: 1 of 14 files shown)


LEGO1/lego/legoomni/include/legoanimationmanager.h
```diff
@@ -155,7 +155,9 @@ class LegoAnimationManager : public MxCore {
 	MxBool FindVehicle(const char* p_name, MxU32& p_index);
 	MxResult ReadAnimInfo(LegoFile* p_file, AnimInfo* p_info);
 	MxResult ReadModelInfo(LegoFile* p_file, ModelInfo* p_info);
+	void FUN_10060480(LegoChar* p_param1[], int p_param2);
```
foxtacles, line 158: Shouldn't use `int` here (`MxS32` if you are sure it's that? otherwise `undefined4`)

LEGO1/lego/legoomni/src/worlds/legoact2.cpp
```diff
@@ -90,14 +129,116 @@ MxResult LegoAct2::Create(MxDSAction& p_dsAction)
...
+// FUNCTION: BETA10 0x1003a976
 MxResult LegoAct2::Tickle()
 {
-	// TODO
-	return SUCCESS;
+	MxFloat distance;
+
+	if (!m_worldStarted) {
+		LegoWorld::Tickle();
+		return 0;
```
foxtacles, line 140: should be `return SUCCESS`

LEGO1/lego/legoomni/src/worlds/legoact2.cpp
```diff
@@ -90,14 +129,116 @@ MxResult LegoAct2::Create(MxDSAction& p_dsAction)
...
+		break;
+	case 3:
+		FUN_10015820(FALSE, LegoOmni::c_disableInput | LegoOmni::c_disable3d | LegoOmni::c_clearScreen);
+		m_unk0x10d0 = 0;
+		m_unk0x10c4 = 4;
+		FUN_10052560(Act2mainScript::c_tja009ni_RunAnim, TRUE, TRUE, NULL, NULL, NULL);
+
+		AnimationManager()->EnableCamAnims(TRUE);
+		AnimationManager()->FUN_1005f6d0(TRUE);
+		AnimationManager()->FUN_100604f0(&g_unk0x100f43f0[0], sizeOfArray(g_unk0x100f43f0));
```
foxtacles, line 184: would just write `g_unk0x100f43f0` here instead of `&g_unk0x100f43f0[0]` (same as in the next line)

foxtacles (commented):
Looks good!

LEGO1/lego/legoomni/src/worlds/legoact2.cpp
jonschz: Not sure why that didn't work before, but now it does - thanks!

PR #1170, opened 2024-11-26
Implement new functions in LegoAnimManager

jonschz:
The new functions match correctly, but `FUN_100604f0` seems to be susceptible to entropy - it matches to 100 % on BETA10 and it looks correct on LEGO1.

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legoanimmmpresenter.cpp
+++ b/LEGO1/lego/legoomni/src/common/legoanimmmpresenter.cpp
@@ -466,6 +466,12 @@ MxBool LegoAnimMMPresenter::FUN_1004b6d0(MxLong p_time)
 	return TRUE;
 }
 
+// FUNCTION: LEGO1 0x1004b830
+MxBool LegoAnimMMPresenter::FUN_1004b830()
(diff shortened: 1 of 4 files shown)


LEGO1/lego/legoomni/src/common/legoanimmmpresenter.cpp
```diff
@@ -466,6 +466,12 @@ MxBool LegoAnimMMPresenter::FUN_1004b6d0(MxLong p_time)
 	return TRUE;
 }
 
+// FUNCTION: LEGO1 0x1004b830
+MxBool LegoAnimMMPresenter::FUN_1004b830()
+{
+	return !(m_unk0x58 < 6);
```
foxtacles, line 472: I think writing this as: `m_unk0x58 >= e_unk6` is clearer - `m_unk0x58` is some sort of state progression and this checks that it is in e_unk6 or later

jonschz: Thanks! I should have tried this. It's quite odd how the compiler chooses to compile this line.

PR #1174, opened 2024-11-29
Implement/match `LegoAct2::FUN_10052560()`

jonschz:
There is only one slight mismatch which I couldn't work out, the rest matches exactly.

### DIFF
--- a/LEGO1/lego/legoomni/src/worlds/legoact2.cpp
+++ b/LEGO1/lego/legoomni/src/worlds/legoact2.cpp
@@ -360,17 +360,17 @@ undefined4 LegoAct2::FUN_10052560(
 				oneVectorNotNull = TRUE;
 			}
 
-			Vector3 local_114(matrix[0]);
-			Vector3 local_124(matrix[1]);
-			Vector3 local_10c(matrix[2]);
+			Vector3 firstColumn(matrix[0]);
+			Vector3 secondColumn(matrix[1]);
+			Vector3 thirdColumn(matrix[2]);
 
-			local_114.EqualsCross(&local_124, &local_10c);
-			local_114.Unitize();
+			firstColumn.EqualsCross(&secondColumn, &thirdColumn);
+			firstColumn.Unitize();
 
-			MxMatrix* p_matrix = NULL;
+			MxMatrix* pmatrix = NULL;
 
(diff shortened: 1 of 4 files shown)


LEGO1/lego/legoomni/src/worlds/legoact2.cpp
```diff
@@ -299,17 +300,90 @@ MxBool LegoAct2::Escape()
...
+			firstColumn.EqualsCross(&secondColumn, &thirdColumn);
+			firstColumn.Unitize();
+
+			MxMatrix* pmatrix = NULL;
+
+			if (oneVectorNotNull) {
+				pmatrix = &matrix;
+			}
+
+			// This part does not match perfectly
```
foxtacles, line 376: You can match it completely with this structure:

```
			MxResult result;
			if (p_param1 == Act2mainScript::c_tja009ni_RunAnim) {
				result = AnimationManager()->FUN_10060dc0(p_param1, pmatrix, TRUE, FALSE, NULL, TRUE, TRUE, TRUE, TRUE);
			}
			else {
				result =
					AnimationManager()->FUN_10060dc0(p_param1, pmatrix, TRUE, FALSE, NULL, TRUE, TRUE, TRUE, FALSE);
			}
```

The compiler optimizes the code so that there's only one call in the assembly at the end.

LEGO1/lego/legoomni/src/worlds/legoact2.cpp
```diff
@@ -299,17 +300,90 @@ MxBool LegoAct2::Escape()
...
+			MxBool arg9 = FALSE;
+			if (p_param1 == Act2mainScript::c_tja009ni_RunAnim) {
+				arg9 = TRUE;
+			}
+
+			if (!AnimationManager()->FUN_10060dc0(p_param1, pmatrix, TRUE, FALSE, NULL, TRUE, TRUE, TRUE, arg9)) {
```
foxtacles, line 382: If you follow the suggestion I placed above, this should then be: `if (result == SUCCESS)`

LEGO1/lego/legoomni/src/worlds/legoact2.cpp
jonschz: nice, thanks!

PR #1175, opened 2024-11-29
Implement/match `LegoAct2::Notify`

foxtacles:
Match is at ~87%, but the remaining diff seems to be an issue with the stack layout only.

@jonschz Am I using `stackcmp` correctly? I can't seem to get it to work:

```
PS C:\Users\Christian\Downloads\isle> reccmp-stackcmp --target LEGO1 0x10050380
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "c:\users\christian\appdata\local\packages\pythonsoftwarefoundation.python.3.11_qbz5n2kfra8p0\localcache\local-packages\python311\scripts\reccmp-stackcmp.exe\__main__.py", line 7, in <module>
  File "C:\Users\Christian\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\site-packages\reccmp\tools\stackcmp.py", line 322, in main
    origfile, recompfile, target.pdb_path, target.source_root
                          ^^^^^^^^^^^^^^^
AttributeError: 'RecCmpBuiltTarget' object has no attribute 'pdb_path'
```

### DIFF
--- a/LEGO1/lego/legoomni/include/legoact2.h
+++ b/LEGO1/lego/legoomni/include/legoact2.h
@@ -6,6 +6,8 @@
 #include "legoworld.h"
 
 class Act2Actor;
+class LegoPathStructNotificationParam;
+class MxEndActionNotificationParam;
 
 // VTABLE: LEGO1 0x100d4a70
 // SIZE 0x10
@@ -79,6 +81,9 @@ public:
 	// LegoAct2::`scalar deleting destructor'
 
 private:
+	MxLong HandleEndAction(MxEndActionNotificationParam& p_param);
+	MxLong HandleTransitionEnd();
+	MxLong HandlePathStruct(LegoPathStructNotificationParam& p_param);
 	void FUN_10051900();
 
 	Act2Brick m_bricks[10];      // 0x00f8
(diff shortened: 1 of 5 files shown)


jonschz:
It looks like you're not on the latest version of `reccmp` - the error message does not match the current code on `master`. Everyone's setup is different, I have chosen to install my local copy of `reccmp` into my `isle` venv in editable mode, which works quite well. Maybe an update will be sufficient to fix it.

Note that `stackcmp` works best with BETA10, where all stack references are of the form `ebp - [number]`. LEGO1 also uses `[esp + [number]`, which is harder to parse correctly and thus has some known issues. I will open a ticket over at `reccmp`.

On a different note, I think we should coordinate the remaining efforts now that only a couple of functions are left. I had already started working on `LegoAct2::Notify`, though I hadn't spent a lot of time on it yet, fortunately.

foxtacles:
> It looks like you're not on the latest version of `reccmp` - the error message does not match the current code on `master`. Everyone's setup is different, I have chosen to install my local copy of `reccmp` into my `isle` venv in editable mode, which works quite well. Maybe an update will be sufficient to fix it.
> 
> Note that `stackcmp` works best with BETA10, where all stack references are of the form `ebp - [number]`. LEGO1 also uses `esp + [number]`, which is harder to parse correctly and thus has some known issues. I will open a ticket over at `reccmp` (edit: [isledecomp/reccmp#23](https://github.com/isledecomp/reccmp/issues/23))

Thanks, I got it working. For future reference, I've used this command to force a reinstall/upgrade:

`pip install --upgrade --force-reinstall -r tools/requirements.txt`

> On a different note, I think we should coordinate the remaining efforts now that only a couple of functions are left. I had already started working on `LegoAct2::Notify`, though I hadn't spent a lot of time on it yet, fortunately.

Yeah, I was about to text you regarding this on the Matrix chat!

PR #1177, opened 2024-11-29
Clean up unused annotations to reduce errors

jonschz:
This PR has no match difference compared to `master`.

It reduces the errors raised on a `reccmp` run down to the following (with some minor changes in reccmp's logging that I'll PR soon):
```
[ERROR] Failed to find function symbol with filename and line: C:\Users\Jonathan\vscode-workspace\isle\LEGO1\lego\sources\anim\legoanim.cpp:788
[ERROR] Failed to find function symbol with name annotated with 0x10092310: _siglookup
[ERROR] Failed to find function symbol with name annotated with 0x10022360: ?_Construct@@YAXPAPAVMxCore@@ABQAV1@@Z
[ERROR] Failed to find function symbol with name annotated with 0x100c7ef0: list<MxNextActionDataStart *>::insert
[ERROR] Failed to find function symbol with name annotated with 0x100c1bc0: list<MxDSAction *,allocator<MxDSAction *> >::insert
[ERROR] Failed to find string annotated with 0x100fd200: '.'
```
- I can't explain the first one.
- Errors 2, 3, 4, and 5 appear to be functions that are matched correctly but are not called at the moment, and the compiler does not include them in the build.
- The last one looks like an error in `reccmp`. Removing it or converting it into a `STRING` reduces the match percentage.

As for the changes, some of them are fixes of actual mistakes. Regarding the rest, I am not sure if it would be a better approach to try to fix them in `reccmp`, especially regarding reused strings.

### DIFF
--- a/LEGO1/library_msvc.h
+++ b/LEGO1/library_msvc.h
@@ -281,9 +281,6 @@
 // LIBRARY: LEGO1 0x1008fdd0
 // __dosmaperr
 
-// LIBRARY: LEGO1 0x1008fe30
-// __unlock_file
-
 // LIBRARY: LEGO1 0x1008fe50
 // __errno
 
--- a/LEGO1/lego/legoomni/include/buildings.h
+++ b/LEGO1/lego/legoomni/include/buildings.h
@@ -11,7 +11,7 @@ class RaceStandsEntity : public BuildingEntity {
 	// FUNCTION: LEGO1 0x1000efa0
 	const char* ClassName() const override // vtable+0x0c
 	{
-		// STRING: LEGO1 0x100f0300
+		// at LEGO1 0x100f0300, needs no annotation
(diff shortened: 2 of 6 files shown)


LEGO1/library_msvc.h
```diff
@@ -281,9 +281,6 @@
 // LIBRARY: LEGO1 0x1008fdd0
 // __dosmaperr
 
-// LIBRARY: LEGO1 0x1008fe30
-// __unlock_file
-
```
jonschz, line 286: this one already is in the file at the correct address

foxtacles:
Looks good! I think it would be nice if `reccmp` could handle duplicate string annotations. Otherwise it would be kind of random where we place the `STRING` annotation if there are multiple candidates - if we can leave them on multiple globals that would seem more consistent.

PR #1180, opened 2024-11-30
Implement functions in Act3Actor

MishaProductions:
Code is very similar to LegoRaceActor, so I had to modify it a little bit for it to match. Act3Actor::VTable0x90 does not match 100% due to something being weird with the ```+ 2000.0f``` addition


foxtacles (approved):
Looks good! Just adjusted a few minor things for consistency. Remaining diff in `Act3Actor::VTable0x90` is most likely just compiler entropy

PR #1182, opened 2024-11-30
Implement `Act2Actor::Act2Actor()`, add BETA10 references

jonschz:
The static array and its data structure will be refined once we know more about it.

### DIFF
--- a/LEGO1/lego/legoomni/src/actors/act2actor.cpp
+++ b/LEGO1/lego/legoomni/src/actors/act2actor.cpp
@@ -6,8 +6,14 @@
 #include "roi/legoroi.h"
 
 DECOMP_SIZE_ASSERT(Act2Actor, 0x1a8)
+DECOMP_SIZE_ASSERT(Act2Actor::UnknownListStructure, 0x20)
 
-// STUB: LEGO1 0x100187e0
+// TODO: Copy the data once we know more about its fields. Total: 10 entries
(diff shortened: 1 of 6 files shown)


LEGO1/lego/legoomni/src/actors/act2actor.cpp
```diff
@@ -6,8 +6,14 @@
 #include "roi/legoroi.h"
 
 DECOMP_SIZE_ASSERT(Act2Actor, 0x1a8)
+DECOMP_SIZE_ASSERT(Act2Actor::UnknownListStructure, 0x20)
 
-// STUB: LEGO1 0x100187e0
+// TODO: Copy the data once we know more about its fields. Total: 10 entries
+// GLOBAL: LEGO1 0x100f0db8
+static Act2Actor::UnknownListStructure g_unk0x100f0db8[] = {{{0}, 0, {0}}};
```
foxtacles, line 13: We should make this a static class member instead of a static global, otherwise `datacmp` cannot find it for comparison

jonschz: Well, now datacmp (rightfully) fails. I have worked out the details by now, I'll also add them to this PR (but probably tomorrow).

foxtacles: Yeah, in that case we just have to comment out the annotation until we actually want to compare it

PR #1183, opened 2024-11-30
Implement more functions in `Act2Actor`

jonschz:
I will keep posting my progress here, and rebase once #1182 is merged.

Currently, `Act2Actor::VTable0xa0()` matches to 100 %.


jonschz:
Matched `Act2Actor::VTable0xa0()` down to compiler entropy (it was at 96 % but jumped back down when I fixed a minor error). It is also matched against BETA10.

The data array is matched as well, but I can't make much sense of the floats. I suspect that at least some of them are defined as a ratio that isn't obvious to me. I tried to minimize the number of necessary significant digits.

LEGO1/lego/legoomni/include/act2actor.h
```diff
@@ -2,6 +2,7 @@
 #define ACT2ACTOR_H
 
 #include "legoanimactor.h"
+#include "legopathedgecontainer.h"
```
foxtacles, line 5: I don't think we need this include in the header? Let's move it to the compilation unit to avoid an extra dependency

LEGO1/lego/legoomni/src/actors/act2actor.cpp
```diff
@@ -82,19 +95,165 @@ void Act2Actor::SetWorldSpeed(MxFloat p_worldSpeed)
...
+	Mx3DPointFloat local4c(0.0, 0.0, 0.0);
+
+	if (m_grec) {
+		delete m_grec;
+	}
+
+	m_grec = new LegoPathEdgeContainer();
+	assert(m_grec);
+
+	local38 = Vector3(g_unk0x100f0db8[p_param].m_unk0x00);
```
foxtacles, line 112: We should remove the explicit call to `Vector3` here - the compiler will do it implicitly

LEGO1/lego/legoomni/src/actors/act2actor.cpp
```diff
@@ -82,19 +95,165 @@ void Act2Actor::SetWorldSpeed(MxFloat p_worldSpeed)
...
+	local4c = Vector3(g_unk0x100f0db8[p_param].m_unk0x0c);
+
+	LegoPathBoundary* otherBoundary = m_controller->GetPathBoundary(g_unk0x100f0db8[p_param].m_unk0x18);
+
+	MxResult sts = m_controller->FUN_10048310(
+		m_grec,
+		Vector3(m_roi->GetWorldPosition()),
```
foxtacles, line 119: We should remove the explicit call to `Vector3` here - the compiler will do it implicitly

LEGO1/lego/legoomni/include/act2actor.h
jonschz: I must have added it to the wrong file by accident, thanks

PR #1191, opened 2024-12-04
Rename c_notificationDrag* enum

disinvite:
The constructor for `MxNotification` is at `0x101253c6` in the beta, and it shows a debug string set at the `+0x08` offset. This conveniently provides the listing of the enum for notification type. There are other differences, but I noticed the notifications for dragging an object were out of order, so I changed those to match. `LegoCameraController::Notify` reads better now:
- if `c_notificationDragEnd` call `OnLButtonUp/OnRButtonUp`
- if `c_notificationDragStart` call `OnLButtonDown/OnRButtonDown`
- if `c_notificationDrag` call `OnMouseMove`

Accuracy boost due to slight tweaks in `LegoInputManager::FUN_1005cdf0`, but no diff other than that.

### DIFF
--- a/LEGO1/lego/legoomni/include/legoeventnotificationparam.h
+++ b/LEGO1/lego/legoomni/include/legoeventnotificationparam.h
@@ -54,9 +54,15 @@ public:
 	MxS32 GetY() const { return m_y; }
 
 	void SetROI(LegoROI* p_roi) { m_roi = p_roi; }
+
+	// FUNCTION: BETA10 0x1007d620
 	void SetModifier(MxU8 p_modifier) { m_modifier = p_modifier; }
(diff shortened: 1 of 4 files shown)


LEGO1/lego/legoomni/include/legoeventnotificationparam.h
```diff
@@ -54,9 +54,15 @@ class LegoEventNotificationParam : public MxNotificationParam {
 	MxS32 GetY() const { return m_y; }
 
 	void SetROI(LegoROI* p_roi) { m_roi = p_roi; }
+
+	// FUNCTION: BETA10 0x1007d620
```
foxtacles, line 58: can you put a blank line after the function declaration? thanks,

PR #1192, opened 2024-12-05
Implement `LegoPathController::FUN_10048310`

foxtacles:
The match is only at ~55% since some of the functions aren't inlined correctly. I've matched it as best as possible to the BETA version (except for adding the BETA annotations) and it's very close to matching, but they probably refactored some stuff in the final version. Functionally it should be accurate.

I've also raised the max functions of `LEGO1.DLL` in the CI build by 50.

@disinvite @jonschz there is one oddity regarding this symbol:

`_Tree<LegoPathCtrlEdge *,LegoPathCtrlEdge *,set<LegoPathCtrlEdge *,LegoPathCtrlEdgeCompare,allocator<LegoPathCtrlEdge *> >::_Kfn,LegoPathCtrlEdgeCompare,allocator<LegoPathCtrlEdge *> >::erase`

apparently this symbol exists twice in the PDB and refers to two different functions. Unfortunately for the purposes of comparison/matching, `reccmp` is picking the wrong one. Is there any way we can disambiguate the two? If `reccmp` could issue a warning if there's multiple candidates that would help too.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -139,7 +139,7 @@ jobs:
       run: |
         reccmp-reccmp -S CONFIGPROGRESS.SVG --svg-icon assets/config.png --target CONFIG | tee CONFIGPROGRESS.TXT
         reccmp-reccmp -S ISLEPROGRESS.SVG --svg-icon assets/isle.png --target ISLE | tee ISLEPROGRESS.TXT
-        reccmp-reccmp -S LEGO1PROGRESS.SVG -T 4252 --svg-icon assets/lego1.png --target LEGO1 | tee LEGO1PROGRESS.TXT
+        reccmp-reccmp -S LEGO1PROGRESS.SVG -T 4302 --svg-icon assets/lego1.png --target LEGO1 | tee LEGO1PROGRESS.TXT
 
     - name: Compare Accuracy With Current Master
       shell: bash
--- a/LEGO1/lego/legoomni/include/act2actor.h
+++ b/LEGO1/lego/legoomni/include/act2actor.h
@@ -11,10 +11,10 @@
 class Act2Actor : public LegoAnimActor {
 public:
 	struct UnknownListStructure {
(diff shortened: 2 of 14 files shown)


disinvite:
The quick fix is to change the annotation to:
```cpp
// TEMPLATE: LEGO1 0x10049160
// ?erase@?$_Tree@PAULegoPathCtrlEdge@@PAU1@U_Kfn@?$set@PAULegoPathCtrlEdge@@ULegoPathCtrlEdgeCompare@@V?$allocator@PAULegoPathCtrlEdge@@@@@@ULegoPathCtrlEdgeCompare@@V?$allocator@PAULegoPathCtrlEdge@@@@@@QAEIABQAULegoPathCtrlEdge@@@Z
```

The two symbols are:
```
S_PUB32: [0001:00051090], Flags: 00000000, ?erase@?$_Tree@PAULegoPathCtrlEdge@@PAU1@U_Kfn@?$set@PAULegoPathCtrlEdge@@ULegoPathCtrlEdgeCompare@@V?$allocator@PAULegoPathCtrlEdge@@@@@@ULegoPathCtrlEdgeCompare@@V?$allocator@PAULegoPathCtrlEdge@@@@@@QAE?AViterator@1@V21@@Z
S_PUB32: [0001:00054920], Flags: 00000000, ?erase@?$_Tree@PAULegoPathCtrlEdge@@PAU1@U_Kfn@?$set@PAULegoPathCtrlEdge@@ULegoPathCtrlEdgeCompare@@V?$allocator@PAULegoPathCtrlEdge@@@@@@ULegoPathCtrlEdgeCompare@@V?$allocator@PAULegoPathCtrlEdge@@@@@@QAEIABQAULegoPathCtrlEdge@@@Z
```

Both have the same function name because they differ in arguments and return type. We pick whichever one is first when matching. Most of the time (by chance I guess) we choose correctly, but not in this case.

PR #1193, opened 2024-12-06
(Proposal) Introduce arithmetic operators to `Vector2`

foxtacles:
This PR attempts to resolve the following awkward syntax we've been seeing across the code base:

`((Vector3&) positionRef).Sub(g_unk0x10104ef0);`

In all of those instances, what we are finding in the original assembly is a dynamic dispatch to the `Add`/`Sub`/`Mul`/`Div` virtual functions. However, whenever the type is known to the compiler (usually when we have a `Mx3DPointFloat` on the stack), the compiler optimizes the call and inlines the virtual function (almost all the time). What we've done so far is take a base class reference which strips away the type information, which forced dynamic dispatch. However:

* It is very unlikely the developers wrote code like this
* In *all* cases in the code base, `Add`/`Sub`/`Mul`/`Div` are never actually inlined in the original assembly
* This strange behavior is only limited to the `Add`/`Sub`/`Mul`/`Div` functions. Other virtual inline functions implemented in `vector.h` *are* being inlined

An intuition to resolve this might be to implement those 4 functions in a compilation unit so that they cannot be inlined. However, in this case the compiler will generate a static call to those functions and not use dynamic dispatch if the type is fully known. This is not what we are seeing in the original assembly.

The only reasonable explanation must be an additional indirection: if those functions are called from *within* the scope of a function implemented in the base class, the compiler will in all cases use dynamic dispatch. This looks like a limitation in terms of what MSVC 4.20 is able to optimize.

The most likely scenario is that the developers defined `operator+=`, `operator-=`, `operator*=` and `operator/=` which forward to `Add`/`Sub`/`Mul`/`Div`. This introduces the consistent dynamic dispatch - which was most likely unintended, since it's quite inefficient.

This PR introduces the operators and makes the original functions private, and refactors the existing code to use the operators. As far as I can tell this actually improves matches in many of the functions that use them.

Note: they must have added those operators after BETA 1.0 since they are not present there. Unfortunately this means we have no confirmation that they did indeed use the operators, but considering what we can see it seems very likely to me.

Note 2: I've deliberatedly decided for the operators not to return a reference to `*this` (for now). We could however experiment with this and see if it improves matches, or if chaining the operators is something they might have done in some places.

### DIFF
--- a/LEGO1/lego/legoomni/src/actors/act3actors.cpp
+++ b/LEGO1/lego/legoomni/src/actors/act3actors.cpp
@@ -52,7 +52,7 @@ MxU32 Act3Actor::VTable0x90(float p_float, Matrix4& p_transform)
 			m_state = 0;
 			m_unk0x1c = 0;
 
-			((Vector3&) positionRef).Sub(g_unk0x10104ef0);
+			positionRef -= g_unk0x10104ef0;
 			m_roi->FUN_100a58f0(p_transform);
 			m_roi->VTable0x14();
 			return TRUE;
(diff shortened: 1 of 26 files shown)


foxtacles:
I'm going ahead with merging this since in combination with https://github.com/isledecomp/isle/pull/1195, I've found that this fixes the inlining of `LegoPathController::FUN_10048310` which futher confirms that they must have used these operators.

PR #1194, opened 2024-12-07
fix uppercase for Copyright name in CONFIG.rc

BJNFNE:


foxtacles:
The original `CONFIG.EXE` has it lowercased:

image

PR #1201, opened 2024-12-09
Implement `Act2Actor::VTable0x70`

jonschz:
I didn't quite get this one perfect, unfortunately. Both LEGO1 and BETA10 are matched structurally, but the stack is not correct on LEGO1 and other small mistakes might hide somewhere. BETA10 is quite different - we can remove the `#ifdef NDEBUG` when the match on LEGO1 is good enough.

### DIFF
--- a/LEGO1/lego/legoomni/src/actors/act2actor.cpp
+++ b/LEGO1/lego/legoomni/src/actors/act2actor.cpp
@@ -1,6 +1,7 @@
 #include "act2actor.h"
 
 #include "3dmanager/lego3dmanager.h"
+#include "actions/act2main_actions.h"
 #include "legoact2.h"
 #include "legocachesoundmanager.h"
 #include "legopathcontroller.h"
@@ -32,6 +33,21 @@ Act2Actor::UnknownListStructure g_unk0x100f0db8[] = {
 	{{-36.689, -0.978409, 31.449}, {0.083792, -0.94303, -0.66398698}, "edg00_157", FALSE},
 	{{-44.6, 0.1, 45.3}, {0.95, 0.0, -0.3}, "edg00_154", FALSE},
 };
+
+// GLOBAL: LEGO1 0x100f0f1c
+MxFloat g_unk0x100f0f1c = 0.0f;
+
+// GLOBAL: LEGO1 0x10102b1c
+// GLOBAL: BETA10 0x10209f60
+undefined4 g_unk0x10102b1c = 0;
+
+// GLOBAL: LEGO1 0x10102b20
+// GLOBAL: BETA10 0x10209f64
+undefined4 g_unk0x10102b20 = 0;
+
+// GLOBAL: LEGO1 0x10102b24
+// GLOBAL: BETA10 0x10209f68
+undefined4 g_unk0x10102b24 = 0;
 
 // FUNCTION: LEGO1 0x100187e0
 // FUNCTION: BETA10 0x1000c7fb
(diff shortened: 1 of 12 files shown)


LEGO1/lego/legoomni/src/actors/act2actor.cpp
```diff
@@ -115,11 +122,200 @@ MxResult Act2Actor::VTable0x9c()
 	}
 }
 
-// STUB: LEGO1 0x10018c30
-// STUB: BETA10 0x1000cb52
+// GLOBAL: LEGO1 0x100f0f1c
+MxFloat g_unk0x100f0f1c = 0.0f;
```
jonschz, line 126: I'll move this one up in the next commit

LEGO1/lego/legoomni/src/paths/legoanimactor.cpp
```diff
@@ -30,7 +30,7 @@ LegoAnimActorStruct::~LegoAnimActorStruct()
 }
 
 // FUNCTION: LEGO1 0x1001c130
-// FUNCTION: BETA10 0x1003df5f
+// FUNCTION: BETA10 0x1003df3a
```
jonschz, line 33: This was incorrect - the function didn't begin at that point

LEGO1/lego/legoomni/src/actors/act2actor.cpp
```diff
@@ -1,11 +1,19 @@
 #include "act2actor.h"
 
+#include "3dmanager/lego3dmanager.h"
+#include "actions/act2main_actions.h"
```
foxtacles, line 4: We don't need the `actions/` prefix for those headers

LEGO1/lego/legoomni/src/actors/act2actor.cpp
```diff
@@ -115,11 +138,197 @@ MxResult Act2Actor::VTable0x9c()
...
+								childPosition = childROI->GetWorldPosition();
+								distance3 = DISTSQRD3(pepperWorldPosition, childPosition);
+
+								if (distance3 > distance2) {
+									FUN_100199f0(0);
+								}
+								else
+#ifdef NDEBUG
+									// `distance2` is guessed, looks promising
+									if (distance2 - m_unk0x24 > 3000.0f) {
```
foxtacles, line 297: I'm fairly sure this should be `p_time` instead of `distance2`

LEGO1/lego/legoomni/src/actors/act2actor.cpp
```diff
@@ -115,11 +138,197 @@ MxResult Act2Actor::VTable0x9c()
...
-// STUB: LEGO1 0x10018c30
-// STUB: BETA10 0x1000cb52
+// FUNCTION: LEGO1 0x10018c30
+// FUNCTION: BETA10 0x1000cb52
 void Act2Actor::VTable0x70(float p_time)
 {
-	// TODO
+	int dummy1; // for BETA10, not sure what it is being used for
+	ViewManager* vm;
+	LegoROI* roiPepper;
```
foxtacles, line 147: Moving the declarations of all those variables to where they are used seems to improve the match to about ~78% for me - it still retains the stack allocation issue (and the stack size mismatch) but the diff becomes much more readable and it seems reasonable that at least some of those variables were not declared at the beginning (I'd say at least `childPosition` to `distance3`, but maybe also the ROIs)

LEGO1/lego/legoomni/src/actors/act2actor.cpp
jonschz: looks more promising!

LEGO1/lego/legoomni/src/actors/act2actor.cpp
jonschz: Interesting - the exact reverse happened when I had `distance3` instead of `p_time` in the other place. I changed it back.

PR #1206, opened 2024-12-11
Act3List Insert() and Clear()

disinvite:
Both 100%. Second arg to `Insert` is probably an enum of some kind.


LEGO1/lego/legoomni/include/act3.h
```diff
@@ -21,6 +21,11 @@ struct Act3ListElement {
 	undefined4 m_unk0x04; // 0x04
 	undefined m_unk0x08;  // 0x08
 
+	Act3ListElement(){};
```
foxtacles, line 24: shouldn't have a trailing ;

PR #1209, opened 2024-12-11
Act3Cop::VTable0x94 and related

disinvite:
All 100%

### DIFF
--- a/LEGO1/lego/legoomni/include/act3.h
+++ b/LEGO1/lego/legoomni/include/act3.h
@@ -5,6 +5,7 @@
 #include "legogamestate.h"
 #include "legostate.h"
 #include "legoworld.h"
+#include "mxgeometry/mxmatrix.h"
 
 class Act3Brickster;
 class Act3Cop;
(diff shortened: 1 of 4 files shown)


LEGO1/lego/legoomni/include/act3.h
```diff
@@ -5,6 +5,7 @@
 #include "legogamestate.h"
 #include "legostate.h"
 #include "legoworld.h"
+#include "mxgeometry/mxmatrix.h"
```
foxtacles, line 8: I don't think we need this include? At most a forward declaration of `class Matrix4` should do I think

LEGO1/lego/legoomni/include/act3cop.h
```diff
@@ -19,12 +20,15 @@ class Act3Cop : public Act3Actor {
...
 
+	void FUN_10040360();
+
 	// SYNTHETIC: LEGO1 0x10043120
 	// Act3Cop::`scalar deleting destructor'
 
 private:
 	MxFloat m_unk0x20;       // 0x20
-	undefined4 m_unk0x24[3]; // 0x24
+	Act3* m_world;           // 0x24
```
foxtacles, line 30: I think this member is called `a3`. In other classes, I've given the `Act3*` reference the name `m_a3` but we can also add an exception to the `ncc` naming and make it consistently `a3`.

Source: BETA assert at `1001b069`.

The assert also indicates that the `Act3Actor` based classes and `Act3Actor` itself (`Act3Actor`, `Act3Cop`, `Act3Brickster`) - in that order, and maybe also `Act3Shark` at the end even though it's not `Act3Actor` based - were located in a single compilation unit `act3actors.cpp` (which makes sense I guess considering the naming). We could refactor this here or in a separate PR.

disinvite: Thanks. Still looks fine in the beta, but the function differs in other areas anyway.

We should condense those files in a separate PR.

PR #1223, opened 2024-12-14
Implement/match `LegoPlantManager::FUN_10026410`

jonschz:
Matches down to compiler entropy (and one small issue in `reccmp` which we haven't addressed yet). I also found small BETA10 improvements for `LegoBuildingManager::FUN_10030630`.

I did notice an oddity with `LegoWEGEdge::GetUnknown0x14()` in BETA10, which appears to be defined twice. Not sure what is going on here, opinions are appreciated.

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legobuildingmanager.cpp
+++ b/LEGO1/lego/legoomni/src/common/legobuildingmanager.cpp
@@ -221,6 +221,7 @@ char* LegoBuildingManager::g_customizeAnimFile = NULL;
 MxS32 g_buildingManagerConfig = 1;
 
 // GLOBAL: LEGO1 0x10104c30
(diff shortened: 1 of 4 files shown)


foxtacles:
> I did notice an oddity with LegoWEGEdge::GetUnknown0x14() in BETA10, which appears to be defined twice. Not sure what is going on here, opinions are appreciated.

It could be that they just re-defined the same function on `LegoPathBoundary`. Unfortunately this is one of those things that is very hard to know: when did they use a base class pointer + casts, when did they use `LegoPathBoundary` directly etc. - maybe we can figure it out if we look at all the usage sites and try the different combinations and see if anything matches better, but it's similarly challenging like the vector/matrix classes.

PR #1229, opened 2024-12-15
Add `g_copDest`

jonschz:
Getting this ready to merge early in order to unblock `Act3Cop::FUN_10040360`.


LEGO1/lego/legoomni/src/worlds/act3.cpp
```diff
@@ -728,6 +728,18 @@ MxResult Act3::FUN_10073390(Act3Ammo& p_ammo, const Vector3& p_param2)
 	return SUCCESS;
 }
 
+// FUNCTION: LEGO1 0x100733d0
+// FUNCTION: BETA10 0x10016b5d
+void Act3::AddCop(Act3Cop* p_cop)
```
foxtacles, line 733: I think I'd call it SetCop to be consistent with brickster

PR #1230, opened 2024-12-15
Implement/match `Act3Cop::ParseAction`

jonschz:
Down to entropy on LEGO1 and BETA10. The assertions are quite helpful here.

### DIFF
--- a/LEGO1/lego/legoomni/src/actors/act3actors.cpp
+++ b/LEGO1/lego/legoomni/src/actors/act3actors.cpp
@@ -272,7 +272,6 @@ void Act3Cop::ParseAction(char* p_extra)
 	}
 
 	assert(m_eatAnim);
-	return;
 }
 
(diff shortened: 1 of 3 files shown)


LEGO1/lego/legoomni/src/actors/act3actors.cpp
```diff
@@ -210,11 +211,68 @@ MxResult Act3Cop::HitActor(LegoPathActor* p_actor, MxBool p_bool)
...
+	}
+
+	for (i = 0; i < m_animMaps.size(); i++) {
+		if (m_animMaps[i]->GetUnknown0x00() == -1.0f) {
+			m_eatAnim = m_animMaps[i];
+		}
+	}
+
+	assert(m_eatAnim);
+	return;
```
foxtacles, line 275: we can probably remove this return statement

PR #1242, opened 2024-12-19
Progress on `Helicopter::FUN_100042a0`

jonschz:
Here is my progress in case someone wants to continue. I am a bit stuck.

On the bright side, there may be a non-inlined version of `UnknownMx4DPointFloat::Unknown7()` here, where I improved the match. I didn't check the impact on other parts of the code yet.

### DIFF
--- a/LEGO1/lego/legoomni/include/act3.h
+++ b/LEGO1/lego/legoomni/include/act3.h
@@ -130,6 +130,7 @@ public:
 	void FUN_10073400();
 	void FUN_10073430();
 	void GoodEnding(const Matrix4& p_destination);
+	void FUN_10073a60();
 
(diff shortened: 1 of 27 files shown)


foxtacles:
I gave it a shot and improved the match to about ~60%. Still not 100% sure if the logic correct but I think it's closer. Unfortunately the compiler optimizes the vectors a lot and Ghidra isn't particularly helpful with this function.

We can leave this open for a while in case someone else wants to give it a try.

foxtacles:
Actually I think we got somewhere - replaced the `ZEROVEC` macro with a call to `Clear` and now we have a ~95% match.

jonschz:
Good job!

PR #1251, opened 2024-12-21
Improve `Helicopter::Animate()`

jonschz:
This improves the match and appears to be more correct to me, though I didn't get it above 73 %. The zoom on the brickster is still wrong (I didn't test it before so I don't know if the behaviour has changed).

### DIFF
--- a/LEGO1/mxgeometry/mxgeometry3d.h
+++ b/LEGO1/mxgeometry/mxgeometry3d.h
@@ -196,17 +196,17 @@ inline long UnknownMx4DPointFloat::FUN_10004520()
 // FUNCTION: BETA10 0x1004ab10
 inline int UnknownMx4DPointFloat::FUN_100040a0(Vector4& p_v, float p_f)
 {
-	if (m_unk0x30 == 1) {
+	if (m_unk0x30 == c_bit1) {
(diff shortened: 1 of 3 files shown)


foxtacles:
It works now:

https://github.com/user-attachments/assets/5560c3e5-e293-488b-afb4-2cf99a94f77a

LEGO1/mxgeometry/mxgeometry3d.h
```diff
@@ -170,7 +170,7 @@ int UnknownMx4DPointFloat::BETA_1004aaa0(Matrix4& p_matrix, float p_f)
 }
 
 // FUNCTION: LEGO1 0x10004520
-inline long UnknownMx4DPointFloat::FUN_10004520()
+long UnknownMx4DPointFloat::FUN_10004520()
```
foxtacles: This shouldn't make a difference functionally - the function will be treated as to be potentially inlined no matter what. For consistency we have been adding that keyword for header-defined functions where the body is outside the class definition (due to its size)

jonschz: yeah, you're right - I made a mistake when comparing and it looked like it made a change even when it didn't

PR #1256, opened 2024-12-22
readme refresh 2

Ramen2X:
This PR brings the project's README more in line with the current status.

The changes I've made are mostly in regard to the reality that we unfortunately cannot guarantee byte-for-byte matching binaries at this point in time. These parts of the README were written very early on in the project, when we didn't fully understand the concept of compiler entropy and were still under the assumption that having complete `LEGO1` headers would guarantee identical binary output. Because of the uncertainness and complexity of the situation with the compiler, I have put a bigger focus on functional accuracy instead.

I have also made changes to the `LEGO1.DLL` status description now that the game is fully playable when built from our source code. As far as we know, there should be no more inaccuracies regarding save data either, so I have taken that part out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,9 @@ This is a **work-in-progress** decompilation of LEGO Island (Version 1.1, Englis
 
 <img src="https://legoisland.org/progress/ISLEPROGRESS.SVG" width="50%"><img src="https://legoisland.org/progress/LEGO1PROGRESS.SVG" width="50%">
 
-Currently, `ISLE.EXE` is completely decompiled and is functionally identical to the original. Due to various complexities with regard to the compiler, it is not a byte-for-byte match of the original executable. We remain hopeful that this can be resolved at some point.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,16 +8,16 @@ This is a **work-in-progress** decompilation of LEGO Island (Version 1.1, Englis
 
 <img src="https://legoisland.org/progress/ISLEPROGRESS.SVG" width="50%"><img src="https://legoisland.org/progress/LEGO1PROGRESS.SVG" width="50%">
 
-Currently, `ISLE.EXE` is completely decompiled and behaves identically to the original. A handful of stubborn instructions are not yet matching; however, we anticipate they will as more of the overall codebase is implemented.
+Currently, `ISLE.EXE` is completely decompiled and is functionally identical to the original. Due to various complexities with regard to the compiler, it is not a byte-for-byte match of the original executable. We remain hopeful that this can be resolved at some point.
```
foxtacles, line 11: I think I'd merge these two paragraphs. Everything that is said applies equally to both modules imo

Ramen2X: Good idea!

Ramen2X: I merged the ISLE and LEGO1 statues into one paragraph. I kept the second paragraph but kept it limited to the note about accuracy only, just because I feel that this breaks up the reading flow better.

foxtacles (approved):

PR #1257, opened 2024-12-22
`MxDSBuffer::ParseChunk` to 100%

disinvite:
Not a full beta match because the beta does not call `MxDSStreamingAction::FUN_100cd2d0`. It does almost the same thing but like this:

```cpp
if (m_unk0x30->GetDuration() != -1) {
    m_unk0x30->SetLoopCount(m_unk0x30->GetLoopCount() - 1);
    m_unk0x30->SetUnknowna8toDuration()
}
```

Filled in some other beta addrs and fixed member set order in `MxDSBuffer` and `MxDSChunk` constructors.

### DIFF
--- a/LEGO1/omni/src/stream/mxdsbuffer.cpp
+++ b/LEGO1/omni/src/stream/mxdsbuffer.cpp
@@ -258,7 +258,8 @@ MxResult MxDSBuffer::ParseChunk(
 		MxU32 length = p_header->GetLength() + MxDSChunk::GetHeaderSize() + 8;
 		MxDSBuffer* buffer = new MxDSBuffer();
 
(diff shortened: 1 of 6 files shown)


LEGO1/omni/src/stream/mxdsbuffer.cpp
```diff
@@ -254,26 +258,18 @@ MxResult MxDSBuffer::ParseChunk(
...
-				*flags = p_header->GetChunkFlags() & ~DS_CHUNK_SPLIT;
-
-				delete p_header;
-				(*p_streamingAction)->SetUnknowna0(buffer);
-				goto done;
-			}
-		}
-
-		if (buffer) {
+		if (!buffer || buffer->AllocateBuffer(length, e_allocate) || buffer->CalcBytesRemaining((MxU8*) p_data) ||
```
foxtacles, line 261: I think we should test for != SUCCESS on both function calls here so it's clearer

PR #1261, opened 2024-12-23
Implement `MxDisplaySurface::VTable0x24`

disinvite:
As noted in the comments: almost a direct copy of `MxDisplaySurface::VTable0x28`.


disinvite:
One sec. Checking something in the beta, where we do have `VTable0x28`

foxtacles:
Nice! Two go go 😄

PR #1262, opened 2024-12-23
Implement MxDisplaySurface::VTable0x2c

ItzSwirlz:
This is basically the same as VTable0x30 but giving the LPDDSURFACEDESC via parameter instead of creating it in the function.

92.09% match, I fiddled with this and that's the most I got. It's weird - the default case comes *before* the regular cases?


foxtacles:
@madebr I've merged changes from your branch here. Compiler entropy changed things a bit so we have a match of ~83% now but structurally it looks very much correct.

PR #1264, opened 2024-12-23
Implement MxDisplaySurface::VTable0x34

madebr:
Accuracy is an embarrassing 34.91%.

Here's some stack information to get you started:
```
Stack depth is 0x84, DDSURFACEDESC surfaceDesc is at ESP+0x18
The arguments are at:
p_pixels: ESP+0x88
p_bpp: ESP+0x8c
p_width: ESP+0x90
p_height: ESP+0x94
p_x: ESP+0x98
p_y: ESP+0x9c
``` 

(It's a "RawBlitAt" function)

### DIFF
--- a/LEGO1/omni/src/video/mxdisplaysurface.cpp
+++ b/LEGO1/omni/src/video/mxdisplaysurface.cpp
@@ -1,6 +1,7 @@
 #include "mxdisplaysurface.h"
 
 #include "mxbitmap.h"
+#include "mxdebug.h"
 #include "mxmisc.h"
 #include "mxomni.h"
 #include "mxpalette.h"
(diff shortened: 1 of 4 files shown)


foxtacles:
Matched it to 100%. Since the function exists in the BETA at address `0x10141191` it's much easier to use this as a reference.

LEGO1/omni/src/video/mxdisplaysurface.cpp
```diff
@@ -739,10 +740,75 @@ void MxDisplaySurface::DrawTransparentRLE(
...
+			while (p_height--) {
+				memcpy(dst, pixels, p_width);
+				pixels += stride;
+				dst += length;
+			}
+			break;
+		}
+		case 16: {
+			if (p_bpp == 16) {
+				MxU8* dst = (MxU8*) surfaceDesc.lpSurface + p_y * surfaceDesc.lPitch + p_x;
```
madebr, line 782: Is this correct if both dest and src contain 16-bit pixels?
Shouldn't it be `+ 2 * p_x`?

foxtacles: Yes I think you are right. It looks like a bug. The function is never called in the game

Ramen2X:
I was here! Thanks everyone. 🎉

Ramen2X (approved):

PR #1267, opened 2024-12-24
Fix -Wwrite-strings warnings

madebr:
warning: ISO C++ forbids converting a string constant to 'LegoChar*'


foxtacles:
Thanks for contributing this back to the decomp and not just the portable version!

PR #1269, opened 2024-12-24
SkateBoard: name m_unk0x160 -> m_pizzaVisible

ItzSwirlz:
m_unk0x160 is just a boolean for whether the pizza is shown in the hand on the left when skateboarding.

This can easily be tested by setting the value to TRUE in the constructor.

By the way - how should we name variables? This could be m_isPizzaVisible, m_showPizza, m_showingPizza, etc

### DIFF
--- a/LEGO1/lego/legoomni/include/skateboard.h
+++ b/LEGO1/lego/legoomni/include/skateboard.h
@@ -34,7 +34,7 @@ public:
 	MxLong HandleControl(LegoControlManagerNotificationParam& p_param) override; // vtable+0xd4
 	void Exit() override;                                                        // vtable+0xe4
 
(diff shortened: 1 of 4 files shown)


foxtacles:
> By the way - how should we name variables? This could be m_isPizzaVisible, m_showPizza, m_showingPizza, etc

There is not much of a hard rule here other than:

1) If we can find a name confirmed through the BETA10, this name always takes preference 
2) Otherwise, naming should be as concise as possible and most importantly, somewhat consistent with other naming in the code. In this case, `pizzaVisible` seems reasonable to make sure it contains the reference to the pizza, and "visible" is used in several other places already so it's fine.

Other than that, a name must make sense and be verified. "Vague" names are worse than no names (unk style)

PR #1270, opened 2024-12-24
Add BETA10 macro to enable conditional compilation for BETA10 targets

foxtacles:
@madebr @jonschz could you check if that makes sense? Thank you!


madebr (approved):

jonschz:
Looks good overall! A few remarks:
* Are the `#ifdef _DEBUG` macros relevant?
* I assume you left in the `#ifndef NDEBUG` in `decomp.h` on purpose since it affects all builds? Is there a more appropriate definition to test for?
* I would like to preserve the ability of compiling BETA10 only (without compiling LEGO1 as well), since that is significantly faster. Is that currently possible? Not sure what the best way to do that is right now; I played around a bit with different cmake arguments but didn't get anywhere. Maybe we want an additional (default on) flag to enable/disable building LEGO1?

foxtacles:
> * Are the `#ifdef _DEBUG` macros relevant?

Good point. The approach that I'm taking/envisioning is that even though we don't necessarily have proof that code we find in the BETA was present under `DEBUG` conditions in the retail - we should include it if there are no good reasons to assume that it wasn't there. In particular, the code still needs to make sense and work - i.e. the logic shouldn't break. Only in cases where the logic has clearly changed should we use the `BETA10` macro.

> * I assume you left in the `#ifndef NDEBUG` in `decomp.h` on purpose since it affects all builds? Is there a more appropriate definition to test for?

Yeah, I think we should change that part. I think `ENABLE_DECOMP_ASSERTS` should not be unset there; instead, it should only be set under the right conditions in the CMake build system. Basically, "if build is targeting retail and is not debug, set `ENABLE_DECOMP_ASSERTS`". What do you think?

> * I would like to preserve the ability of compiling BETA10 only (without compiling LEGO1 as well), since that is significantly faster. Is that currently possible? Not sure what the best way to do that is right now; I played around a bit with different cmake arguments but didn't get anywhere. Maybe we want an additional (default on) flag to enable/disable building LEGO1?

Absolutely, I think that would be good to have. I don't think the current build setup supports that.

> * Edit: We should update the Readme (`tools/README.md`), section _Re-compiling a beta build_

👍 

@madebr maybe you could assist us with the CMake aspects? (see above)

foxtacles:
@madebr do you think you could help us with the CMake adjustments as described in my last comment? Thank you!

madebr:
Something like this?

<details>

<summary>Add ISLE_BUILD_LEGO1 cmake option</summary>

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 4f995f33..0e60ee50 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,12 @@ option(ISLE_USE_SMARTHEAP "Build LEGO1.DLL with SmartHeap" ${MSVC_FOR_DECOMP})
 option(ISLE_USE_DX5 "Build with internal DirectX 5 SDK" ON)
 option(ISLE_DECOMP_ASSERT "Assert struct size" ${MSVC_FOR_DECOMP})
 cmake_dependent_option(ISLE_USE_DX5_LIBS "Build with internal DirectX 5 SDK Libraries" ON ISLE_USE_DX5 OFF)
-option(ISLE_BUILD_BETA10 "Build BETA10.EXE library" OFF)
+option(ISLE_BUILD_LEGO1 "Build LEGO1.DLL library" ON)
+option(ISLE_BUILD_BETA10 "Build BETA10.DLL library" OFF)
+
+if(NOT (ISLE_BUILD_LEGO1 OR ISLE_BUILD_BETA10))
+  message(FATAL_ERROR "ISLE_BUILD_LEGO1 AND ISLE_BUILD_BETA10 cannot be both disabled")
+endif()
 
 add_cxx_warning(parentheses)
 
@@ -460,14 +465,16 @@ if (ISLE_USE_SMARTHEAP)
   list(APPEND lego1_link_libraries SmartHeap::SmartHeap)
 endif()
 
-add_lego_libraries(lego1
-  LINK_LIBRARIES ${lego1_link_libraries}
-  DLL_OUTPUT_NAME "LEGO1"
-  DLL_PREFIX ""
-  DLL_SUFFIX ".DLL"
-  OUT_TARGETS lego1_targets
-)
-reccmp_add_target(lego1 ID LEGO1)
+if(ISLE_BUILD_LEGO1)
+  add_lego_libraries(lego1
+    LINK_LIBRARIES ${lego1_link_libraries}
+    DLL_OUTPUT_NAME "LEGO1"
+    DLL_PREFIX ""
+    DLL_SUFFIX ".DLL"
+    OUT_TARGETS lego1_targets
+  )
+  reccmp_add_target(lego1 ID LEGO1)
+endif()
 
 if(ISLE_BUILD_BETA10)
   add_lego_libraries(beta10
@@ -497,7 +504,12 @@ if (ISLE_BUILD_APP)
   endif()
 
   # Link DSOUND, WINMM, and LEGO1
-  target_link_libraries(isle PRIVATE dsound winmm lego1)
+  target_link_libraries(isle PRIVATE dsound winmm)
+  if(ISLE_BUILD_LEGO1)
+    target_link_libraries(isle PRIVATE lego1)
+  else()
+    target_link_libraries(isle PRIVATE beta10)
+  endif()
 
   # Make sure filenames are ALL CAPS
   set_property(TARGET isle PROPERTY OUTPUT_NAME ISLE)
@@ -576,10 +588,12 @@ if (MSVC_FOR_DECOMP)
     set_property(TARGET isle ${lego1_targets} PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
   endif()
 
-  target_link_options(lego1 PRIVATE "/OPT:REF")
+  if(TARGET lego1)
+    target_link_options(lego1 PRIVATE "/OPT:REF")
 
-  # Equivalent to target_compile_options(... PRIVATE "/MT$<$<CONFIG:Debug>:d>")
-  set_property(TARGET lego1 ${lego1_targets} ${beta10_targets} PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+    # Equivalent to target_compile_options(... PRIVATE "/MT$<$<CONFIG:Debug>:d>")
+    set_property(TARGET lego1 ${lego1_targets} ${beta10_targets} PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+  endif()
 
   set(CMAKE_CXX_FLAGS "/W3 /GX /D \"WIN32\" /D \"_WINDOWS\"")
   set(CMAKE_CXX_FLAGS_DEBUG "/Gm /Zi /Od /D \"_DEBUG\"")
```

</details>

PR #1274, opened 2024-12-25
Update actions and a bit rewrite readme

larziwau:


README.md
```diff
@@ -1,67 +1,84 @@
...
-
+This project uses the [CMake](https://cmake.org/) build system, allowing compatibility with various compilers and development environments. For the most accurate results, **Microsoft Visual C++ 4.20** (the compiler used to build the original game) is recommended. All contributions will be evaluated based on the output from this compiler to maintain accuracy.
 
-These instructions will outline how to compile this repository using Visual C++ 4.2 into highly-accurate binaries where the majority of functions are instruction-matching with retail. If you wish, you can try using other compilers, but this is at your own risk and won't be covered in this guide.
+These instructions describe how to compile the repository using Visual C++ 4.2 to generate highly-accurate binaries that match the original retail functions. While other compilers may be used, doing so is at your own risk, and is not covered in this guide.
```
Ramen2X, line 19: In my opinion, this implies that the binaries are instruction-matching across every single function, which is a little
bit misleading since they aren't. We had already rewritten the README a couple days ago to avoid this language.

(In general, I personally don't see anything wrong with with how the README is currently written; I went through it a couple days ago, but that might just be me.)

Ramen2X:
This mostly looks good to me, thanks! I just have a few nitpicks regarding the rewriting of the README, but some of the changes (like the LEGO Island version section) look good to me. I do want someone else to look over this since I'm too preoccupied at the moment to be able to catch anything awry.

larziwau:
Thanks! Ive also used an updated version of setup ninja action, so it uses 1.12.1 now.

README.md
foxtacles: I agree - I don't see any material improvement in these changes to the README. We should keep the one we have right now imo.

larziwau:
ive also updated the contributing guide.

README.md
Ramen2X: Yeah, that's the only part where I see a material improvement to be completely honest. I don't see too much value in rewriting the language of the parts that were already clear in my opinion.

Ramen2X:
Would it be possible to scale back this PR to just the CI instead of modifying all of the docs? It gets unwieldy to review with so many modifications. Thanks

README.md
larziwau: Well I can revert that if you want

larziwau:
Okay

PR #1275, opened 2024-12-25
Update actions

larziwau:


foxtacles:
The "compare accuracy" part seems to be broken (check the logs)

foxtacles:
Or actually it might be fine. The output just looks different with this version of actions. Let me test this again.

foxtacles (approved):
LGTM

PR #1277, opened 2024-12-25
(docs)README.md: Update debug builds line

LuisYeah1234-hub:
First time opening a pull request here. Since the LEGO1.DLL has been fully decompiled, I’ve updated the README.md to reflect that debug builds _may now work_ with the decompiled LEGO1.DLL, as the previous line was written when the LEGO1.DLL was still a work in progress.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ cmake <path-to-source> -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=RelWithDebInfo
 ```
   - **Visual C++ 4.2 has issues with paths containing spaces**. If you get configure or build errors, make sure neither CMake, the repository, nor Visual C++ 4.2 is in a path that contains spaces.
   - Replace `<path-to-source>` with the source repository. This can be `..` if your build folder is inside the source repository.
-  - `RelWithDebInfo` is recommended because it will produce debug symbols useful for further decompilation work. However, you can change this to `Release` if you don't need them. `Debug` builds may now work with the decompiled version of `LEGO1.DLL`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ cmake <path-to-source> -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=RelWithDebInfo
 ```
   - **Visual C++ 4.2 has issues with paths containing spaces**. If you get configure or build errors, make sure neither CMake, the repository, nor Visual C++ 4.2 is in a path that contains spaces.
   - Replace `<path-to-source>` with the source repository. This can be `..` if your build folder is inside the source repository.
-  - `RelWithDebInfo` is recommended because it will produce debug symbols useful for further decompilation work. However, you can change this to `Release` if you don't need them. `Debug` builds are not recommended because they are unlikely to be compatible with the retail `LEGO1.DLL`, which is currently the only way to use this decompilation for gameplay.
+  - `RelWithDebInfo` is recommended because it will produce debug symbols useful for further decompilation work. However, you can change this to `Release` if you don't need them. `Debug` builds may now work with the decompiled version of `LEGO1.DLL`, However they might still have some bugs or issues.
```
Ramen2X, line 41: I would rewrite this to say that, while Debug builds can be compiled and used, they are not recommended if a developer's primary intention is matching code to the original binary, as the retail binaries were naturally compiled as Release builds.

Ramen2X (commented):
Nice catch, but I have some thoughts about how this should be rewritten.

LuisYeah1234-hub:
Is this any better now? again sorry this is my first time doing a pr here 😅

PR #1282, opened 2024-12-26
Refactor `MxRect32`, `MxSize32`, `MxPoint32` to use MFC base classes

foxtacles:


foxtacles:
Closing this for the time being since it's no longer clear they actually used the MFC classes as base. See BETA base constructor of `0x10031860` which does not exist in the MFC classes.

PR #1304, opened 2024-12-29
Fix warnings triggered by isle-portable

madebr:
I could no longer live with these warnings ;)

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legobackgroundcolor.cpp
+++ b/LEGO1/lego/legoomni/src/common/legobackgroundcolor.cpp
@@ -121,9 +121,9 @@ void LegoBackgroundColor::SetLightColor(float p_r, float p_g, float p_b)
 {
 	if (!VideoManager()->GetVideoParam().Flags().GetF2bit0()) {
 		// TODO: Computed constants based on what?
-		p_r *= 1.f / 0.23f;
-		p_g *= 1.f / 0.63f;
-		p_b *= 1.f / 0.85f;
+		p_r *= 1. / 0.23;
+		p_g *= 1. / 0.63;
+		p_b *= 1. / 0.85;
 
 		if (p_r > 1.0) {
 			p_r = 1.0;
--- a/LEGO1/lego/legoomni/src/common/legobackgroundcolor.cpp
+++ b/LEGO1/lego/legoomni/src/common/legobackgroundcolor.cpp
@@ -121,9 +121,9 @@ void LegoBackgroundColor::SetLightColor(float p_r, float p_g, float p_b)
 {
 	if (!VideoManager()->GetVideoParam().Flags().GetF2bit0()) {
 		// TODO: Computed constants based on what?
-		p_r *= 4.3478260869565215;
-		p_g *= 1.5873015873015872;
-		p_b *= 1.1764705882352942;
+		p_r *= 1.f / 0.23f;
+		p_g *= 1.f / 0.63f;
+		p_b *= 1.f / 0.85f;
 
 		if (p_r > 1.0) {
 			p_r = 1.0;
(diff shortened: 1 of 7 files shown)


LEGO1/omni/src/video/mxdisplaysurface.cpp
```diff
@@ -1114,8 +1116,8 @@ LPDIRECTDRAWSURFACE MxDisplaySurface::CreateCursorSurface()
 
 		newSurface->Unlock(ddsd.lpSurface);
 		DDCOLORKEY colorkey;
-		colorkey.dwColorSpaceHighValue = 31775;
-		colorkey.dwColorSpaceLowValue = 31775;
+		colorkey.dwColorSpaceHighValue = RGB555_CREATE(0x1f, 0, 0x1f);
```
foxtacles, line 1119: I think using 255 instead of 0x1f in these cases is easier to read/more common

LEGO1/lego/legoomni/src/common/legobackgroundcolor.cpp
```diff
@@ -121,9 +121,9 @@ void LegoBackgroundColor::SetLightColor(float p_r, float p_g, float p_b)
 {
 	if (!VideoManager()->GetVideoParam().Flags().GetF2bit0()) {
 		// TODO: Computed constants based on what?
-		p_r *= 4.3478260869565215;
-		p_g *= 1.5873015873015872;
-		p_b *= 1.1764705882352942;
+		p_r *= 1.f / 0.23f;
```
foxtacles, line 124: I can't check the assembly diff right now but there seems to be a regression - I think the constant needs to be a double instead of a float.

LEGO1/omni/src/video/mxdisplaysurface.cpp
madebr: It's `31` actually (5 bits maximum): `0b11111` (but that's not supported by older compilers)
I chose hexadecimal representation as those are also used by css colors, and also shows that the number of bits is maximum 5.

LEGO1/lego/legoomni/src/common/legobackgroundcolor.cpp
```diff
@@ -121,9 +121,9 @@ void LegoBackgroundColor::SetLightColor(float p_r, float p_g, float p_b)
...
+		p_g *= 1.f / 0.63f;
+		p_b *= 1.f / 0.85f;
```
madebr, line 126:
```suggestion
		p_r *= 1. / 0.23;
		p_g *= 1. / 0.63;
		p_b *= 1. / 0.85;
```

LEGO1/omni/src/video/mxdisplaysurface.cpp
foxtacles: Ah yes of course. My bad.

foxtacles:
There seems to be regressions in the functions that use the color value macro - it might be compiler randomness but I'll have to check

madebr:
> There seems to be regressions in terms of accuracy in the functions that use the color value macro - it might be compiler randomness but I'll have to check

If it's causing issues, then perhaps replace it with a `#define RGB565_MAGENTA 0x7c1f // RGB1555(0, 0x1f, 0, 0x1f)

foxtacles:
> > There seems to be regressions in terms of accuracy in the functions that use the color value macro - it might be compiler randomness but I'll have to check
> 
> If it's causing issues, then perhaps replace it with a `#define RGB565_MAGENTA 0x7c1f // XRGB1555(0x1f, 0, 0x1f)` macro

It looks like it doesn't compute the right value:

image

foxtacles:
It was just an issue with a logical OR instead of bitwise OR in the macro - should be good now!

madebr:
> It was just an issue with a logical OR instead of bitwise OR in the macro - should be good now!

Oh, that was a dumb mistake.
Thanks for noticing!

PR #1305, opened 2024-12-30
Some tweaks to `LegoCharacterManager`

disinvite:
- `GetActorROI` used `operator==` and a `const_iterator`. I'm thinking about ways we can identify blocks of these template functions all at once in BETA10 because many have identical structures. (In Ghidra, use Search > For Matching Instructions > Exclude Operands).
- `FindChildROI` has the iterator constructed outside of the loop.
- Beta shows that many functions here have an unnecessary final `else`.
- `FUN_100849a0`: Arg name `p_texture` confirmed by assert.

The next thing I want to try is separating the first entry of `g_actorLODs` from the rest of the list. The loop in `LegoCharacterManager::SwitchColor` suggests that this may be correct in the beta.

### DIFF
--- a/LEGO1/lego/legoomni/include/legocharactermanager.h
+++ b/LEGO1/lego/legoomni/include/legocharactermanager.h
@@ -67,7 +67,7 @@ public:
 	void ReleaseActor(const char* p_name);
 	void ReleaseActor(LegoROI* p_roi);
 	void ReleaseAutoROI(LegoROI* p_roi);
-	MxBool FUN_100849a0(LegoROI* p_roi, LegoTextureInfo* p_textureInfo);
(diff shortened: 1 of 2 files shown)


foxtacles:
I think the type `LegoTextureInfo*` has been introduced after the BETA - before that, it was `LegoTexture*` iirc. That's why I also renamed the parameter to avoid the confusion between the two. I wouldn't mind keeping the name in `p_texture` though, just wanted to mention it. Looks good!

disinvite:
Well, no problem to change it back if you think it's better the old way. It's more clear as `p_textureInfo`. I don't know if assert matching is a priority or is even possible.

PR #1306, opened 2025-01-01
Add `MXIOINFO::CreateChunk`

disinvite:
Another beta-only function that is never called. I have to assume this file was shared with other internal Mindscape tools for writing data.

I got the name for this function (and other things) from the mmio pages on Microsoft's site. https://learn.microsoft.com/en-us/windows/win32/api/mmiscapi/nf-mmiscapi-mmiocreatechunk

I also went through and added the `MMSYSERR_NOERROR` macro (replacing `0`), also referenced in those doc pages.

### DIFF
--- a/LEGO1/omni/include/mxio.h
+++ b/LEGO1/omni/include/mxio.h
@@ -29,6 +29,7 @@ public:
 	MxU16 Advance(MxU16);
 	MxU16 Descend(MMCKINFO*, const MMCKINFO*, MxU16);
 	MxU16 Ascend(MMCKINFO*, MxU16);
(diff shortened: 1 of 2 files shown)


Ramen2X:
Interesting... I wonder if this could have been a part of Weaver?

ItzSwirlz:
> Interesting... I wonder if this could have been a part of Weaver?

Might be a bit of a farfetched idea but since Weaver was used across different games, maybe it should be stored in a separate repo (and we can make sure we got Weaver matching perfectly) and then have it stored as a submodule in this repo.

PR #1310, opened 2025-01-04
Add smack functions and math globals

jonschz:
This PR increases the coverage on LEGO1 for some unused functions. Also, the end of `_SmackDoFrameToBuffer` was misdetected, now the match is better. I got some of the matches by importing `smack.lib` into Ghidra (not shared at the moment). Note that `smack.lib` does not use `INT3` to fill unreachable code segments, but an odd alternation of
```asm
xchg ebx, ebx
xchg ecx, ecx
```
which we also see in some parts of LEGO1. I am still quite suprised that the smack code does not match perfectly, but maybe we don't have the correct version? This would be out of the scope of this PR.

### DIFF
--- a/LEGO1/library_msvc.h
+++ b/LEGO1/library_msvc.h
@@ -639,6 +639,15 @@
 // GLOBAL: LEGO1 0x101028da
 // __OP_POWjmptab
 
+// GLOBAL: LEGO1 0x1010292a
+// __OP_LOG10jmptab
+
+// GLOBAL: LEGO1 0x1010294a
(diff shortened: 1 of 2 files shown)


foxtacles:
> which we also see in some parts of LEGO1. I am still quite suprised that the smack code does not match perfectly, but maybe we don't have the correct version? This would be out of the scope of this PR.

Yes, unfortunately we don't have the right version. We've spent quite some time finding a closer one but what we have is the closest we could find

PR #1313, opened 2025-01-06
Refactor surrounding `MxDSChunk` utility functions

disinvite:
Specifically: the functions `IntoType` and `IntoLength` from the `MxDSChunk` header. They seek a pointer into the right place when reading the chunk, but the beta shows they are not used. The `Size` and `End` functions are used, but they are different from what we had.

These two functions are users of the utility functions:
- `MxDiskStreamProvider::FUN_100d1b20` now 100%
- `ReadData` (from `MxRAMStreamProvider`) still 100% but with entropy

I'm not sure why, but this has also caused us to jump from 141 to 432 functions aligned. It may just be chance, but these two runs of functions now line up:

```
0001:0000b110          0  fun  103     MxEntity::~MxEntity
to
0001:00012c40          0  fun  458     LegoJetski::HitActor

and

0001:00016650          0  fun  390     CarRace::HandleClick
to
0001:00016da0          0  fun  215     PizzeriaState::Serialize
```

### DIFF
--- a/LEGO1/omni/include/mxdsbuffer.h
+++ b/LEGO1/omni/include/mxdsbuffer.h
@@ -68,7 +68,7 @@ public:
 	// FUNCTION: BETA10 0x10148c60
 	MxU8* GetBuffer() { return m_pBuffer; }
 
-	MxU8** GetBufferRef() { return &m_pBuffer; }
+	// FUNCTION: BETA10 0x10164240
 	undefined4 GetUnknown14() { return m_unk0x14; }
 
 	// FUNCTION: BETA10 0x10156420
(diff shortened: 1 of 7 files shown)


foxtacles:
> I'm not sure why, but this has also caused us to jump from 141 to 432 functions aligned. It may just be chance, but these two runs of functions now line up:
> 
> ```
> 0001:0000b110          0  fun  103     MxEntity::~MxEntity
> to
> 0001:00012c40          0  fun  458     LegoJetski::HitActor
> 
> and
> 
> 0001:00016650          0  fun  390     CarRace::HandleClick
> to
> 0001:00016da0          0  fun  215     PizzeriaState::Serialize
> ```

Probably entropy, but looks good!

PR #1314, opened 2025-01-06
Beta match `MxDSStreamingAction`

disinvite:
Removed `this->` and added beta addresses. `Init()` can be void and still use `xor eax, eax` to null out all members.

The beta shows that the `MxDSAction` version of `operator=` is used instead of one from `MxDSStreamingAction`, so I have this as `*(MxDSAction*) this = ...`. Is that syntax okay?

### DIFF
--- a/LEGO1/omni/src/action/mxdsstreamingaction.cpp
+++ b/LEGO1/omni/src/action/mxdsstreamingaction.cpp
@@ -10,7 +10,7 @@ MxDSStreamingAction::MxDSStreamingAction(MxDSAction& p_dsAction, MxU32 p_offset)
 {
 	Init();
 
-	*(MxDSAction*) this = p_dsAction;
+	MxDSAction::operator=(p_dsAction);
 	m_unk0x94 = p_offset;
 	m_bufferOffset = p_offset;
 }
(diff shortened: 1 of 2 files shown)


LEGO1/omni/src/action/mxdsstreamingaction.cpp
```diff
@@ -5,94 +5,97 @@
...
 {
-	*this = p_dsStreamingAction;
-	this->m_unk0x94 = p_dsStreamingAction.m_unk0x94;
-	this->m_bufferOffset = p_dsStreamingAction.m_bufferOffset;
-	this->m_unk0x9c = p_dsStreamingAction.m_unk0x9c;
-	this->m_unk0xa0 = NULL;
-	this->m_unk0xa4 = NULL;
-	this->m_unk0xac = p_dsStreamingAction.m_unk0xac;
-	this->m_unk0xa8 = p_dsStreamingAction.m_unk0xa8;
+	*(MxDSAction*) this = p_dsStreamingAction;
```
foxtacles, line 66: I think the `MxDSAction::operator=(p_dsStreamingAction);` would be cleaner and I think we have used that in other places too

foxtacles (approved):

PR #1315, opened 2025-01-08
Use `COMPAT_MODE` to fix inline function not defined warnings on modern compilers

foxtacles:
CC @madebr


madebr:
I compared the `build.yml/Current msys2 mingw32` action job and it's warning-free now, where it is not on current master.
Great job!

madebr (approved):

PR #1329, opened 2025-01-11
Fix regression in `LegoWorld::Enable`

foxtacles:


disinvite:
Thanks. I think that was just my goof. `size() == 0` is correct for both:

```diff
@@ BETA10 0x100dbacb @@
call Thunk of 'set<MxCore *,CoreSetCompare,allocator<MxCore *> >::size' (FUNCTION)
test eax, eax
-jne 0x3a5
+jne 0x3a2
call Thunk of 'UserActor' (FUNCTION)
```

PR #1373, opened 2025-01-24
Match `HistoryBook::ReadyWorld()`

### LINKED ISSUE #1369: Match `HistoryBook::ReadyWorld`

jonschz:
Closes #1369.

Matches to about 53 % on BETA10, though most of the mismatch comes from incorrect registers / entropy. The structural match isn't perfect, but quite decent.

I got LEGO1 to 91 %. Some of it appears to be entropy, but I can't make the beginning work:
```diff
0x1008271e : mov ecx, 'A_Bitmap' (STRING)
-0x10082723 : -cmp si, 0x1a
0x10082727 : lea edx, [esp + 0x40]
0x1008272b : mov cl, byte ptr [ecx + 8]
0x1008272e : mov dword ptr [edx], eax
0x10082730 : mov dword ptr [edx + 4], ebx
0x10082733 : mov byte ptr [edx + 8], cl
-0x10082736 : -jge 0x25
0x10082738 : lea eax, [esp + 0x40]
0x1008273c : mov ecx, dword ptr [esp + 0x18]
0x10082740 : push eax
0x10082741 : push 'MxStillPresenter' (STRING)
0x10082746 : call LegoWorld::Find(char const *, char const *) (FUNCTION)
```

It looks as if the code does something like
```c++
MxS16 i = 0;
char bitmap[] = "A_Bitmap";

for (; i < 26; i++) {
// [...]
}
```
but that only makes the match worse.

### DIFF
--- a/LEGO1/lego/legoomni/include/historybook.h
+++ b/LEGO1/lego/legoomni/include/historybook.h
@@ -41,8 +41,11 @@ public:
 private:
 	LegoGameState::Area m_destLocation; // 0xf8
 	MxStillPresenter* m_alphabet[26];   // 0xfc
-	MxStillPresenter* m_names[20][7];   // 0x164
-	MxStillPresenter* m_scores[20];     // 0x394
+
+	// variable name verified by BETA10 0x1002bd27
+	MxStillPresenter* m_name[20][7]; // 0x164
+
+	MxStillPresenter* m_scores[20]; // 0x394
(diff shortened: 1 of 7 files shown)


foxtacles:
Matched it to 100% by adding another `if (i < 26)` check in the loop. Doesn't make any sense but I guess that's what they did.

PR #1376, opened 2025-01-25
BETA10 match `RegistrationBook::ReadyWorld()`

jonschz:
As noted on Element, this function is pretty fragile and highly susceptible to small changes. It now closely resembles BETA10, but has gone down from 100 % to 100 % effective relative to `master` - I hope this is due to entropy.

I was also able to track down a few more BETA10 functions and verify the existence of the inline function `RegistrationBook::PlayAction()` (which is actually `static` - maybe it belongs somewhere else? Its BETA10 version is only called  from `RegistrationBook`, though).

One detail I couldn't work:
```diff
0x100f3a90 : push 0x1f4
-0x100f3a95 : -call RegistrationBook::PlayAction (FUNCTION)
+           : +call Thunk of 'RegistrationBook::PlayAction' (FUNCTION)
0x100f3a9a : add esp, 4
```

Not sure how we can declare a non-thunked function in BETA10 that is inlined in LEGO1.

### DIFF
--- a/LEGO1/lego/legoomni/include/infocenter.h
+++ b/LEGO1/lego/legoomni/include/infocenter.h
@@ -50,6 +50,9 @@ public:
 
 	void SetUnknown0x74(MxU32 p_unk0x74) { m_unk0x74 = p_unk0x74; }
 
+	// FUNCTION: BETA10 0x10031bd0
+	MxBool FirstLetterIsNotNull() { return m_letters[0] != NULL; }
+
 	// SYNTHETIC: LEGO1 0x10071900
 	// InfocenterState::`scalar deleting destructor'
 
--- a/LEGO1/lego/legoomni/include/infocenter.h
+++ b/LEGO1/lego/legoomni/include/infocenter.h
@@ -40,7 +40,10 @@ public:
 	MxS16 GetMaxNameLength() { return sizeOfArray(m_letters); }
 	MxStillPresenter* GetNameLetter(MxS32 p_index) { return m_letters[p_index]; }
 	void SetNameLetter(MxS32 p_index, MxStillPresenter* p_letter) { m_letters[p_index] = p_letter; }
+
+	// FUNCTION: BETA10 0x10031bd0
 	MxBool HasRegistered() { return m_letters[0] != NULL; }
+
 	Playlist& GetExitDialogueAct1() { return m_exitDialogueAct1; }
 	Playlist& GetExitDialogueAct23() { return m_exitDialogueAct23; }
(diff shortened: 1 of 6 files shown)


LEGO1/lego/legoomni/src/worlds/registrationbook.cpp
```diff
@@ -440,7 +454,15 @@ void RegistrationBook::ReadyWorld()
...
 	}
 
-	if (m_infocenterState->m_letters[0] != NULL) {
+#ifdef BETA10
+	InfocenterState* infocenterState = (InfocenterState*) GameState()->GetState("InfocenterState");
+	assert(infocenterState);
+
+	if (infocenterState->FirstLetterIsNotNull())
+#else
+	if (m_infocenterState->FirstLetterIsNotNull())
```
jonschz, line 463: Adding either the assertions above or changing this call causes the match to get worse. Adding both appears to be ok, though.

LEGO1/lego/legoomni/include/infocenter.h
```diff
@@ -50,6 +50,9 @@ class InfocenterState : public LegoState {
 
 	void SetUnknown0x74(MxU32 p_unk0x74) { m_unk0x74 = p_unk0x74; }
 
+	// FUNCTION: BETA10 0x10031bd0
```
foxtacles, line 53: I think we should call this function `HasSignedIn` since it indicates whether the player has signed in at the Infocenter

foxtacles: it exists as `HasRegistered` already (should remove that one)

LEGO1/lego/legoomni/src/worlds/registrationbook.cpp
```diff
@@ -440,7 +454,18 @@ void RegistrationBook::ReadyWorld()
 		}
 	}
 
-	if (m_infocenterState->m_letters[0] != NULL) {
+	// TODO: Remove this one if the function mismatches again
```
foxtacles, line 457: No need to add something like this. We have a lot of functions susceptible to entropy so unless this is a permanent fix that keeps it at 100% at all circumstances there's not much point to it

jonschz: I went back and validated that this is actually due to entropy. Removing the void call and adding
```
class Stuff {
	void Tmp();
};

void Stuff::Tmp() {

}
```
above the function makes it go back up to 100 % (non-effective).

foxtacles (approved):

PR #1378, opened 2025-01-27
Tweaks to `LegoRaceCar` and `LegoJetski`

disinvite:
Corrects beta address for `g_skBMap` array.

Moved `MxMatrix` variable init position in `HitActor` functions to match beta. Removed `timeElapsed` variable per the beta.

Renamed global variables related to sounds and timed events. This was partly based on checking the actual files with SIEdit. I didn't check in the actual game so feel free to correct me if I'm wrong here. I'm basing the name `g_playerHitStudsSounds` on the fact that `srt018sl` through `srt029sl` are grunts and pain sounds. In contrast, `srt006sl` is "Eat my bricks buddy!" so this seems like a sound for when Studs hits you instead of you hitting him. Does this mean that the second bool parameter of all the `HitActor` functions tells whether the player is moving at the time of the collision? (As a way to tell who hit who)

Named the `InitSoundIndices` functions for both `LegoRaceCar` and `LegoJetski` 

Removed the `animMap` variable and two string variables in `LegoRaceCar::ParseAction` per the beta. This is still 100%.

I think the accuracy dip in `LegoJetski::HitActor` is from entropy near the `strcmpi` calls.

### DIFF
--- a/LEGO1/lego/legoomni/include/legoanimactor.h
+++ b/LEGO1/lego/legoomni/include/legoanimactor.h
@@ -96,7 +96,11 @@ protected:
 // Vector<unsigned char *>::~Vector<unsigned char *>
 
 // TEMPLATE: LEGO1 0x1001c7c0
+// TEMPLATE: BETA10 0x1000fb40
 // vector<LegoAnimActorStruct *,allocator<LegoAnimActorStruct *> >::size
+
+// TEMPLATE: BETA10 0x1000fb90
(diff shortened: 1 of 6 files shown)


foxtacles:
> Does this mean that the second bool parameter of all the `HitActor` functions tells whether the player is moving at the time of the collision? (As a way to tell who hit who)

Yes I'm pretty sure. I've looked at that logic before and that seems to be what it is.

foxtacles (approved):

PR #1384, opened 2025-02-13
Add built-with-entropy pipeline

foxtacles:


jonschz:
This looks like a great idea! I'd try to seed the randomness for the CI in order to keep the CI builds deterministic.

foxtacles:
@jonschz I've added seed functionality to the script - in the CI builds, it will use the commit SHA as the base seed so given the same commit, we get the same determinstic / reproducible results.

@disinvite Do you think you can adapt the pipeline to the `denoise` script once it's ready? Right now it is only creating the build artifacts, but will still need to create the aggregation files for comparison.

tools/entropy.sh
```diff
@@ -0,0 +1,67 @@
...
+generate_name() {
+  local seed="$1"
+  local length="$2"
+  local char_count=0
+  local random_bytes=""
+
+  while [ "$char_count" -lt "$length" ]; do
+    random_bytes+=$(od -vAn -N4 -tu4 < /dev/urandom | tr -d '[:space:]')
+    char_count=$((char_count + 1))
+  done
```
disinvite, line 16: I think this does nothing

tools/entropy.sh
```diff
@@ -0,0 +1,67 @@
...
+  RANDOM="$seed"
+
+  local result=""
+  for ((i=0; i<length; i++)); do
+    # Get a pseudo-random number between 0 and 25 (inclusive)
+    local rand_index=$((RANDOM % 26))
+    # Convert to lowercase ASCII character (a=97)
+    local char_code=$((97 + rand_index))
+    # Append to the result string
+    result+=$(printf "\\$(printf '%o' "$char_code")")
```
disinvite, line 29: The `$()` evaluation seems to be the bottleneck, but I'm no bash expert.
```bash
printf -v hex "%x" $char_code
printf -v char "\x$hex"
result+="$char"
```

PR #1390, opened 2025-02-21
Fix `Matrix4::operator[]` annotation

foxtacles:
This fixes the BETA annotation of `Matrix4::operator[]`, which was previously annotated as `MxMatrix::operator[] const`. The best place to verify that this is correct is `LegoView1::Create`, since it accesses both a `Matrix4` and `MxMatrix`.

In consequence, several function parameters and members which were previously `MxMatrix` were changed to be of type `Matrix4` instead. Matches of affected functions stay the same. (but plenty of entropy in the code base as usual)

### DIFF
--- a/LEGO1/lego/legoomni/include/legoanimpresenter.h
+++ b/LEGO1/lego/legoomni/include/legoanimpresenter.h
@@ -93,7 +93,7 @@ public:
 
 	void SetCurrentWorld(LegoWorld* p_currentWorld) { m_currentWorld = p_currentWorld; }
 	void SetUnknown0x0cTo1() { m_unk0x9c = 1; }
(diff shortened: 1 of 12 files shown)


jonschz (approved):
This looks correct to me. Good catch with `Lego1View::Create`!

Have you checked all other instances where an `MxMatrix` may be referenced by pointer (e.g. `LegoAnimationManager::FUN_100609f0`)? I guess it will be impossible to tell for some.

foxtacles:
> This looks correct to me. Good catch with `Lego1View::Create`!
> 
> Have you checked all other instances where an `MxMatrix` may be referenced by pointer (e.g. `LegoAnimationManager::FUN_100609f0`)? I guess it will be impossible to tell for some.

I've only checked the instances of where the BETA operator[]'s are used to reconcile the annotations. There might be more instances of incorrect types but like you said, it might be impossible to know in some cases

PR #1398, opened 2025-03-02
Add various BETA10 references and fixes

jonschz:
A collection of BETA10 addresses that I collected while working on various things (that mostly didn't get anywhere).

### DIFF
--- a/LEGO1/library_msvc.h
+++ b/LEGO1/library_msvc.h
@@ -783,4 +783,11 @@
 // GLOBAL: BETA10 0x101faf78
 // __app_type
 
+// GLOBAL: LEGO1 0x100db6e0
+// GUID_SysKeyboard
+
+// Cannot be handled right now due to anonymous pointer in struct
+// // GLOBAL: LEGO1 0x10098f80
+// c_dfDIKeyboard
+
 #endif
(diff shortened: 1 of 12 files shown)


LEGO1/library_msvc.h
```diff
@@ -783,4 +783,11 @@
 // GLOBAL: BETA10 0x101faf78
 // __app_type
 
+// GLOBAL: LEGO1 0x100db6e0
+// GUID_SysKeyboard
+
+// Cannot be handled right now due to anonymous pointer in struct
+// // GLOBAL: LEGO1 0x10098f80
+// c_dfDIKeyboard
```
jonschz, line 791: I didn't have a good idea how to solve this one. It will fail `datacmp` because it points to a structure that contains a pointer. However, the pointee of that structure does not have a name/debug symbol.

foxtacles (approved):

LEGO1/library_msvc.h
jonschz: We would likely need an annotation like
```
// GLOBAL: BETA10 0x10097f80
// c_dfDIKeyboard.rgodf
```
but we don't have anything like at right now, as far as I am aware.

PR #1399, opened 2025-03-04
Use MxGeometry header

disinvite:
Moving all the 2D datatypes under one roof, following the example from the recently-surfaced `MxGeometry.h`.

Our hope was that the stubborn `GetRectIntersection` function could match higher than single digits with this change. Initial tests have shown that we can get to 100%, but it is rare and dependent on the whims of the compiler. That's why we need the big guns: #1384.

Summary:

- `MxPoint32`, `MxSize32`, `MxRect32`, and `MxRect16` are now housed in `mxgeometry.h` instead of their own headers. All files using these structs had their imports corrected.

- Similarly, `MxRectList` is now `MxRect32List`, also in `mxgeometry.h`.

- We had names for the functions on `MxRect32` that altered the data. These are now replaced by C++ `operator` functions. e.g. `rect_a.Intersect(rect_b)` is now `rect_a &= rect_b`.

- Modern compilers complained about the templated utility functions in `legoutil.h`. They clash with the ones in `mxutilities.h` because these are now used by `mxgeometry.h`. The three functions aren't used in `legocarbuildpresenter.cpp`, but `legoutil.h` comes from the 96 source so this may well have been correct. They are excluded with an `#ifdef` for now.

- `InfocenterMapEntry` uses the templated struct `MxRect<MxS32>` directly instead of the subclass `MxRect32`. I changed `Infocenter::InitializeBitmaps` to use these rects so it matches the beta a little better. It can never fully match because this function was changed further for retail.

- In `legometerpresenter.h`, `MeterRect` is now `MxRect16`. We knew from the beta that there were two classes here, but not why.

- `MxStillPresenter::SetPosition` had a note to try using the `MxSize32` and `MxPoint32` structures more. I've done that and it now matches 100%.


LEGO1/lego/sources/misc/legoutil.h
```diff
@@ -1,6 +1,9 @@
 #ifndef __LEGOUTIL_H
 #define __LEGOUTIL_H
 
+// TODO: exclude from modern compilers due to clash with mxutilities.h
```
jonschz, line 4: I am not quite sure what the remaining issue here is. The CI build for modern compilers is working, right? Maybe open a follow-up ticket if there is some more cleanup to be done?

jonschz (approved):
Looks very promising overall! Too bad that we still have a bunch of decreased matches despite the entropy build, but it would likely be a lot worse without it.

LEGO1/lego/sources/misc/legoutil.h
disinvite: I think this is just my bad habit of writing `// TODO` when I mean "we should maybe reconsider this later" versus "there's something to work on here."

foxtacles (approved):

PR #1412, opened 2025-03-15
Match `MxControlPresenter::HasTickleStatePassed`

foxtacles:
The only way to get this to match is to have an inline function like `advance` (with that exact interface). However, standard STL `advance` doesn't work because of the bidirectional nature of the list iterator; it will decrement the iterator if `n` is negative. But this is not part of the original assembly.

Defining a new (template) method like `advance` separately would work, but doesn't make a whole lot of sense in the context of that class with the required interface so I used a direct call to internal `_Advance` with the `forward_iterator` tag instead. Maybe they defined a function like that in the global scope like other template functions in `mxutilities.h`? Or they had a diferent version of the STL?


disinvite (approved):
Nice find. Looks good!

This is done differently in the beta, so no direct help there. I checked all the `list<MxPresenter *>` functions anyway to see if there's any chance this is a different STL structure, but they all matched up.

PR #1432, opened 2025-04-11
Beta match and cleanup of `MxDS*` classes

disinvite:
- Added beta addrs for (hopefully) all functions.
- Removed `this->`.
- All vtable offsets use our preferred format `// vtable+0x__` instead of `// vtable+__`.
- Copy constructor added for all classes (unused in and removed from retail)
- `SetType` removed from each constructor, per the beta
- Renamed `m_actions` to `m_actionList` to match beta assert

I excluded some beta matches that (for now) decreased retail.

### DIFF
--- a/LEGO1/omni/include/mxdsaction.h
+++ b/LEGO1/omni/include/mxdsaction.h
@@ -47,17 +47,17 @@ public:
 		return !strcmp(p_name, MxDSAction::ClassName()) || MxDSObject::IsA(p_name);
 	}
 
-	undefined4 VTable0x14() override;                            // vtable+0x14;
-	MxU32 GetSizeOnDisk() override;                              // vtable+0x18;
-	void Deserialize(MxU8*& p_source, MxS16 p_unk0x24) override; // vtable+0x1c;
-	virtual MxLong GetDuration();                                // vtable+0x24;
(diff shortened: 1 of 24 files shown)


LEGO1/omni/include/mxdsaction.h
```diff
@@ -47,17 +47,17 @@ class MxDSAction : public MxDSObject {
...
-	void Deserialize(MxU8*& p_source, MxS16 p_unk0x24) override; // vtable+1c;
-	virtual MxLong GetDuration();                                // vtable+24;
-	virtual void SetDuration(MxLong p_duration);                 // vtable+28;
-	virtual MxDSAction* Clone();                                 // vtable+2c;
-	virtual void MergeFrom(MxDSAction& p_dsAction);              // vtable+30;
-	virtual MxBool HasId(MxU32 p_objectId);                      // vtable+34;
-	virtual void SetUnknown90(MxLong p_unk0x90);                 // vtable+38;
-	virtual MxLong GetUnknown90();                               // vtable+3c;
-	virtual MxLong GetElapsedTime();                             // vtable+40;
+	undefined4 VTable0x14() override;                            // vtable+0x14;
```
foxtacles, line 50: maybe we can also remove the trailing semicolon from the annotations

foxtacles (approved):

PR #1436, opened 2025-04-19
Add all `TglImpl::ImplementationData` methods

disinvite:
The 9 classes now use a `DataType` typedef as in the 96 code.

To get all the methods to appear, I added more abstraction functions in `RendererImpl` and `MeshImpl`. That resulted in:

```
0x100a1900 - TglImpl::RendererImpl::CreateDevice(struct Tgl::DeviceDirectDrawCreateData const &) (68.60% -> 98.88%)
0x100a1a00 - TglImpl::RendererImpl::CreateView (77.73% -> 100.00%)
0x100a1f50 - TglImpl::RendererImpl::CreateTexture(int, int, int, void const *, int, int, struct Tgl::PaletteEntry const *) (79.70% -> 100.00%)
0x100a20d0 - TglImpl::RendererImpl::CreateTexture(void) (92.06% -> 100.00%)
0x100a4030 - TglImpl::MeshImpl::DeepClone (46.92% -> 100.00%)
```

`CreateDevice` is 99% because reccmp doesn't pick up the static variable. The function where it lives gets inlined so we can't match it with our current mechanism.

The beta asserts use `pMesh` or `rMesh` to refer to both `MeshImpl` and `MeshBuilderImpl`, so it's not obvious at first glance. But I'm pretty sure these addresses are right.

### DIFF
--- a/LEGO1/tgl/d3drm/impl.h
+++ b/LEGO1/tgl/d3drm/impl.h
@@ -105,6 +105,15 @@ public:
 	inline Result Create();
 	inline void Destroy();
 	inline Result CreateLight(LightType type, float r, float g, float b, LightImpl& rLight);
+	inline Result CreateView(
+		const DeviceImpl& rDevice,
+		const CameraImpl& rCamera,
+		unsigned long x,
+		unsigned long y,
+		unsigned long width,
+		unsigned long height,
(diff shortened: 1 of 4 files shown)


foxtacles (approved):
Nice work!

PR #1439, opened 2025-05-07
Documentation + clearing lots of unknowns

floriandotorg:
Hey everyone!

First off, a big thank you for your amazing work! LEGO Island is was one of my childhood favorites and now see how it all works—brilliant.

Now, based this code, a friend and I are remaking LEGO Island for the browser (https://github.com/floriandotorg/brickolini-island). To do that, I’ve been digging into the codebase as much as possible.

In this PR, I’ve added probably hundreds of `unknowns`. I also started documenting the code a bit—I thought that might be useful to share as well.

Additionally, I’ve included a small AI script I wrote. It uses **Gemini** to ask questions about the codebase. It’s not perfect, but it helps clarify more complex parts here and there.

```bash
#!/bin/bash

output="all_code.txt"
> "$output"

find . -type f \( -name "*.cpp" -o -name "*.h" \) | while read -r file; do
  echo "<$file>" >> "$output"
  cat "$file" >> "$output"
  echo "" >> "$output"
done
```

```python
import os

import dotenv
from google import genai
from google.genai import types

dotenv.load_dotenv()

with open("all_code.txt", "r", encoding="utf-8") as f:
    code_content = f.read()

question = "What does MxDSBuffer? Detailed"

client = genai.Client(api_key=os.environ["GEMINI_API_KEY"])

res = client.models.generate_content(
    model="gemini-1.5-pro",
    contents=f"Here is the code:\n\n{code_content}\n\n{question}",
    config=types.GenerateContentConfig(
        temperature=1,
        system_instruction="You are a master reverse engineer. You will be given a whole codebase and a question. You will need to answer the question based on the codebase.",
    ),
)

print(question)
print(res.text)
```

Unfortunately, the codebase is too large (about **1.5 million tokens**) to feed into more powerful LLMs like GPT-4.1 at once.

This PR isn’t really meant to be merged—it's more of a reference. That said, I think the `unknowns` I’ve added are pretty accurate and could likely be merged into the main codebase.

I'll keep this up to date as I go.

Cheers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ skeleton/
 .DS_Store
 node_modules
 code_embeddings.db
+file_embeddings.json
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ LEGO1.DLL
 all_code.txt
 ask.py
 into-one.sh
+skeleton/
+.DS_Store
+node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ tools/ghidra_scripts/import.log
 # These entries are kept for now since that convention has not always been around.
 ISLE.EXE
 LEGO1.DLL
+all_code.txt
+ask.py
+into-one.sh
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ into-one.sh
 skeleton/
 .DS_Store
 node_modules
+code_embeddings.db
--- a/AI_SUMMARY.md
+++ b/AI_SUMMARY.md
@@ -1,0 +1,209 @@
+# View Management and Rendering
+
+## ViewManager
+- **`ViewManager::ViewManager(Tgl::Renderer*, Tgl::Group*, const OrientableROI*)`**: Constructor.
+- **`ViewManager::IsBoundingBoxInFrustum(const BoundingBox&)`**: Checks if a bounding box is within the view frustum.
+- **`ViewManager::ProjectedSize(const BoundingSphere&)`**: Calculates projected screen size of a bounding sphere.
+- **`ViewManager::Update(float, float)`**: Updates visibility and LOD for all ROIs.
+- **`ViewManager::CalculateFrustumTransformations()`**: Calculates frustum planes and vertices.
+- **`ViewManager::UpdateViewTransformations()`**: Transforms frustum vertices into world space.
+- **`ViewManager::SetFrustrum(float, float, float)`**: Sets frustum parameters (FOV, near, far).
+- **`ViewManager::SetResolution(int, int)`**: Sets screen resolution.
+- **`ViewManager::SetPOVSource(const OrientableROI*)`**: Sets the point of view.
+- **`ViewManager::Pick(Tgl::View*, unsigned long, unsigned long)`**: Ray picking to find a `ViewROI`.
+
+## ViewROI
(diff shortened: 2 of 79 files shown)


ItzSwirlz:
I personally like the m_buttonX and m_buttonY names instead of just m_X and m_Y. In the context of what you are working with it does not matter, but it feels like it's an X and Y value attached to something instead of just a random value.

floriandotorg:
I found it a bit confusing because it appears like a button press state or something similar, but it's rather just the mouse position. Maybe `GetXPostion()` is a better name?

floriandotorg:
Guten Abend @foxtacles,

1. Makes sense. My time is a bit limited, but I'll see what I can do.
2. Alright, as soon as you decided how you wanna do it, I'll happily contribute. Have to say, though, I like the idea of using Doxygen. Esp. since this way, everything is also documented inside the code, making it easier to follow. But you decide and let me know.

foxtacles:
> 2. Alright, as soon as you decided how you wanna do it, I'll happily contribute. Have to say, though, I like the idea of using Doxygen. Esp. since this way, everything is also documented inside the code, making it easier to follow. But you decide and let me know.

We'll consider it, nothing is ruled out yet - my main concern with in-code documentation is that it would add too much noise, and distract from the primary goal of this repo to provide an accurate decompilation. We already have a lot of "decomp" documentation (`reccmp` annotations) that are embdedded in the code, so the documentation would be added on top of it.

Happy to hear opinions from everyone else though.

floriandotorg:
Good point. Maybe a possibility would be a repo with just skeleton code + annotations.

floriandotorg:
You gave me ideas. I wanted to try the skeleton idea and thought I could use AI to create some basic annotations. But man, it just completely documented the code. Now we have a fully functional Doxygen doc—including class hierarchies and diagrams. My hand-made docs are also included: https://github.com/floriandotorg/isle-documentation

CleanShot 2025-05-07 at 22 54 35

floriandotorg:
It even added documentation for unknown functions:

CleanShot 2025-05-07 at 22 57 10

This needs to be validated, of course.

foxtacles:
Looks like a viable approach to me.  We could take a version of that into the `isledecomp` org, but if we want to bootstrap it with AI I'd suggest that all AI-generated parts should be marked in some way. Ideally everything would eventually be manually verified (and improved) - generally we are striving for maximum accuracy and correctness as a team.

PyroFilmsFX:
@floriandotorg Just some AI related insight. To take advantage of smaller, smarter models, you could index the codebase by function or similar into a vector store. Using an agentic approach, you could then fetch the relevant functions and their relationships, providing more targeted context.

I'd also advise giving Llama 4 Scout a go. It's got a 10m context window :)

floriandotorg:
Sounds good, so I'll do the following:

1. I'm gonna prepare some PRs to clear the unknowns, would like to have them in the code base first
2. I like the idea of the vector store, I'll give it a try (I already tried models with larger context, even Gemini 1.5 works, but it showed that they are not smart enough to give some real insights).
3. I will mark everything AI generated with “[AI]” (good point).

After this, when everybody's happy, feel free to move the repo into the org.

floriandotorg:
64,830,386 tokens later..

I tinkered around a lot with the prompt, but there is still A TON of cleaning up to do. Everything is marked with [AI], so we can verify step-by-step: https://floriandotorg.github.io/isle-documentation/

PR #1449, opened 2025-05-08
Refactor MxDSAction to use SetFlags instead of SetUnknown24 across mu…

floriandotorg:
…ltiple files

### DIFF
--- a/ISLE/isleapp.cpp
+++ b/ISLE/isleapp.cpp
@@ -149,7 +149,7 @@ IsleApp::~IsleApp()
 void IsleApp::Close()
 {
 	MxDSAction ds;
-	ds.SetUnknown24(-2);
+	ds.SetFlags(-2);
 
 	if (Lego()) {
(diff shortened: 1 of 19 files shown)


foxtacles:
I think the title of this PR is incorrect. The changes apply to `MxDSObject` instead of `MxDSAction`.

foxtacles:
The way how `MxDSObject::m_unk0x24` is being used throughout the code base does imply that it's not a collection of flags (even though in some places we have named a corresponding parameter `p_flags`).

For instance:

image
image

I don't think the purpose is obvious (not to me, anyway). Unless we discover what it is we should probably consistently keep the `unk` pattern for naming (i.e. `m_unk0x24`, `p_unk0x24` etc.)

floriandotorg:
Very fair points. I'll close this for now. Thanks for checking. I'll be a bit more careful with the other unknowns.

PR #1455, opened 2025-05-09
Rename `FUN_100c8670` to `AddStreamingAction` in `MxDiskStreamControl…

floriandotorg:
…ler`


foxtacles:
I believe we need more information/context before we can name this function. For example, what's the difference between `MxDiskStreamController::FUN_100c8670` and `MxDiskStreamController::FUN_100c7f40`? They both add a streaming action, but to different lists. Both could be given the name `AddStreamingAction`. But add to what? What are the lists used for in each? Once we figured that out, we can give names to those lists, and then we can name the functions.

In more general terms, we really want unambiguous naming that's properly communicating what's happening on a higher level, especially for functions.

floriandotorg:
Excellent points @foxtacles, I'm closing this until we know more.

PR #1457, opened 2025-05-09
Clear unknowns in LegoROI and related classes

floriandotorg:


foxtacles:
Thank you. I've made some changes:

* The "ROI Handler" is a function that allows overriding the color mapping for a given name. I've named everything accordingly.
* `SetColorNamed2` and `SetCustomLodColor2` only forward to the `SetLodColor` function. We can assume it was meant to do something "extra" at some point, but we will probably never know what. Reverted to `FUN_*` naming pattern to indicate purpose is unknown.
* Remove the `*GlobalHandler` notion from the public function names. Using these functions to get data is an internal implementation detail and likely shouldn't be exposed in the names of the interface.
* Renamed a few other functions for consistency.

PR #1460, opened 2025-05-09
Rename functions in `LegoWorldPresenter` : `FUN_10067360` to `LoadWor…

floriandotorg:
…ldPart` and `FUN_100674b0` to `LoadWorldModel`. Update `m_unk0x50` to `m_maxObjectId`.

### DIFF
--- a/LEGO1/lego/legoomni/src/entity/legoworldpresenter.cpp
+++ b/LEGO1/lego/legoomni/src/entity/legoworldpresenter.cpp
@@ -46,7 +46,7 @@ void LegoWorldPresenter::configureLegoWorldPresenter(MxS32 p_legoWorldPresenterQ
 // FUNCTION: LEGO1 0x100665c0
 LegoWorldPresenter::LegoWorldPresenter()
 {
-	m_unk0x50 = 50000;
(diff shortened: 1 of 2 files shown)


LEGO1/lego/legoomni/src/entity/legoworldpresenter.cpp
```diff
@@ -359,8 +359,8 @@ MxResult LegoWorldPresenter::FUN_100674b0(ModelDbModel& p_model, FILE* p_wdbFile
 	action.SetDirection(p_model.m_direction);
 	action.SetUp(p_model.m_up);
 
-	MxU32 objectId = m_unk0x50;
-	m_unk0x50++;
+	MxU32 objectId = m_maxObjectId;
+	m_maxObjectId++;
```
foxtacles, line 363: Given this logic, I think a better name would be `m_nextObjectId` - it uses the current value for the next object, then increments it to get a new ID for the next object and so on

floriandotorg: Good point, changed it.

PR #1469, opened 2025-05-11
Fix definition of WriteRegisterSettings()

AJenbo:
I couldn't find a way to get this working on Linux so hoping this is the correct fix.


foxtacles:
> so hoping this is the correct fix

Yes it is. The function shouldn't be prefixed with the class name in the declaration, but some compilers (MSVC *cough*) may permit it.

PR #1472, opened 2025-05-11
Correct argument names for FUN_100bc8b0

AJenbo:


LEGO1/omni/src/video/mxdisplaysurface.cpp
```diff
@@ -1369,7 +1369,7 @@ void MxDisplaySurface::VTable0x2c(
 }
 
 // FUNCTION: LEGO1 0x100bc8b0
-LPDIRECTDRAWSURFACE MxDisplaySurface::FUN_100bc8b0(MxS32 width, MxS32 height)
+LPDIRECTDRAWSURFACE MxDisplaySurface::FUN_100bc8b0(MxS32 p_width, MxS32 p_height)
```
foxtacles, line 1372: can you adjust the header as well?. Thank you

PR #1475, opened 2025-05-12
Refactor `LegoPathController` variable names

floriandotorg:


LEGO1/lego/legoomni/include/legopathcontroller.h
```diff
@@ -189,12 +189,12 @@ class LegoPathController : public MxCore {
 
 	LegoPathBoundary* m_boundaries; // 0x08
 	LegoPathCtrlEdge* m_edges;      // 0x0c
-	Mx3DPointFloat* m_unk0x10;      // 0x10
+	Mx3DPointFloat* m_nodes;        // 0x10
 	LegoPathStruct* m_structs;      // 0x14
-	MxU16 m_numL;                   // 0x18
```
foxtacles, line 194: You couldn't know this, but these variable names (`m_numL`, ...) were names originally used by the developers so we want to keep them for authenticity (even if they are not very good). We know that due to assertions in the BETA build.

I've attached a CSV with all the strings from the `LEGO1D.DLL` from the BETA, if that helps
[beta.csv](https://github.com/user-attachments/files/20164382/beta.csv)

Or you can explore this using our Ghidra server if you want:

image

jonschz: Maybe this is a good opportunity to add a comment like `// name verified by BETA10 0x<address>` to the variable name. I tried to do that for every BETA10 variable name I found during decompilation.

rawr51919: > Maybe this is a good opportunity to add a comment like `// name verified by BETA10 0x<address>` to the variable name. I tried to do that for every BETA10 variable name I found during decompilation.

Another possible solution is to note the BETA10 equivalent name beside the corrected ones so at least those original developer-written variable names are honored in some way while making the decomp more readable.

PR #1476, opened 2025-05-12
Match `DecodeSS2`

### LINKED ISSUE #1427: Match `DecodeSS2`
For once, the question of what this should do is answered easily: https://github.com/thinkbeforecoding/nomemalloc.handson/blob/master/flic.txt

I have these variables worked out from the beta if it helps. Beyond that, the code is a mess.

```cpp
short xofs = 0;
short yofs = 0;

short width = (short) p_flcHeader->width;
short token = 0;

short xmax = xofs + width - 1;

short row = yofs + p_flcHeader->height - 1;
short* data = (short*) p_data;

// The first word in the data following the chunk header contains the number of lines in the chunk.
// The line count does not include skipped lines.
short lines = *data++;
```

jonschz:
I have significantly improved `DecodeSS2`, but it isn't perfect yet. Would appreciate if someone else took a look.

A lot of `// LINE` annotations are used, based on [this PR](https://github.com/isledecomp/reccmp/pull/107). They aren't technically needed anymore since the structure matches, but it might still be handy to keep them around (maybe as a reference?) ~The CI currently fails because of these annotations~.

Most of the code matches, there are only two parts I couldn't get quite right. The rest comes down to the stack order / variable names.

Closes #1427.

### DIFF
--- a/LEGO1/omni/src/video/flic.cpp
+++ b/LEGO1/omni/src/video/flic.cpp
@@ -368,7 +368,7 @@ void DecodeSS2(LPBITMAPINFOHEADER p_bitmapHeader, BYTE* p_pixelData, BYTE* p_dat
 	union {
 		BYTE* byte;
 		WORD* word;
-	 } data = { p_data };
+	} data = {p_data};
 
 	// The first word in the data following the chunk header contains the number of lines in the chunk.
 	// The line count does not include skipped lines.


LEGO1/omni/src/video/flic.cpp
```diff
@@ -355,65 +355,104 @@ void DecodeLC(LPBITMAPINFOHEADER p_bitmapHeader, BYTE* p_pixelData, BYTE* p_data
...
+	do {
+		// LINE: BETA10 0x1013e692
+		token = *(short*) data;
+		data += 2;
+
+		if (token >= 0) {
+			goto column_loop;
+		}
+		// TODO: Can't get this if-check to be quite right. Union type didn't help either
+		if (token & 0x4000) {
```
jonschz, line 395: I can't make this if-check match. Doesn't look like entropy to me.
```diff
0x1013e6b6 :
-test byte ptr [ebp - 0x17], 0x40
+movsx eax, word ptr [ebp - 0x10]
+test ah, 0x40
```

LEGO1/omni/src/video/flic.cpp
```diff
@@ -355,65 +355,104 @@ void DecodeLC(LPBITMAPINFOHEADER p_bitmapHeader, BYTE* p_pixelData, BYTE* p_data
...
 			}
+			break;
 		}
 
-		short column = 0;
-		do {
-			column += *(data++);
-			short type = *((char*) data++);
-			type += type;
+		// TODO: Something is still slightly incorrect here - orig has two `jmp`s, recomp only has one
```
jonschz, line 412: I was able to match most of the `jmp` instructions, both in number and offset. The two `jmp` instructions suggest that something else might be used here (maybe another loop ends), but everything I tried creates regressions elsewhere.

foxtacles:
@jonschz just making sure, this diff is expected?

Decreased (1):
0x100bdac0 - DecodeSS2 (38.58% -> 23.53%)

LEGO1/omni/src/video/flic.cpp
jonschz: Fixed

jonschz:
I looked again and finally got it to an effective match on LEGO1, with roughly 55 % and a full structural match on BETA10.

foxtacles (approved):
Awesome work!

PR #1477, opened 2025-05-12
build improvements

dtomvan:
This PR improves the build system in various ways, mostly related to the Docker build (on linux).

It improves build times by leveraging `ninja`'s parallel builds. This of course
could also be used when `ninja` is installed on a Windows machine, but I'm not
sure how to document that properly.

Times are on an `AMD Ryzen 5 2600 (12) @ 3.40 GHz`.

Before this PR, with NMake:
```
real    3m1.153s
user    0m0.501s
sys     0m0.848s
```

After, with `JOBS=12`:
```
real    1m14.885s
user    0m0.603s
sys     0m0.737s
```

This PR also introduces the `build.sh` convenience script for Linux, which:
- Copies the source to a temporary directory
- Finds either `podman` or `docker`
- Builds the OCI image
- Mounts the temporary source under `/isle`
- Sets the amount of parallel build jobs to `nproc` by default
- Mounts `/build` as a `tmpfs` so it also gets wiped after the build
- Outputs artifacts (and _only_ the artifacts to `./result`)

This makes it much simpler to build than having to manually set this up as
documented in the README every time, especially if you want the build to be
isolated.

Future improvement could be to document how to use `ninja` on Windows with this
project or even write a PowerShell script for it.


build.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
```
foxtacles, line 1: This file should probably go into the `docker` folder

dtomvan: I used `cd $(dirname "${BASH_SOURCE[0]}")`, so the script kind-of revolves around being in the root dir. Should I append `..` or should I use `git rev-parse --show-toplevel` to get the root directory instead?

CMakeLists.txt
```diff
@@ -472,6 +472,7 @@ if(ISLE_BUILD_LEGO1)
     OUT_TARGETS lego1_targets
   )
   reccmp_add_target(lego1 ID LEGO1)
+  install(TARGETS lego1 DESTINATION bin)
```
madebr, line 475: I'd just install in `.` here.
The isle repo is purely for decompilation project, and will only ever support Windows.
```suggestion
  install(TARGETS lego1 DESTINATION ".")
```
In isle-portable, I'd use [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) and install to _bindir and _libdir. etc

build.sh
foxtacles: `git rev-parse --show-toplevel` seems like the way to go to me, but I'd rather defer to @madebr or @Ramen2X who might be more familiar with builds or the Docker situation.

I believe until now we have generally avoided including build scripts/helpers in the repo, so I can't say whether we'll want to move forward with this in general or not.

CMakeLists.txt
madebr: Can you also apply this change to the other `install`'s as well?

dtomvan: Yes, I am currently rebasing.

build.sh
dtomvan: Yeah, I get that. This isn't needed at all, just nice-to-have.

docker/build.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+# two `cd`'s so Git can find the root dir even if this script is called from
+# outside the repo
+cd $(dirname "${BASH_SOURCE[0]}")
```
madebr, line 7: I think this is more portable.
Some people use zsh etc
```suggestion
cd -P -- "`dirname -- "$0"`"
```

(but perhaps those people are not trying to build isle :) )

docker/build.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
```
madebr, line 3: I'm not sure all flags are supported by all shells, but let's ignore that for now.

docker/entrypoint.sh
```diff
@@ -3,17 +3,23 @@
...
-wine reg ADD 'HKCU\Environment' /v PATH /d 'C:\msvc\bin;C:\msvc\bin\winnt;C:\cmake\bin;C:\windows\system32' /f
+wine reg ADD 'HKCU\Environment' /v PATH /d 'C:\msvc\bin;C:\msvc\bin\winnt;C:\cmake\bin;C:\windows\system32;C:\' /f
 wine reg ADD 'HKCU\Environment' /v INCLUDE /d 'C:\msvc\include;C:\msvc\mfc\include' /f
 wine reg ADD 'HKCU\Environment' /v LIB /d 'C:\msvc\lib;C:\msvc\mfc\lib' /f
 wine reg ADD 'HKCU\Environment' /v TMP /d 'Z:\build' /f
 wine reg ADD 'HKCU\Environment' /v TEMP /d 'Z:\build' /f
 
 # Configure build with CMake
-wine cmake -B build isle -G "NMake Makefiles" $CMAKE_FLAGS
+wine cmake -B build isle -G "Ninja" $CMAKE_FLAGS
```
madebr, line 13: I think the reason we stuck with `NMake Makefiles` was that it has correct dependency tracking.
It did not work well with Ninja.

foxtacles: I remember that too, yes. It might not be a concern if the Docker build/script is used only for "one-off" compilation scenarios?

madebr: Let's assume that

docker/build.sh
dtomvan: Most people have bash installed anyways, hence the bash shebang. But seems fine.

docker/entrypoint.sh
dtomvan: The build definitely "works". This `build.sh` does everything isolated from the source repo, throws all intermediate artifacts away, and hence needs to recompile everything every time, yes. But at least a full rebuild is much faster this way. There still seems to be a lot of "critical path moments" in the build, where it still compiles everything one-by-one. That could be improved.

docker/build.sh
dtomvan: `dash` doesn't support `pipefail`, we could remove that?

docker/entrypoint.sh
dtomvan: I think I know what you mean. The readme says:
> NMake Makefiles is most recommended because it will be immediately compatible with Visual C++ 4.2. For faster builds, you can use Ninja (if you have it installed), however due to limitations in Visual C++ 4.2, you can only build Release builds this way (debug symbols cannot be generated with Ninja).

So if I understand it correctly, Ninja could be used to build all release builds, even incremental ones?

madebr: incremental builds are the trouble makers when using an older MSVC: files whose includes have changed will not rebuild.

dtomvan: Hmm. That's a little weird to me. Ninja is basically just a command runner, so if CMake can generate something that works for NMake, why would that break if all CMake does is (theoretically, I haven't looked at CMake's output myself) generate something equivalent for Ninja?

madebr: We're talking about 1997-era tooling. Ninja works fine for today's MSVC toolchains.
Until I understand the issue better, I'm assuming there's an external communication channel between nmake and cl.

dtomvan: Yeah, I guess it is *that* old. Should we split it into "I just want to build once" and "I want incremental builds" or drop ninja at all? But in that case this PR isn't worth much anymore.

dtomvan:
Another reduction, with `USE_NIX_BUILD=1`, I got the size to `1.15GB` _with_ ninja (due to WoW64 being needed), and _without Ninja_ all the way to `741MB`!!! I know I'm pushing Nix a little too hard here, but the actual code at hand is not that difficult, it's way more minimal and usable, and it works just fine.

foxtacles:
Thank you for this effort, but I think we won't add this to the repo for now.

PR #1481, opened 2025-05-13
Fix wrong use of `delete` vs `delete[]`

foxtacles:
As seen in this video: https://github.com/isledecomp/isle-portable/pull/76

CC @AJenbo


AJenbo:
I was hopping someone would notice, thanks :)

Issue #1485, opened 2025-05-14
What are the constants used for lighting


AJenbo:
https://github.com/isledecomp/isle/blob/88651404cbca2c6771e6331747f948a5f02a3655/LEGO1/lego/legoomni/src/common/legogamestate.cpp#L1307

My guess is it's simply to make it fit the ambient sky color:

![Image](https://github.com/user-attachments/assets/bb154a84-daa3-4bcd-8629-a5edf4e17f35)

PR #1486, opened 2025-05-14
Rename F2bit0 to LacksLightSupport

AJenbo:
A better name for the flag. It's one of the most common places for the miniwin directx implementation to die so I have been starting at this very bit of code for to many times to live with this name any longer.

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legogamestate.cpp
+++ b/LEGO1/lego/legoomni/src/common/legogamestate.cpp
@@ -1303,7 +1303,7 @@ void LegoBackgroundColor::ToggleSkyColor()
 // FUNCTION: BETA10 0x10086984
 void LegoBackgroundColor::SetLightColor(float p_r, float p_g, float p_b)
 {
-	if (!VideoManager()->GetVideoParam().Flags().GetF2bit0()) {
+	if (!VideoManager()->GetVideoParam().Flags().GetLacksLightSupport()) {
 		// TODO: Computed constants based on what?
 		p_r *= 1. / 0.23;
 		p_g *= 1. / 0.63;
--- a/LEGO1/lego/legoomni/src/video/legovideomanager.cpp
+++ b/LEGO1/lego/legoomni/src/video/legovideomanager.cpp
@@ -126,13 +126,13 @@ MxResult LegoVideoManager::Create(MxVideoParam& p_videoParam, MxU32 p_frequencyM
 	m_direct3d->SetDevice(deviceEnumerate, driver, device);
 
 	if (!driver->m_ddCaps.dwCaps2 && driver->m_ddCaps.dwSVBRops[7] != 2) {
-		p_videoParam.Flags().SetF2bit0(TRUE);
+		p_videoParam.Flags().SetLacksLightSupport(TRUE);
 	}
 	else {
-		p_videoParam.Flags().SetF2bit0(FALSE);
+		p_videoParam.Flags().SetLacksLightSupport(FALSE);
 	}
 
-	ViewROI::SetUnk101013d8(p_videoParam.Flags().GetF2bit0() == FALSE);
+	ViewROI::SetUnk101013d8(p_videoParam.Flags().GetLacksLightSupport() == FALSE);
 
 	if (!m_direct3d->Create(
 			hwnd,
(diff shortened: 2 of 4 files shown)


LEGO1/lego/legoomni/src/video/legovideomanager.cpp
```diff
@@ -125,14 +125,18 @@ MxResult LegoVideoManager::Create(MxVideoParam& p_videoParam, MxU32 p_frequencyM
...
 	m_direct3d->SetDevice(deviceEnumerate, driver, device);
 
-	if (!driver->m_ddCaps.dwCaps2 && driver->m_ddCaps.dwSVBRops[7] != 2) {
-		p_videoParam.Flags().SetF2bit0(TRUE);
+	/*
+	 * BUG: should be:
+	 *  if ((driver->m_ddCaps.dwCaps2 & DDCAPS2_CERTIFIED) == 0 && driver->m_ddCaps.dwSVBRops[7] != 2) {
+	 */
+	if (driver->m_ddCaps.dwCaps2 != DDCAPS2_CERTIFIED && driver->m_ddCaps.dwSVBRops[7] != 2) {
+		p_videoParam.Flags().SetBrokenLighting(TRUE);
```
foxtacles, line 133: I don't think we should embed the fact that the condition is broken into the naming of the function :D  I'd suggest `SetDisableLighting`, as that seems to be what it's doing

LEGO1/lego/legoomni/src/common/legogamestate.cpp
```diff
@@ -1303,7 +1303,7 @@ void LegoBackgroundColor::ToggleSkyColor()
 // FUNCTION: BETA10 0x10086984
 void LegoBackgroundColor::SetLightColor(float p_r, float p_g, float p_b)
 {
-	if (!VideoManager()->GetVideoParam().Flags().GetF2bit0()) {
+	if (!VideoManager()->GetVideoParam().Flags().GetBrokenLighting()) {
 		// TODO: Computed constants based on what?
 		p_r *= 1. / 0.23;
```
disinvite, line 1308: The beta supports changing these to:
```cpp
p_r /= 0.23;
p_g /= 0.63;
p_b /= 0.85;
```
if you wanted to do it here. Could be its own PR too.

LEGO1/lego/legoomni/src/video/legovideomanager.cpp
AJenbo: Ha ha, no that name isn't related to the borken code, we could call it GetLacksLightSupport() instead to make that more clear.

foxtacles: Oh I see. Yes, let's call it something other than "broken". `LacksLightSupport` sounds reasonable

LEGO1/lego/legoomni/src/common/legogamestate.cpp
AJenbo: I'm thinking that's better addressed with this issue: https://github.com/isledecomp/isle/issues/1485

LEGO1/lego/legoomni/src/video/legovideomanager.cpp
AJenbo: Done

LEGO1/lego/legoomni/src/video/legovideomanager.cpp
```diff
@@ -125,14 +125,18 @@ MxResult LegoVideoManager::Create(MxVideoParam& p_videoParam, MxU32 p_frequencyM
 
 	m_direct3d->SetDevice(deviceEnumerate, driver, device);
 
-	if (!driver->m_ddCaps.dwCaps2 && driver->m_ddCaps.dwSVBRops[7] != 2) {
-		p_videoParam.Flags().SetF2bit0(TRUE);
+	/*
+	 * BUG: should be:
+	 *  if ((driver->m_ddCaps.dwCaps2 & DDCAPS2_CERTIFIED) == 0 && driver->m_ddCaps.dwSVBRops[7] != 2) {
+	 */
+	if (driver->m_ddCaps.dwCaps2 != DDCAPS2_CERTIFIED && driver->m_ddCaps.dwSVBRops[7] != 2) {
```
foxtacles, line 132: Unfortunately this does decrease the assembly accuracy:

```
Decreased (1):
0x1007ac40 - LegoVideoManager::Create (98.63% -> 98.17%)
```

image

AJenbo: Maybe they really did check against it being empty as a way to tell if it's a DirectX 5 feature level driver or something.

AJenbo: I'm not going to juggle the ways this could be written or setup the needed tools to make a comparable version so I have reduced this PR to just be the name change for the flag.

PR #1494, opened 2025-05-15
Use consts in callbacks

AJenbo:
`const` qualifiers are stripped during compilation and don't affect the generated code — they're primarily a tool to help developers avoid unintended mutations.

This change adds `const` in several places where it makes sense for clarity and code safety. While we can't be certain whether these were originally present (since decompiled code doesn't retain them), adding them improves code quality and intent.

### DIFF
--- a/LEGO1/mxdirectx/mxdirectxinfo.h
+++ b/LEGO1/mxdirectx/mxdirectxinfo.h
@@ -67,16 +67,16 @@ struct Direct3DDeviceInfo {
 	~Direct3DDeviceInfo();
 	Direct3DDeviceInfo(
 		LPGUID p_guid,
-		LPSTR p_deviceDesc,
-		LPSTR p_deviceName,
+		LPCSTR p_deviceDesc,
+		LPCSTR p_deviceName,
 		LPD3DDEVICEDESC p_HWDesc,
 		LPD3DDEVICEDESC p_HELDesc
 	);
 
 	void Initialize(
 		LPGUID p_guid,
-		LPSTR p_deviceDesc,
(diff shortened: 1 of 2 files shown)


LEGO1/mxdirectx/mxdirectxinfo.h
```diff
@@ -194,19 +194,19 @@ class MxDeviceEnumerate {
...
+		LPCSTR p_deviceDesc,
+		LPCSTR p_deviceName,
 		LPD3DDEVICEDESC p_HWDesc,
 		LPD3DDEVICEDESC p_HELDesc
 	);
 	const char* EnumerateErrorToString(HRESULT p_error);
 	static void BuildErrorString(const char*, ...);
 	static BOOL CALLBACK
-	DirectDrawEnumerateCallback(LPGUID p_guid, LPSTR p_driverDesc, LPSTR p_driverName, LPVOID p_context);
+	DirectDrawEnumerateCallback(LPGUID p_guid, LPCSTR p_driverDesc, LPCSTR p_driverName, LPVOID p_context);
```
madebr, line 209: I think you have to assume callbacks can modify its arguments.

Or you explicitly cast with `(LPDDENUMCALLBACKA)`, but then you lose type safety.

AJenbo: Yeah it looks like i sadly can't provide a stricter implementation of the callbacks, I  have reverted that part which should let it compile. But unfortunately that means it doesn't help me on the miniwin implementation.

madebr (approved):
Pending accuracy review by the reversing gang :)

PR #1502, opened 2025-05-17
Rename FUN_10018980 into SetShootAnimMap

doggywatty:
Renames the animation map with GetUnknown0x00() == -1.0f as the "shoot" animation and stores it in m_shootAnim. Ensures that a valid shoot animation was found. Retrieves the "xarrow" sound from the sound manager and assigns it to m_unk0x38, unless the BETA10 flag is set, in which case it uses the "bcrash" sound instead. Sets sound attenuation distances depending on the build, and in non-BETA10 builds, makes the ROI visible.

### DIFF
--- a/LEGO1/lego/legoomni/include/act2actor.h
+++ b/LEGO1/lego/legoomni/include/act2actor.h
@@ -37,7 +37,7 @@ public:
 	MxResult VTable0x9c() override;                     // vtable+0x9c
 	MxS32 VTable0xa0() override;                        // vtable+0xa0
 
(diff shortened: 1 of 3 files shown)


foxtacles:
That name doesn't really make sense to me. It also sets `m_unk0x38`, so it's not just limited to setting the shoot anim. Seems more like a generic initialization function. We'd have to resolve more of the involved unknowns to understand it better before we can give it a good name.

doggywatty:
Oh alright, I understand.

PR #1505, opened 2025-05-18
Use sizeof(type) instead of sizeof(variableName) when (de)serializing data

madebr:
Ideally, this should cause no accuracy regression.


foxtacles:
We seem to have a huge amount of entropy accuracy changes for some reason..I think we'll have to run entropy builds to cut down on the noise and check the remaining diffs manually.

@madebr can you run the workflow on the branch?

madebr:
> can you run the workflow on the branch?

I started it on my fork:
https://github.com/madebr/isle/actions/runs/15099671260

foxtacles:
Ran the decomp and everything works as expected.

PR #1512, opened 2025-05-20
Rename functions and parameters in OrientableROI

estchd:
Gave all unnamed functions in `OrientableROI` understandable names based on what they do, how and where they are called and other similar functions in `OrientableROI`.

Also renamed some previously named functions in `OrientableROI` in case the previous names conflicted with the new names or where misleading.

Reasons for each renaming:

- `OrientableROI::VTable0x14` -> `OrientableROI::WrappedUpdateWorldData`: 
The function wraps `OrientableROI::VTable0x1c` and i renamed that to `OrientableROI::UpdateWorldData` so i named this like other wrapping methods in `OrientableROI`

- `OrientableROI::VTable0x1c` -> `OrientableROI::UpdateWorldData`: 
The function does the same as the function that was previously named `OrientableROI::UpdateWorldData` but without taking a "parent" transform or updating child transforms so i made this function the base name and renamed the other methods to show what extra work they do

- `OrientableROI::SetLocalTransform` -> `OrientableROI::SetLocal2WorldWithWorldDataUpdate`:
The function name was misleading because "LocalTransform" could mean one of two things, either "local2world" or "local2parent". Since the function updates the former i named it as such.
Since there is another function that only updates the `local2world` transform, i also renamed this function to show that it also updates the world data

- `OrientableROI::Vtable0x24` -> `OrientableROI::UpdateWorldDataWithTransform`:
The function does the same as the function that is now named `OrientableROI::UpdateWorldData` but also takes a transform to apply

- `OrientableROI::UpdateWorldData` -> `OrientableROI::UpdateWorldDataWithTransformAndChildren`:
The function does the same as the function that is now named  `OrientableROI::UpdateWorldDataWithTransform` but also applies that transform to the ROIs Children

- `OrientableROI::WrappedSetLocalTransform` -> `OrientableROI::WrappedSetLocal2WorldWithWorldDataUpdate`:
`OrientableROI::SetLocalTransform` got renamed

- `OrientableROI::WrappedVTable0x24` -> `OrientableROI::WrappedUpdateWorldDataWithTransform`:
`OrientableROI::VTable0x24` got renamed

- `OrientableROI::FUN_100a58f0` -> `OrientableROI::SetLocal2World`:
The function does the same as the function that was previously named `OrientableROI::SetLocalTransform` but without updating the world data like that function does

- `OrientableROI::FUN_100a5a30` -> `OrientableROI::SetWorldVelocity`:
The only thing the function does is setting the world velocity

- `OrientableROI::ToggleUnknown0xd8` -> `OrientableROI::SetNeedsWorldDataUpdate`:
This rename is speculative but the function is only called in two places, `OrientableROI::OrientableROI`, which does not update the world data and the specific `OrientableROI::SetLocal2World` variant does not update the world data.
It is weird, that this flag is never read, but almost every context that calls `OrientableROI::SetLocal2World` also immediately calls `OrientableROI::UpdateWorldData` in some form so that might just be unnecessary.

Some notes:

- `OrientableROI::m_unk0xd8` appears to only be storing a `needsWorldDataUpdate` Flag so it could be renamed as such. I did not rename it because it bit fiddling in addition to storing the flag that i am not sure why it exists.

- Many of the functions do similar things as other functions but add functionality.
 Could it be possible that those functions were originally calling each other but got inlined to produce what looks like a lot of duplicated code? 
I don't know enough about decompilation to tell if it is that or if it is just how the code is.

- I am not sure about the true purpose of the `p_transform` parameter for the `UpdateWorldData` methods because it is used in different ways in different contexts so i didn't rename it.

### DIFF
--- a/LEGO1/lego/legoomni/src/actors/act2actor.cpp
+++ b/LEGO1/lego/legoomni/src/actors/act2actor.cpp
@@ -611,8 +611,8 @@ MxU32 Act2Actor::FUN_10019700(MxFloat p_param)
 
 	MxMatrix matrix = m_roi->GetLocal2World();
 	matrix[3][1] += 1.0f;
-	brickstrROI->FUN_100a58f0(matrix);
-	brickstrROI->VTable0x14();
+	brickstrROI->SetLocal2World(matrix);
+	brickstrROI->WrappedUpdateWorldData();
 
 	Vector3 col0(matrix[0]);
 	Vector3 col1(matrix[1]);
--- a/LEGO1/lego/legoomni/src/actors/act3actors.cpp
+++ b/LEGO1/lego/legoomni/src/actors/act3actors.cpp
@@ -138,8 +138,8 @@ MxU32 Act3Actor::VTable0x90(float p_time, Matrix4& p_transform)
 			m_unk0x1c = 0;
 
 			positionRef -= g_unk0x10104ef0;
-			m_roi->FUN_100a58f0(p_transform);
-			m_roi->VTable0x14();
+			m_roi->SetLocal2World(p_transform);
+			m_roi->WrappedUpdateWorldData();
 			return TRUE;
 		}
 	}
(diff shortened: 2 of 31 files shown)


foxtacles:
> Could it be possible that those functions were originally calling each other but got inlined to produce what looks like a lot of duplicated code? I don't know enough about decompilation to tell if it is that or if it is just how the code is.

We could take a look at the BETA 1.0. If they used inline functions, they would show up there.

Thanks for the name changes, it all looks reasonable to me

PR #1515, opened 2025-05-24
Refactor `LegoUnknown100db7f4` to `LegoOrientedEdge`

floriandotorg:


floriandotorg:
Verification fails, can we not rename classes without braking compatibly?

LEGO1/lego/sources/geom/legoorientededge.h
```diff
@@ -101,20 +101,20 @@ struct LegoUnknown100db7f4 : public LegoEdge {
 	// SYNTHETIC: LEGO1 0x1009a6c0
 	// LegoUnknown100db7f4::`scalar deleting destructor'
```
jonschz, line 102: You missed a rename in this line, that's why the pipeline fails

PR #1520, opened 2025-05-25
Fix ambiguous matches in BETA10

jonschz:
Gets rid of all _Ambiguous match_ warnings in BETA10. There is no diff in the match %. I checked every entry whether the other match might be more correct, and I added comments in a few places that were truly ambiguous (i.e. no diff for either case).

### DIFF
--- a/LEGO1/omni/src/video/flic.cpp
+++ b/LEGO1/omni/src/video/flic.cpp
@@ -364,10 +364,10 @@ void DecodeSS2(LPBITMAPINFOHEADER p_bitmapHeader, BYTE* p_pixelData, BYTE* p_dat
 	// LINE: BETA10 0x1013e643
 	short xmax = xofs + width - 1;
 
-	// LINE: BETA10 0x1013e652
 	union {
 		BYTE* byte;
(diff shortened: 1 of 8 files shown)


LEGO1/omni/src/video/flic.cpp
```diff
@@ -364,10 +364,10 @@ void DecodeSS2(LPBITMAPINFOHEADER p_bitmapHeader, BYTE* p_pixelData, BYTE* p_dat
 	// LINE: BETA10 0x1013e643
 	short xmax = xofs + width - 1;
 
-	// LINE: BETA10 0x1013e652
 	union {
 		BYTE* byte;
 		WORD* word;
+		// LINE: BETA10 0x1013e652
```
jonschz, line 370: Small drive-by, I missed this in a previous PR

jonschz:
> There seems to be a diff in some functions

Thanks for checking, this genuinely surprised me. Can't tell for certain what's going on, but the issue seems to be fixed for now.

foxtacles (approved):

PR #1538, opened 2025-06-03
Add names to LegoActorInfo.Part

xZise:
Each part consists of two "blocks" where every block has a list of indices which reference into a list of names and the actual index used for the list of indices.

By doing that, it is possible to reuse the list of name for every actor and instead have a "sublist" through the indices list.


foxtacles:
Thank you!

PR #1541, opened 2025-06-05
Name for bounding box in ROI

xZise:


LEGO1/realtime/orientableroi.h
```diff
@@ -70,7 +70,7 @@ class OrientableROI : public ROI {
 protected:
 	MxMatrix m_local2world;                 // 0x10
 	BoundingBox m_world_bounding_box;       // 0x58
-	BoundingBox m_unk0x80;                  // 0x80
+	BoundingBox m_boundingBox;              // 0x80
```
foxtacles, line 73: Can you use snake_case for this name here? (only inside this class) Thank you

PR #1542, opened 2025-06-05
Clear unknowns in 'LegoCarBuild' and 'LegoCarBuildPresenter'

MasterTimeThief:
Mostly just renamed variables and functions that hadn't been named, based on what I discovered they do.

### DIFF
--- a/LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
+++ b/LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
@@ -355,7 +355,7 @@ void LegoCarBuildAnimPresenter::InitBuildPlatform()
 	LegoAnimNodeData* destData;
 	LegoTreeNode** children;
 
-	//Get Platform data, Shelf Frame data, and number of build parts
+	// Get Platform data, Shelf Frame data, and number of build parts
 	for (i = 0; i < totalNodes; i++) {
 		LegoAnimNodeData* data = (LegoAnimNodeData*) GetTreeNode(m_anim->GetRoot(), i)->GetData();
 		name = data->GetName();
@@ -385,7 +385,7 @@ void LegoCarBuildAnimPresenter::InitBuildPlatform()
 	m_parts = new UnknownListEntry[m_numberOfParts];
 	assert(m_parts);
 
-	//Go through and add the wired name of each part
+	// Go through and add the wired name of each part
 	for (i = 0; i < totalNodes; i++) {
 		name = ((LegoAnimNodeData*) GetTreeNode(m_anim->GetRoot(), i)->GetData())->GetName();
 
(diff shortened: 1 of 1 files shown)


LEGO1/lego/legoomni/include/legocarbuildpresenter.h
```diff
@@ -87,10 +87,10 @@ class LegoCarBuildAnimPresenter : public LegoAnimPresenter {
 	void SetPartObjectIdByName(const LegoChar* p_name, MxS16 p_objectId);
 
 	// FUNCTION: BETA10 0x10070180
-	void SetUnknown0xbc(undefined2 p_unk0xbc) { m_unk0xbc = p_unk0xbc; }
+	void SetShelfState(undefined2 p_state) { m_shelfState = p_state; }
```
foxtacles, line 90: the parameter should be named like the member it sets (i.e. `p_shelfState`)

LEGO1/lego/legoomni/src/build/legocarbuild.cpp
```diff
@@ -96,7 +96,7 @@ LegoCarBuild::LegoCarBuild()
 	m_unk0x110 = 0;
 	m_unk0xf8 = c_unknownminusone;
 	m_unk0x2d4 = FALSE;
-	m_unk0x258 = 0;
+	m_animPresenter = 0;
```
foxtacles, line 99: should be `= NULL`

LEGO1/lego/legoomni/src/build/legocarbuildpresenter.cpp
```diff
@@ -114,14 +114,14 @@ inline void LegoCarBuildAnimPresenter::Beta10Inline0x100733d0()
 // FUNCTION: BETA10 0x10070ab1
 void LegoCarBuildAnimPresenter::PutFrame()
 {
-	switch (m_unk0xbc) {
-	case 0:
+	switch (m_shelfState) {
+	case 0:	//Selected piece for moving / recoloring / decal
```
foxtacles, line 118: If you figured out the meaning of these values it might be worth declaring an enum for it (see other classes for examples)

LEGO1/lego/legoomni/include/legocarbuildpresenter.h
```diff
@@ -116,29 +123,29 @@ class LegoCarBuildAnimPresenter : public LegoAnimPresenter {
 private:
 	void Beta10Inline0x100733d0();
 
-	MxU16 m_unk0xbc; // 0xbc
+	ShelfState m_shelfState; // 0xbc
```
foxtacles, line 126: The type of this needs to remain `MxU16`. Using `ShelfState` will cause it to be an `int`

MasterTimeThief:
Added `InitBuildPlatform` which sets up stuff for the build game. It's really interesting and fun picking the code apart and seeing how it works!

PR #1545, opened 2025-06-07
Clear unknowns/numbers in `Hospital` and `HospitalState`

xZise:
I've tried to clear unknowns in `Hospital`, and maybe the names are not the best.

For example I'm not really sure about those:
* `m_interactionMode` is weird, as the values seem to be `0`, `1` and `3`. Not sure if those specific values have been used elsewhere.
* `m_setWithCurrentAction` is only set and never read. But it is always (and only) set when an animation is started.
* `g_deleteCounter` just "enables" you to skip the animation with space. Not sure why it is 3 or decremented with each tickle.

If you want I can split this into separate pull requests, depending on the "likeliness" of being correct.

While working on the states I noticed that there the state 10 can never be reached, and because of this 11 is also out of reach. Not sure if this a decompilation artefact or unfinished code. In theory state 4 can never be reached and I'm not sure what the difference is to state 3, which is why those are all `unknown`.

### DIFF
--- a/LEGO1/lego/legoomni/src/worlds/hospital.cpp
+++ b/LEGO1/lego/legoomni/src/worlds/hospital.cpp
@@ -43,7 +43,7 @@ Hospital::Hospital()
 	m_currentAction = HospitalScript::c__StartUp;
 	m_copLedBitmap = NULL;
 	m_pizzaLedBitmap = NULL;
-	m_unk0x118 = 0;
+	m_flashingLeds = 0;
 	m_copLedAnimTimer = 0;
 	m_pizzaLedAnimTimer = 0;
 	m_unk0x128 = 0;
@@ -250,7 +250,7 @@ MxLong Hospital::HandleEndAction(MxEndActionNotificationParam& p_param)
 
 		m_currentAction = HospitalScript::c_hho006cl_RunAnim;
 		m_unk0x108 = 1;
-		m_unk0x118 = 1;
+		m_flashingLeds = 1;
 		g_unk0x100f7918 = 0;
 		break;
 	case 6:
@@ -643,7 +643,7 @@ MxResult Hospital::Tickle()
 
 	MxLong time = Timer()->GetTime();
 
-	if (m_unk0x118 != 0) {
+	if (m_flashingLeds != 0) {
 		if (time - m_copLedAnimTimer > 300) {
 			m_copLedAnimTimer = time;
 			g_copLedEnabled = !g_copLedEnabled;
--- a/LEGO1/lego/legoomni/src/worlds/hospital.cpp
+++ b/LEGO1/lego/legoomni/src/worlds/hospital.cpp
@@ -61,7 +61,7 @@ Hospital::~Hospital()
 	ControlManager()->Unregister(this);
 	TickleManager()->UnregisterClient(this);
 
-	m_hospitalState->m_state = 3;
+	m_hospitalState->m_state = HospitalState::e_unknown3;
 
 	NotificationManager()->Unregister(this);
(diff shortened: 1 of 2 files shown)


LEGO1/lego/legoomni/src/worlds/hospital.cpp
```diff
@@ -559,16 +559,16 @@ MxBool Hospital::HandleControl(LegoControlManagerNotificationParam& p_param)
 			BackgroundAudioManager()->RaiseVolume();
 			DeleteObjects(&m_atomId, HospitalScript::c_hho002cl_RunAnim, HospitalScript::c_hho006cl_RunAnim);
 
-			if (m_unk0x100 == 1) {
-				m_hospitalState->m_unk0x08 = 14;
+			if (m_interactionMode == 1) {
+				m_hospitalState->m_state = HospitalState::e_exitViaInfo;
```
xZise, line 563:
```suggestion
				m_hospitalState->m_state = HospitalState::e_exitViaDoor;
```
I've noticed that clicking the info/door results in different destinations depending on when you click them. Should I fix it here or in a different pull request?

If this change should only convert numbers into their constants it should be changed here. If the issue should be fixed, I would need to change the state below.

I've compared it to the original game and there it works as expected and not depending on the timing.

LEGO1/lego/legoomni/src/worlds/hospital.cpp
```diff
@@ -580,16 +580,16 @@ MxBool Hospital::HandleControl(LegoControlManagerNotificationParam& p_param)
 		case HospitalScript::c_Door_Ctl:
 			DeleteObjects(&m_atomId, HospitalScript::c_hho002cl_RunAnim, HospitalScript::c_hho006cl_RunAnim);
 
-			if (m_unk0x100 == 1) {
-				m_hospitalState->m_unk0x08 = 15;
+			if (m_interactionMode == 1) {
+				m_hospitalState->m_state = HospitalState::e_exitViaInfo;
```
xZise, line 584:
```suggestion
				m_hospitalState->m_state = HospitalState::e_exitViaDoor;
```

The current code is (imho) correct, but not a direct translation from the previous code (see above).

xZise (commented):
Currently it shouldn't be merged as it is "half" broken.

LEGO1/lego/legoomni/src/worlds/hospital.cpp
foxtacles: It seems that the correct value should be exitViaInfo, since it maps to 14.

Below, it should be exitViaDoor because it's 15.

We should leave the mappings the way they are - according to the accuracy comparison, HandleControl matches 100% so functionally it should be correct and act the same way as the original. If you see a difference in game behavior regardless (please double check), the issue is likely not to be found in this function.

https://github.com/isledecomp/isle/actions/runs/15517986153/job/43687412923?pr=1545 You can see that currently based on the state of the branch the accuracy has regressed. This is likely to the use of exitViaInfo

xZise: Okay I'm going to double check this.

It seems to me that you switched up the values for `exitVia...` and that this change should be applied to `exitViaDoor` (at least according to the values they represent). I named them this way, because they lead to the "normal" behavior (`exitViaDoor` positions you outside the building and `exitViaInfo` spawns you in the info center). Should I name them the over way around?

foxtacles: Sorry, I'm a bit confused with the naming changes. What matters in the end that the values don't change. I'll leave the naming up to you.

PR #1551, opened 2025-06-10
Clear unknowns in `PoliceState`

xZise:
While in others (e.g. Hospital) a field like `m_unk0x0c` is used as a state and can have multiple values it appears to be that the police has only two states: Whether an animation is playing or not.

If you want I can use a name like `m_state` and some enum with `e_noAnimation` and `e_playingAnimation`.


foxtacles:
> While in others (e.g. Hospital) a field like `m_unk0x0c` is used as a state and can have multiple values it appears to be that the police has only two states: Whether an animation is playing or not.
> 
> If you want I can use a name like `m_state` and some enum with `e_noAnimation` and `e_playingAnimation`.

Actually, that would be great. Maybe you can open another PR.

PR #1560, opened 2025-06-13
Gasstation

xZise:
I renamed the global animation counter variable as it clashed with the one in `hospital.cpp`. Should I also add a commit renaming the variable in `hospital.cpp` to a similar style?

Also maybe the waiting state is overkill in `GasStation`.


foxtacles:
> I renamed the global animation counter variable as it clashed with the one in `hospital.cpp`. Should I also add a commit renaming the variable in `hospital.cpp` to a similar style?

Yes please

> Also maybe the waiting state is overkill in `GasStation`.

Seems fine to me. We want to avoid magic numbers whereever we can

xZise:
I've renamed the global variable in `hospital.cpp` as well.

PR #1579, opened 2025-06-22
Add reference to `isle-portable` to README

### LINKED ISSUE #1578: Add section early in readme regarding this not being portable
It's a common point of confusion (not just on this project) for users to seek out the "original" project and not understanding that it's purly for reverse engineering and not the version that will run well on Tempel OS.

To mitigate this it might be good to add a section early in the readme stating that they should seek out https://github.com/isledecomp/isle-portable instead if that is there goal.

jonschz:
Closes #1578. Feedback on the formulation is welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
   
 This is a functionally complete decompilation of LEGO Island (Version 1.1, English). It aims to be as accurate as possible, matching the recompiled instructions to the original machine code as much as possible. The goal is to provide a workable codebase that can be modified, improved, and ported to other platforms later on.
 
-> **Note:** This repository is for decompilation only and its code is true to the original release. For a modern adaptation of the LEGO Island codebase with native compatibility for Windows, Mac, Linux, and Web, see [isle-portable](https://github.com/isledecomp/isle-portable) instead.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,8 @@
   
 This is a functionally complete decompilation of LEGO Island (Version 1.1, English). It aims to be as accurate as possible, matching the recompiled instructions to the original machine code as much as possible. The goal is to provide a workable codebase that can be modified, improved, and ported to other platforms later on.
 
+> **Note:** This repository is for decompilation only and its code is true to the original release. For a modern adaptation of the LEGO Island codebase with native compatibility for Windows, Mac, Linux, and Web, see [isle-portable](https://github.com/isledecomp/isle-portable) instead.
```
Ramen2X, line 7: We maybe could simplify it and say "native compatibility for all major platforms and the Web" rather than listing out the platforms? We support more than Windows/Mac/Linux in isle-portable so it might be better to word it more inclusively.

Ramen2X, line 7: Also, we may want to add a second line in the middle here; something like: "It will not compile on platforms other than 32-bit Windows." just to be absolutely clear.

Ramen2X (approved):

foxtacles (approved):

PR #1580, opened 2025-06-22
Clear unknowns in `LegoAnimActorEntry` and `LegoAnimKey`

xZise:
Unfortunately I wasn't able to determine what those actor types mean, but it is something. The other change, renaming `FUN_100739a0`, clears the last change related to #1565.


LEGO1/lego/sources/anim/legoanim.h
```diff
@@ -289,8 +289,16 @@ class LegoAnimNodeData : public LegoTreeNodeData {
 
 // SIZE 0x08
 struct LegoAnimActorEntry {
-	LegoChar* m_name;     // 0x00
-	undefined4 m_unk0x04; // 0x04
+	enum {
+		e_actorType2,
```
foxtacles, line 293: enums begin with 0 implicitly. you need to assign a different initial value here

xZise: Oh you are correct! Will fix.

PR #1582, opened 2025-06-22
Update legogamestate.cpp

foxtacles:
This assert doesn't make sense because `p_scorehist` is always `NULL`.

CC @jonschz


jonschz:
my bad, I thought I had removed all assertions from the LEGO1 code. Must have slipped through.

PR #1590, opened 2025-06-26
Clear unknown in `LegoLOD`

xZise:
The function does mostly the same as `SetTextureInfo` but does not explicitly set the color as well.

### DIFF
--- a/LEGO1/lego/sources/roi/legolod.h
+++ b/LEGO1/lego/sources/roi/legolod.h
@@ -31,7 +31,7 @@ public:
 	LegoLOD* Clone(Tgl::Renderer* p_renderer);
 	LegoResult SetColor(LegoFloat p_red, LegoFloat p_green, LegoFloat p_blue, LegoFloat p_alpha);
 	LegoResult SetTextureInfo(LegoTextureInfo* p_textureInfo);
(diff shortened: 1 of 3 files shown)


LEGO1/lego/sources/roi/legolod.h
```diff
@@ -31,7 +31,7 @@ class LegoLOD : public ViewLOD {
 	LegoLOD* Clone(Tgl::Renderer* p_renderer);
 	LegoResult SetColor(LegoFloat p_red, LegoFloat p_green, LegoFloat p_blue, LegoFloat p_alpha);
 	LegoResult SetTextureInfo(LegoTextureInfo* p_textureInfo);
-	LegoResult FUN_100aad70(LegoTextureInfo* p_textureInfo);
+	LegoResult SetTextureInfoWithoutColor(LegoTextureInfo* p_textureInfo);
```
foxtacles, line 34: I think I'd call this `UpdateTextureInfo` - it seems to me like `SetTextureInfo` must always be used first since it also sets an essential flag `m_textured`, and `UpdateTextureInfo` can only be used subsequently to update the texture only.

xZise: Makes sense and describes the usecase better.

PR #1594, opened 2025-06-28
Clear unknowns in GetSoundId

xZise:


LEGO1/lego/legoomni/src/entity/legoentity.cpp
```diff
@@ -263,23 +263,23 @@ void LegoEntity::ParseAction(char* p_extra)
 
 // FUNCTION: LEGO1 0x10010f10
 // FUNCTION: BETA10 0x1007ee87
-void LegoEntity::ClickSound(MxBool p_und)
+void LegoEntity::ClickSound(MxBool p_basedOnMood)
```
foxtacles, line 266: Can you change the param name in the header as well?

PR #1600, opened 2025-07-02
Clear unknown in `MxWavePresenter`

xZise:


LEGO1/omni/src/audio/mxwavepresenter.cpp
```diff
@@ -67,7 +67,7 @@ MxS8 MxWavePresenter::GetPlayedChunks()
 }
 
 // FUNCTION: LEGO1 0x100b1ba0
-MxBool MxWavePresenter::FUN_100b1ba0()
+MxBool MxWavePresenter::HasUnwrittenChunks()
```
foxtacles, line 70: I'd call this `ReadyForNextChunk`

PR #1601, opened 2025-07-02
Fix wrong notification type in `LegoRace::Notify`

foxtacles:
Fixing this bug: https://github.com/isledecomp/isle-portable/issues/481

`reccmp` didn't catch this, the function was 100%* before but used a wrong value in the switch-case. Something to look at? @disinvite @jonschz


AJenbo:
Are jump tables compared, they are not easy to analyze.

disinvite:
@jonschz Not at the moment. I thought you might want to have a look yourself this weekend. But I can add something today if you want.

jonschz:
No need to hurry, I'll take a look at the weekend. Just wanted to avoid both of us working on it in parallel. I'll assign the ticket to me.

PR #1623, opened 2025-07-13
Match `FUN_1006b140()` and related functions

### LINKED ISSUE #1357: Match `LegoAnimPresenter::FUN_1006b140`

jonschz:
Didn't significantly improve the match on `LegoAnimPresenter::FUN_1006b140()`, but I think that function is correct. There is only one minor real diff which is inlined from `Matrix4::Invert()`:
```diff
0x1006b31a : cmp edx, 4
-0x1006b31d : -jge 0x2f
-0x1006b31f : -lea ebx, [edi + edx*4]
+           : +jge 0x34
+           : +mov eax, edx
+           : +mov edx, dword ptr [ebp - 0x48]
+           : +lea ebx, [edx + eax*4]
0x1006b322 : shl ebx, 2
```
However, `Matrix4::Invert()` has a full structural match against BETA10 and I found no trivial changes around the affected location that had an impact on this issue. I think we can close #1357.

### DIFF
--- a/LEGO1/realtime/matrix4d.inl.h
+++ b/LEGO1/realtime/matrix4d.inl.h
@@ -290,6 +290,8 @@ void Matrix4::RotateZ(const float& p_angle)
 // FUNCTION: BETA10 0x1005a590
 int Matrix4::Invert(Matrix4& p_mat)
 {
+	// Inlined at LEGO1 0x1006b2d3
+
 	float copyData[4][4];
 	Matrix4 copy(copyData);
 	copy = *this;
@@ -312,6 +314,7 @@ int Matrix4::Invert(Matrix4& p_mat)
 		}
 
 		if (copy[i][i] < 0.001f && copy[i][i] > -0.001f) {
+			// FAILURE from mxtypes.h
 			return -1;
 		}
 
@@ -349,6 +352,7 @@ int Matrix4::Invert(Matrix4& p_mat)
 		}
 	}
 
+	// SUCCESS from mxtypes.h
 	return 0;
 }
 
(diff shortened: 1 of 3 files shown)


LEGO1/realtime/matrix4d.inl.h
```diff
@@ -312,6 +314,7 @@ int Matrix4::Invert(Matrix4& p_mat)
 		}
 
 		if (copy[i][i] < 0.001f && copy[i][i] > -0.001f) {
+			// FAILURE from mxtypes.h
```
jonschz, line 317: Not sure if we want to import `mxtypes.h` here, or move `SUCCESS` and `FAILURE` elsewhere, or leave it like this. I think importing `mxtypes.h` may cause a dependency cycle somewhere.

LEGO1/realtime/matrix4d.inl.h
```diff
@@ -349,14 +352,19 @@ int Matrix4::Invert(Matrix4& p_mat)
...
+// FUNCTION: BETA10 0x1005aa20
 void Matrix4::Swap(int p_d1, int p_d2)
 {
-	for (int i = 0; i < 4; i++) {
-		float e = m_data[p_d1][i];
+	// This function is affected by entropy even in debug builds
+	int i;
+	float e;
+	for (i = 0; i < 4; i++) {
+		e = m_data[p_d1][i];
```
jonschz, line 367: This had no meaningful impact on entropy (it was rather neutral), but it did improve the stack on BETA10.

foxtacles (approved):

PR #1630, opened 2025-07-16
Clear unknowns in `LegoROI`

xZise:


LEGO1/lego/sources/roi/legoroi.h
```diff
@@ -91,10 +96,10 @@ class LegoROI : public ViewROI {
 	// LegoROI::`scalar deleting destructor'
 
 private:
-	LegoChar* m_name;        // 0xe4
-	BoundingSphere m_sphere; // 0xe8
-	undefined m_unk0x100;    // 0x100
-	LegoEntity* m_entity;    // 0x104
+	LegoChar* m_name;             // 0xe4
+	BoundingSphere m_sphere;      // 0xe8
+	undefined m_definedElsewhere; // 0x100
```
foxtacles, line 101: That should probably be an MxBool.  I'd also call it `m_sharedLodList` since it seems to indicate this ROI gets its LODs from another ROI

xZise: Applied your suggestions

PR #1634, opened 2025-07-19
Improve `FUN_10061010`, other fixes

jonschz:
I tried to match `LegoAnimationManager::FUN_10061010` and got some improvements. I couldn't get one function to inline (specifically `MxListEntry<LegoTranInfo *>::MxListEntry<LegoTranInfo *>` around `LEGO1 0x1006120f`, called from `MxList::Append` -> `MxList::InsertEntry` -> `MxListEntry::MxListEntry`). Not sure if @foxtacles wants to take a look or if we just merge this. #1355 cannot be closed until that issue is fixed, unfortunately. The rest looks like it could match with a different stack layout.

### DIFF
--- a/LEGO1/lego/legoomni/include/legoanimationmanager.h
+++ b/LEGO1/lego/legoomni/include/legoanimationmanager.h
@@ -303,4 +303,7 @@ private:
 // TEMPLATE: LEGO1 0x10061750
 // MxListCursor<LegoTranInfo *>::MxListCursor<LegoTranInfo *>
 
+// TEMPLATE: BETA10 0x1004b5d0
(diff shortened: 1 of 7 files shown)


jonschz:
The interesting part is that `MxList` is actually part of the BETA, including the function I cannot get to inline. It may be that the code block that differs is itself part of an inlined function, though - haven't experimented with that yet.

foxtacles:
The MxList code is likely correct, because we have a lot of other places that confirm its correctness. The incorrect inlining doesn't just depend on that but also the code surrounding it

foxtacles (approved):

PR #1639, opened 2025-07-19
Build `BETA10` in CI

### LINKED ISSUE #1637: Add CI for BETA build

disinvite:
Fixes #1637. Might want to double check the CMake options I used.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,6 +100,43 @@ jobs:
           build/LEGO1.DLL
           build/LEGO1.PDB
 
+  build-beta:
+    name: 'MSVC 4.20 (BETA10)'
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - uses: actions/checkout@v4
+      with:
+        repository: itsmattkc/msvc420
+        path: msvc420
+
+    - name: Setup cmake
+      uses: jwlawson/actions-setup-cmake@v2
+      with:
+        # Use minimum supported version
+        cmake-version: '3.15.x'
+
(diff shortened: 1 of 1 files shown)


foxtacles (approved):

.github/workflows/build.yml
```diff
@@ -100,6 +100,43 @@ jobs:
...
+
+    - name: Patch MSVC 4.2
+      run: |
+        tools/patch_c2.py msvc420/bin/C2.EXE
+
+    - name: Build
+      shell: cmd
+      run: |
+        call .\msvc420\bin\VCVARS32.BAT x86
+        cmake -B build -DCMAKE_BUILD_TYPE=Debug -DISLE_INCLUDE_ENTROPY=OFF -DISLE_BUILD_BETA10=ON -DISLE_BUILD_LEGO1=OFF -DISLE_BUILD_APP=OFF -DISLE_BUILD_CONFIG=OFF -G "NMake Makefiles"
```
jonschz, line 129: I was surprised to learn that this runs without ` -DISLE_USE_SMARTHEAP=OFF`. Also, any reason why you don't use `-DCMAKE_BUILD_TYPE=RelWithDebInfo`? I kept using it from LEGO1 but didn't really question it.

disinvite: The function counts (for me) are 3475 for Debug and 2155 for RelWithDebInfo. It must be dropping all the inlined functions if you don't build with debug.

In the current CMakeLists file we don't try to link SH with `BETA10`, so `DISLE_USE_SMARTHEAP=OFF` must not be needed any more.

PR #1650, opened 2025-07-23
Clear remaining unknown in `Ambulance`

xZise:


LEGO1/lego/legoomni/src/actors/ambulance.cpp
```diff
@@ -41,7 +41,7 @@ Ambulance::Ambulance()
 	m_atBeachTask = 0;
 	m_taskState = Ambulance::e_none;
 	m_lastAction = IsleScript::c_noneIsle;
-	m_unk0x172 = 0;
+	m_enableRandomAudio = 0;
```
xZise, line 44: I'm wondering whether it should use `FALSE` and `TRUE` instead of `0` and `1` even though the type is 2 bytes long?

PR #1658, opened 2025-07-25
Clear unknowns in `MxSmkPresenter` and `MxLoopingSmkPresenter`

xZise:


LEGO1/omni/src/video/mxloopingsmkpresenter.cpp
```diff
@@ -39,7 +39,7 @@ void MxLoopingSmkPresenter::Destroy(MxBool p_fromDestructor)
 }
 
 // FUNCTION: LEGO1 0x100b4a00
-void MxLoopingSmkPresenter::VTable0x88()
+void MxLoopingSmkPresenter::FixCurrentFrame()
```
foxtacles, line 42: I don't really understand this name. How is it fixing anything? It seems more like it's resetting to the first frame.

xZise: Yeah, I used "fix" to say "fix `m_currentFrame` if required". Maybe not the best name, but unsure what a better name could be. Maybe `CheckAndResetCurrentFrame`?

xZise: I personally don't like the "two-verb"-names, so maybe `ResetCurrentFrameAtEnd` is better?

PR #1660, opened 2025-07-28
Clear unknowns in `Isle`

xZise:


LEGO1/lego/legoomni/src/worlds/isle.cpp
```diff
@@ -1056,11 +1056,11 @@ MxLong Isle::HandleTransitionEnd()
 }
 
 // FUNCTION: LEGO1 0x10032d30
-void Isle::FUN_10032d30(
+void Isle::LoadStartScript(
```
foxtacles, line 1059: I think we should find a better name for this one. What it seems to be doing is transition to the Island at a given camera location with a 2D overlay (i.e. elevator, any of the vehicles)

xZise: Fair point, although I have a hard time to come up with a better name. The "best" I got would be `SetupDestinationWIthOverlay`.

foxtacles: The script passed to this function is the 2D overlay bitmap, i.e.:

`IsleScript::c_SeaView_Background_Bitmap`

I think `TransitionToOverlayDestination` would be a good name

PR #1661, opened 2025-07-30
Cleared unknown in `Jetski`

xZise:


xZise:
Okay I have no idea, why the verification fails as I've just renamed a non-virtual method. Maybe somebody could give me a hint.

foxtacles:
This will be fixed soon, it's a reccmp issue

PR #1674, opened 2025-08-09
Clear unknowns in `LegoWorld`

xZise:


LEGO1/lego/legoomni/include/act3.h
```diff
@@ -106,7 +106,7 @@ class Act3 : public LegoWorld {
 	MxResult Tickle() override;               // vtable+0x08
 
 	// FUNCTION: LEGO1 0x10072500
-	MxBool VTable0x5c() override { return TRUE; } // vtable+0x5c
+	MxBool EnabledAfterDestruction() override { return TRUE; } // vtable+0x5c
```
foxtacles, line 109: That name doesn't make sense. The `LegoWorld` isn't destroyed. This attribute is queried in the destructor of the world presenter. The world presenter is destroyed as soon as its job is finished (world is loaded).

What this actually controls is whether 3D movement should be enabled in that moment. That happens when the function returns `FALSE`, not `TRUE`. So the name is incorrect in that sense as well.

Let's try to get an actual understanding of the code before naming things. A purely surface level descriptive naming isn't helpful here.

PR #1679, opened 2025-08-25
Clear unknowns in `LegoAct2`

xZise:


LEGO1/lego/legoomni/src/worlds/legoact2.cpp
```diff
@@ -938,169 +939,169 @@ MxResult LegoAct2::BadEnding()
...
-		local30 -= position;
-		Mx3DPointFloat local44 = local30;
-		local30.Unitize();
-		FUN_10052560(objectId, TRUE, TRUE, &vec, &local30, NULL);
+		Mx3DPointFloat lookingAtPepper(position);
+		Mx3DPointFloat positionPepper(m_pepper->GetWorldPosition());
+		lookingAtPepper -= positionPepper;
+		Mx3DPointFloat local44 = lookingAtPepper;
+		lookingAtPepper.Unitize();
+		StartAction(objectId, TRUE, TRUE, &positionPepper, &lookingAtPepper, NULL);
```
foxtacles, line 1086: This is still incorrect. You are introducing a bug here. Please carefully look at which arguments / variables you are passing compared to the original version. Double check everything else.

xZise: It should now be correct with starting the action at the actor's position which isn't being shadowed the original variable name.

PR #1681, opened 2025-09-06
Small BETA10 corrections

jonschz:
I noticed a few BETA10 annotation errors when working on the Ghidra import


LEGO1/lego/legoomni/src/paths/legopathstruct.cpp
```diff
@@ -19,7 +19,6 @@ DECOMP_SIZE_ASSERT(LegoPathStruct, 0x14)
 extern MxU32 g_isleFlags;
 
 // GLOBAL: LEGO1 0x100f119c
-// GLOBAL: BETA10 0x100f119c
```
jonschz, line 22: This was likely a mistake. I found no evidence that this global exists in BETA10.

Issue #1663, opened 2025-07-31
Explore `BETA10` diffs on overloaded thunks


disinvite:
This requires isledecomp/reccmp#181 to verify. That change provides thunks with a unique name (including the parameter string) if the parent is overloaded.

The list of new `BETA10` diffs is:
```
Increased (1):
0x1003bcbc - LegoAct2::FUN_100516b0 (39.77% -> 40.94%)

Decreased (36):
0x1000cb52 - Act2Actor::Animate(float) (57.49% -> 56.97%)
0x100142f1 - LegoAct2::FUN_100521f0 (69.97% -> 69.35%)
0x10018abf - Act3Cop::Animate(float) (64.62% -> 63.85%)
0x10018c6a - Act3Cop::FUN_10040360 (38.41% -> 38.08%)
0x1001a407 - Act3Brickster::FUN_100417c0 (49.27% -> 49.02%)
0x1001b017 - Act3Brickster::FUN_10042300 (74.15% -> 73.90%)
0x1001db2a - Act3Ammo::FUN_10053b40 (80.75% -> 79.81%)
0x10026ceb - Doors::ParseAction (93.20% -> 92.52%)
0x1002ef2f - Infocenter::InitializeBitmaps (35.84% -> 34.25%)
0x1003b049 - LegoAct2::Notify (34.19% -> 32.21%)
0x1003c990 - BoundingSphere::BoundingSphere(class BoundingSphere const &) (100.00% -> 95.45%)
0x1004152b - LegoAnimationManager::FUN_100605e0 (63.84% -> 63.50%)
0x10041a38 - LegoAnimationManager::FUN_100609f0 (87.60% -> 87.05%)
0x1004293c - LegoAnimationManager::Tickle (44.54% -> 44.26%)
0x10042f41 - LegoAnimationManager::FUN_10062110 (85.71% -> 84.95%)
0x10045daf - LegoAnimationManager::FUN_100648f0 (75.00% -> 72.32%)
0x10074e4f - LegoCharacterManager::CreateActorROI (51.05% -> 49.51%)
0x1007ebbe - LegoEntity::GetWorldPosition (100.00% -> 97.67%)
0x1008114a - LegoExtraActor::HitActor(class LegoPathActor *, unsigned char) (80.86% -> 80.60%)
0x100af35e - LegoPathActor::VTable0x68 (50.00% -> 49.00%)
0x100af899 - LegoPathActor::VTable0x9c (40.72% -> 40.39%)
[...]

PR #1682, opened 2025-09-06
Match Mx3DPointFloat overloads on BETA10

jonschz:
As discussed in #1663. We cannot close that ticket yet since there are other functions still to be matched.

There should be no regressions on BETA10 on this one, but a double check wouldn't hurt.

The constructors on `Vector3::Vector3` appear to be the right way round - switchting them causes major regressions with very few improvements.


foxtacles (approved):

jonschz:
>  Increased (1):
> 0x100605e0 - LegoAnimationManager::FUN_100605e0 (95.72% -> 100.00%)

A nice surprise I didn't even notice

PR #1688, opened 2025-09-21
Clear unknown parameter in `LegoEntity::SetROI`

xZise:


LEGO1/lego/legoomni/src/entity/legoentity.cpp
```diff
@@ -124,12 +124,12 @@ void LegoEntity::SetWorld()
 
 // FUNCTION: LEGO1 0x100108a0
 // FUNCTION: BETA10 0x1007e724
-void LegoEntity::SetROI(LegoROI* p_roi, MxBool p_bool1, MxBool p_bool2)
+void LegoEntity::SetROI(LegoROI* p_roi, MxBool p_bool1, MxBool p_updateTransform)
```
foxtacles, line 127: This is different from the header naming

xZise: Oh in that case, which name do you prefer? I personally think the other one doesn't really provide more information which is why I refactored the name to `p_updateTransform` but forgot that it doesn't update the header.

foxtacles: Sounds good

foxtacles:
I'd suggest adjusting all the SetROI overrides as well

PR #1696, opened 2025-11-02
BETA10: Fix reccmp regressions, match `RealtimeView`

jonschz:
I noticed that a number of BETA10 functions regressed since September due to changes in `reccmp` - more symbols were detected, which previously had less precise `OFFSET<>` matches.

I also added some matches for `RealtimeView` and found references to a `.cpp` file which was removed between BETA10 and LEGO1.

Todo:
- [x] Check for LEGO1 regressions

### DIFF
--- a/LEGO1/viewmanager/viewmanager.cpp
+++ b/LEGO1/viewmanager/viewmanager.cpp
@@ -605,7 +605,7 @@ inline undefined4 GetD3DRM(IDirect3DRM2*& d3drm, Tgl::Renderer* p_tglRenderer)
 {
 	assert(p_tglRenderer);
 	TglImpl::RendererImpl* renderer = (TglImpl::RendererImpl*) p_tglRenderer;
-	// TODO: Not thunked in BETA10
+	// Note: Diff in BETA10 (thunked in recompile but not in orig)
 	d3drm = renderer->ImplementationData();
 	return 0;
 }


LEGO1/viewmanager/viewmanager.cpp
```diff
@@ -9,19 +9,23 @@
...
 int g_planePointIndexMap[18] = {0, 1, 5, 6, 2, 3, 3, 0, 4, 1, 2, 6, 0, 3, 2, 4, 5, 6};
 
 // GLOBAL: LEGO1 0x10101050
+// GLOBAL: BETA10 0x10205914
 float g_LODScaleFactor = 4.0F;
 
 // GLOBAL: LEGO1 0x10101054
-float g_minLODThreshold = 0.00097656297;
+// GLOBAL: BETA10 0x10205918
+float g_minLODThreshold = 1.0000005F / 1024;
```
jonschz: same constant, but this value is more likely to be the original one

LEGO1/lego/legoomni/src/common/legoanimationmanager.cpp
```diff
@@ -2833,7 +2833,7 @@ void LegoAnimationManager::FUN_100648f0(LegoTranInfo* p_tranInfo, MxLong p_unk0x
 			actor->SetWorldSpeed(0.0f);
 		}
 
-		LegoLocation* location = NavController()->GetLocation(p_tranInfo->m_location);
+		const LegoLocation* location = NavController()->GetLocation(p_tranInfo->m_location);
```
jonschz: this fixes the use of the wrong `Vector3::Vector3()` in BETA10

PR #1697, opened 2025-11-13
cmake: support parallel MSVC420 builds by embedding debug information in the objects

madebr:
Only disadvantage is that the object files will be bigger. It remains possible to create a PDB file with all debug information

The reason parallel msvc420 were  impossible is because writing the shared build-time `*.pdb` file used naive file locking which fails for parallel jobs. Using `/Z7` instead of `/Zi` gets rid of the pdb.

https://learn.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format

I *think* this change does not impact accuracy, but please verify for yourself.

*HOWEVER*, it looks like reccomp does not support the (probably) slightly changed pdb.
Where before, a function was matched with 100% accuracy, it now fails when accessing a global value.

For example `LegoCacheSound::Stop` is 100% matched in current git master,
but using this PR it is 96.55% with this diff:
```diff
 mov eax, dword ptr [esi + 8]
 push eax 	(legocachsound.cpp:216)
 mov ebx, dword ptr [eax]
 call dword ptr [ebx + 0x48]
 mov byte ptr [esi + 0x58], 0 	(legocachsound.cpp:219)
 mov byte ptr [esi + 0x6a], 0 	(legocachsound.cpp:220)
 lea ecx, [esi + 0x10] 	(legocachsound.cpp:222)
 call Lego3DSound::Reset (FUNCTION)
 cmp word ptr [esi + 0x80], 0 	(legocachsound.cpp:223)
 je 0xd
-push "" (STRING)
+push <OFFSET2> 	(legocachsound.cpp:224)
 lea ecx, [esi + 0x74]
 call MxString::operator=(2) (FUNCTION)
 pop esi 	(legocachsound.cpp:226)
 pop ebx
 add esp, 4
 ret
```

If confirmed, I'll create an issue at the reccmp repo.


madebr:
I don't know enough about the PDB file format. Looking through the microsoft-pdb repository, there are lots of reasons a `EC_CORRUPT` ("Corrupt PDB") can be emitted.

Perhaps it's time to have a proper look at https://github.com/phire/simcopter_tool and see what can be re-used.
Maybe the error is recoverable.

PR #1756, opened 2026-06-10
Rename FUN_* symbols

Verociti:
Rename's FUN_* symbols.


DimaLeon2000:
Why bother closing a pull request without merging it beforehand?

PR #1761, opened 2026-08-30
Byte-identical reproduction of LEGO1.DLL, ISLE.EXE, and CONFIG.EXE

foxtacles:
This branch completes the byte-identity milestone: **LEGO1.DLL, ISLE.EXE, and CONFIG.EXE are reproduced byte-for-byte identical to the retail binaries from the checked-in source with the original MSVC 4.20 toolchain**, and CI proves it on every run.

## What's in here

**Source.** The decompiled source is restored toward its authentic 1997 form where the compiler's output demanded it: original string spellings (including the explicit trailing `\0` in the DirectDraw error tables), declaration shapes and ordering, include seating, and whitespace structure. Two functions (`LegoPathController::FindPath`, `MxVideoPresenter::PutFrame`, 572 bytes total) are not yet fully reproduced from source and are carried as explicitly disclosed, digest-pinned reference-byte exceptions.

**Verification.** The build is certified by [ReproBit](https://github.com/isledecomp/reprobit), a new library grown out of this work: it pins the toolchain (authenticated from the archaic-msvc repositories), controls the compiler-visible paths and build environment, applies each entropy intervention through reviewed, individually-proven operations, and verifies the result with a sealed byte-for-byte comparison the producer can never see. Project records live under `reprobit/`; every intervention carries a cost so remaining compromises stay visible (`rbit cost .`).

**CI.** The `ReproBit` workflow downloads and authenticates the retail references, provisions the authenticated compiler, rebuilds everything from scratch on native Windows, and fails unless all three binaries hash identical to retail. A comparison job then requires reccmp to score **every row of all three targets at 100% with every annotated function compared** (currently 4489 + 172 + 105 functions), and publishes the accuracy reports, SVGs, and the ReproBit verification report. The development builds (`Build` workflow) provision the same authenticated toolchain instead of a second MSVC checkout, and `tools/patch_c2.py` is retired (the provisioned C2 already carries the C4786 fix).

**reccmp.** The comparison gate depends on four upstream fixes (isledecomp/reccmp#523, #524, #526, plus the one-past-the-end/table handling they build on); until they merge, CI pins a fork branch carrying exactly those commits. Trailing-null string extents are handled with CSV data sources per the review of isledecomp/reccmp#525, and two PDB-invisible CRT statics are excluded via `report.ignore-functions` (their bytes remain covered by the byte-exact gate).

## Verifying locally

```console
rbit setup .          # once: fetch + authenticate MSVC 4.2
rbit verify . --report-dir build/reprobit-report
[...]

PR #1762, opened 2026-08-31
Annotations and repository metadata

foxtacles:
Updates the comments that tie this source code to the original game, plus a few repository settings. The compiled output does not change.

- The markers that link a function to its address in the original game now use the exact linker name instead of the human-readable one. Different tools print human-readable names differently, so the old markers sometimes failed to match; the exact name always matches.
- Addresses recovered from the 1997 beta build are recorded.
- Description-only headers are added for library code we do not compile ourselves (the C runtime, Smacker, SmartHeap and DirectX).
- The naming checker learns a few 1997 spellings that later steps bring back.
- The code formatter stops reordering `#include` lines, so a later step can restore the 1997 order.
- Git is told to keep the exact bytes of a few files.

---

**About this series:** step 1 of 12 (#1762–#1773). Together the twelve steps make the build produce the original LEGO1.DLL, ISLE.EXE and CONFIG.EXE byte for byte. Each step builds and passes the usual checks on its own; the last step (#1773) adds the check that the output is identical to the original files. See #1761 for the full story.

---


CONFIG/library_dx.h
```diff
@@ -0,0 +1,13 @@
+#if 0
```
disinvite: Future reccmp code parser will drop these marks if wrapped in an `#if 0`. (#1757)

tools/requirements.txt
```diff
@@ -1,3 +1,5 @@
-reccmp @ git+https://github.com/isledecomp/reccmp
+# Temporarily pinned to upstream master + isledecomp/reccmp#530 (LINE markers
+# inside function bodies); revert to isledecomp/reccmp once that PR merges.
+reccmp @ git+https://github.com/foxtacles/reccmp@c2dac1bcba99e30c229fc5670fe584c896f1a54c
```
disinvite: OK to revert

ISLE/library_msvc.h
```diff
@@ -344,3 +352,11 @@
...
 // GLOBAL: ISLE 0x4139c0
 // __acmdln
+
+/// Globals from libraries without symbols
+
+// STRING: ISLE 0x40f4a4
+static const char* ___crtLCMapStringA_str = "\0";
+
+// STRING: ISLE 0x40f4a8
+static const wchar_t* ___crtLCMapStringA_wstr = L"\0";
```
disinvite: String pooling probably prevents any issue, but FYI this will work as:
```cpp
// STRING: LEGO1 0x100dabb0
// "\0"

// STRING: LEGO1 0x100dabb4
// L"\0"
```

disinvite (approved):

CONFIG/library_dx.h
foxtacles: Dropped the `#if 0` wrapper — the file is never included, so it was inert anyway, and this keeps it consistent with #1757.

ISLE/library_msvc.h
foxtacles: Switched both to the comment form, thanks:

```cpp
// STRING: ISLE 0x40f4a4
// "\0"

// STRING: ISLE 0x40f4a8
// L"\0"
```

Verified the parser reads them as the same entities it read from the declarations (`0x40f4a4` narrow, `0x40f4a8` widechar, no alerts).

tools/requirements.txt
foxtacles: Reverted to `reccmp @ git+https://github.com/isledecomp/reccmp` and dropped the outdated comment.

The pin here carried exactly one commit over its merge-base (the LINE-marker fix), and upstream has that now as fbd0a16 (#530), so nothing else in this file depended on the fork. The still-open PRs (#523/#524/#526) are only needed by the ReproBit comparison job, which pins them inline in `reprobit.yml` rather than through this file.

Checked before reverting: with upstream reccmp installed clean, `reccmp-decomplint <target> --module <target> --warnfail` passes with 0 errors on CONFIG, ISLE and LEGO1.

PR #1763, opened 2026-08-31
Retail metadata: string extents and resources

foxtacles:
Small fixes so the data inside the binaries matches the original game.

- The DirectDraw error-message tables get their exact original endings back, and the comparison tool is told their true lengths.
- The Windows resource files (icons, version information) return to the original order with the original flags, and LEGO1's resource file is now part of the build.
- The library's declared name matches the original.
- Three C runtime functions use their original spelling.
- One keyboard data table cannot be checked by the comparison tool (isledecomp/reccmp#226), so it is documented and skipped.
- `ClearBackBuffers` declares its line counter next to the line pointer, which makes the compiler produce the original instructions.

---

**About this series:** step 2 of 12 (#1762–#1773). Together the twelve steps make the build produce the original LEGO1.DLL, ISLE.EXE and CONFIG.EXE byte for byte. Each step builds and passes the usual checks on its own; the last step (#1773) adds the check that the output is identical to the original files. See #1761 for the full story.

---


LEGO1/LegoOmni.def
```diff
@@ -1,6 +1,15 @@
...
+; The retail LEGO1.DLL's export directory names the module "Lego1.dll" in mixed
+; case, and retail ISLE.EXE imports it under that spelling; the all-caps file
+; name on the disc is the ISO 9660 master, not the linker's. Without this the
+; linker derives the export name from /OUT: and writes "LEGO1.DLL", which also
+; propagates into LEGO1.lib and so into ISLE.EXE's .idata.
+LIBRARY      Lego1.dll
+
+; LINK 4.20 copies a double-quoted DESCRIPTION into .rdata verbatim, quotes and
+; all; retail's string has none, so 1997 used the single-quoted form.
+DESCRIPTION  'Lego OMNI Windows Dynamic Link Library'
```
disinvite: I don't think this is interesting enough to warrant the AI slop explanation.

reccmp-project.yml
```diff
@@ -4,16 +4,30 @@ targets:
...
+        - _write_multi_char
+        - _write_string
   LEGO1:
     filename: LEGO1.DLL
     source-root: LEGO1
     hash:
       sha256: 14645225bbe81212e9bc1919cd8a692b81b8622abb6561280d99b0fc4151ce17
+    data-sources:
+      # DirectDraw error strings are spelled with an explicit trailing \0 in
+      # the 1997 source; automated string search cannot see the extra byte.
```
disinvite: No explanation needed. We could have just marked all the strings in place with `// STRING` and not used the CSV. If we want that instead, it should be easy enough for claude to do

CONFIG/reccmp.csv
```diff
@@ -0,0 +1,2 @@
+address|type|name|size|symbol
+0x407010|global|___xi_a|4|___xi_a
```
disinvite: I'm reading in order so I don't know if future PRs in the stack use this file more. If not, this should go in one of the `.h` files for MSVC markers

LEGO1/LegoOmni.def
foxtacles: Removed. Trimmed it to a single line noting the mixed-case spelling.

reccmp-project.yml
foxtacles: Comments removed.

On the CSV: in-place `// STRING` markers would work for most of these, but ~20 of the 91 DirectDraw strings are wrapped across two source lines, e.g.

```cpp
return "Informs DirectDraw that the previous Blt which is transfering information to or from this Surface is "
       "incomplete.\0";
```

`get_string_contents()` reads the first literal on the marked line only, so a marker there would record the first fragment and fail the string check. The trailing `\0` itself is fine for markers (`literal_eval` keeps it and the size comes out right) — it is only the wrapped literals that need the CSV. Happy to migrate if reccmp learns to join adjacent literals.

CONFIG/reccmp.csv
foxtacles: Good catch — nothing later in the stack adds to it. Moved to `CONFIG/library_msvc.h` as a `// GLOBAL: CONFIG 0x407010` / `// ___xi_a` marker pair, matching how `ISLE/library_msvc.h` and `LEGO1/library_msvc.h` declare the same CRT global, and deleted `CONFIG/reccmp.csv`.

disinvite (approved):

LEGO1/library_msvc.h
```diff
@@ -1051,10 +1051,12 @@
...
 // CrtMessageWindow
 
-// Cannot be handled right now due to anonymous pointer in struct.
-// We can annotate it on the original side, but we have no symbol on the recomp side.
-// We would need a way of annotating "the pointer at c_dfDIKeyboard+0x14 has orig address 0x10097f80".
-// // GLOBAL: LEGO1 0x10098f80
+// The rgodf member points at each image's own copy of the keyboard object
+// data, so datacmp cannot compare the variable and skips it through
+// report/ignore-variables in reccmp-project.yml.
+// https://github.com/isledecomp/reccmp/issues/535
```
jonschz: wrong issue (this is the closed duplicate)

foxtacles: Fixed — both references now point at isledecomp/reccmp#226. #535 was closed as a duplicate of it after I filed it.

PR #1764, opened 2026-08-31
Core library restructure and inline emission topology

foxtacles:
Puts the source files of the core libraries back into the arrangement the original developers used in 1997.

The compiler's output depends on which functions sit together in one file, so the files have to match the original layout before the output can. Several files are merged or split to restore it: the 3D renderer files become one file, the edge classes merge, the view classes merge into their managers, the animation scene splits out of the animation file, and the spline code moves to the geometry library. A few small inline functions move to the headers where the original compiler first saw them, and ViewROI's accessor functions get their 1997 names back.

Three functions also declare a loop counter earlier (`LegoAnimNodeData::Read`, `LegoTextureInfo::Create` and `LoadBits`), which makes them compile to the original instructions.

---

**About this series:** step 3 of 12 (#1762–#1773). Together the twelve steps make the build produce the original LEGO1.DLL, ISLE.EXE and CONFIG.EXE byte for byte. Each step builds and passes the usual checks on its own; the last step (#1773) adds the check that the output is identical to the original files. See #1761 for the full story.

---


foxtacles:

PR #1765, opened 2026-08-31
Omni translation-unit topology

foxtacles:
The same idea as the previous step, for the engine library ("omni"): source files return to their 1997 arrangement.

Two small option classes move into `mxmain.cpp`, the scheduler moves into the thread file, `define.cpp` joins the library, and the stream providers become their own small library, as in the original build. `main.cpp` is removed because the original entry point came from the C runtime rather than from the game's own code; a description-only header documents it instead.

---

**About this series:** step 4 of 12 (#1762–#1773). Together the twelve steps make the build produce the original LEGO1.DLL, ISLE.EXE and CONFIG.EXE byte for byte. Each step builds and passes the usual checks on its own; the last step (#1773) adds the check that the output is identical to the original files. See #1761 for the full story.

---


foxtacles:

PR #1766, opened 2026-08-31
Legoomni and CONFIG translation-unit topology

foxtacles:
The same idea for the game-logic library ("legoomni") and the CONFIG tool: source files return to their 1997 arrangement.

The data tables for characters, plants and locations move back into the files that use them, the race and Act 2 actor variants merge into their parent files, the building entity merges into the buildings file, and CONFIG's about dialog merges into its main dialog file. The actor-name constants get their original `e_` spelling, and IslePathActor's spawn locations take their original in-class form.

Five functions are also adjusted so they compile to the original instructions: an unused pointer is removed (`Act3Brickster`), a pointer is declared where it is first assigned (`ReleaseAutoROI`), two loop counters are declared earlier (`LegoGameState::Save` and `Load`), and an unnecessary temporary is dropped (`Infocenter::PlayAction`).

---

**About this series:** step 5 of 12 (#1762–#1773). Together the twelve steps make the build produce the original LEGO1.DLL, ISLE.EXE and CONFIG.EXE byte for byte. Each step builds and passes the usual checks on its own; the last step (#1773) adds the check that the output is identical to the original files. See #1761 for the full story.

---


foxtacles:

PR #1767, opened 2026-08-31
Include-directive seating

foxtacles:
Only `#include` lines change in this step. They return to the 1997 order in every file where that was the only remaining difference, and two files get back an include the original had. No code changes.

---

**About this series:** step 6 of 12 (#1762–#1773). Together the twelve steps make the build produce the original LEGO1.DLL, ISLE.EXE and CONFIG.EXE byte for byte. Each step builds and passes the usual checks on its own; the last step (#1773) adds the check that the output is identical to the original files. See #1761 for the full story.

---


foxtacles:

PR #1768, opened 2026-08-31
Omni source-shape restoration

foxtacles:
Small rewrites in the engine library so the compiler produces exactly the original instructions.

- `MXIOINFO` inherits from the Windows type it was based on.
- Some default constructors are split into two, the way the 1997 code wrote them.
- Loop variables are declared before their loops, as pre-standard C++ required. Modern MSVC gets a flag so it accepts this too.
- Global variables return to their original order, and assertions from the beta build are restored.
- Three functions are reshaped: `PutFrame` names two temporaries before its last call, `MxRegion::AddRect` reuses one pointer instead of declaring three, and `MxDSBuffer::FUN_100c6fa0` reads through the address of its cursor and reads the chunk length twice, as the beta build shows. Each change makes the compiler choose the same registers and instruction order as the original.

---

**About this series:** step 7 of 12 (#1762–#1773). Together the twelve steps make the build produce the original LEGO1.DLL, ISLE.EXE and CONFIG.EXE byte for byte. Each step builds and passes the usual checks on its own; the last step (#1773) adds the check that the output is identical to the original files. See #1761 for the full story.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,10 @@ if (MSVC)
   # Visual C++ 4.2 -> cl version 10.2.0
   if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.0")
     set(MSVC_FOR_DECOMP TRUE)
+  else()
+    # The decompiled source preserves 1997 semantics where a for-loop
+    # variable leaks into the enclosing scope.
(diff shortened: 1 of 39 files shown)


foxtacles:

PR #1769, opened 2026-08-31
Legoomni source-shape restoration

foxtacles:
The same kind of small rewrites for the game-logic library.

- Assertions from the beta build are restored, read-only globals get `const` back, and names use their 1997 spelling.
- Three functions that were compiled into the original but never called are recovered (`ModelDbModel::Write`, `LegoTestTimerConsoleControl`, `ConvertRGBToHSV`).
- A few operand orders and member initialization orders are corrected; the byte-identical build proves them right.
- Some blank lines are added so the line numbers inside assertions match the original.
- Four functions are reshaped: `PizzeriaState` resets its five saved states with a plain loop, `RemovePresenter` declares its iterator inside the loop, `TriggerHitSound` writes each table lookup as one expression, and `BumpBouy::Notify` declares the user actor earlier.

---

**About this series:** step 8 of 12 (#1762–#1773). Together the twelve steps make the build produce the original LEGO1.DLL, ISLE.EXE and CONFIG.EXE byte for byte. Each step builds and passes the usual checks on its own; the last step (#1773) adds the check that the output is identical to the original files. See #1761 for the full story.

---

### DIFF
--- a/LEGO1/define.cpp
+++ b/LEGO1/define.cpp
@@ -17,6 +17,10 @@ const char* g_strATTACH_CAMERA = "ATTACH_CAMERA";
 // STRING: LEGO1 0x10102018
 const char* g_strAUTO_CREATE = "AUTO_CREATE";
 
+// GLOBAL: LEGO1 0x10102058
+// STRING: LEGO1 0x10102010
+const char* g_strBANK = "BANK";
+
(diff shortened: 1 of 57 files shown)


foxtacles:

PR #1773, opened 2026-08-31
ReproBit: certified byte-identical reproduction

foxtacles:
This final step adds the proof. A new ReproBit workflow rebuilds the game from scratch on every run and fails unless LEGO1.DLL, ISLE.EXE and CONFIG.EXE come out byte for byte identical to the original files. It also publishes the accuracy reports, the progress pages and the continuous release, and shows how the score and the number of build adjustments compare with master.

ReproBit is the tool that makes this possible. Where the committed source alone does not give the original bytes, it applies small, reviewed build adjustments (603 today, each with a recorded proof and a cost score so they can keep shrinking) and checks the result. Every byte of the three files is produced by the 1997 compiler; nothing is copied from the original binaries (#1347).

Also in this step: the build workflow installs the compiler the same way as the ReproBit workflow, the old sampling scripts and their pages deployment are removed, the developer build keeps its usual comparison as an informative report, and ReproBit is pinned to release 0.1.5 so a tool update cannot silently change the checks.

---

**About this series:** the last of 12 steps (#1762–#1773). The combined result of all twelve is identical to the branch in #1761, which has the full story and instructions for checking the build on your own machine.

---


foxtacles:

PR #1778, opened 2026-09-07
Merge the view manager and LOD helper files into their original files

foxtacles:
`viewlod.cpp` and `viewroi.cpp` merge into `viewmanager.cpp` and `legoroi.cpp`, the LOD inline functions move to `legolod.inl.h` and `viewlod.inl.h`, and ViewROI's LOD-level accessors are renamed: `GetToken` is proven by BETA10's assert text (`vr->GetToken() >= 0`), while `SetToken`, `m_token` and the `c_token*` constants follow by analogy and are not attested. Also brings back asserts and the lost `ViewManager::FlushBuffers` from BETA10.


LEGO1/viewmanager/viewlod.h
```diff
@@ -43,6 +38,10 @@ class ViewLOD : public LODObject {
 	void SetFlag(unsigned char p_flag) { m_flags |= p_flag; }
 	void ClearFlag(unsigned char p_flag) { m_flags &= ~p_flag; }
 
+	// FUNCTION: LEGO1 0x100a5e40
+	// FUNCTION: BETA10 0x10171bdf
```
disinvite, line 42: `// SYNTHETIC` here since we do not implement it?

LEGO1/lego/sources/roi/legoroi.cpp
```diff
@@ -614,6 +623,7 @@ LegoResult LegoROI::SetColorByName(const LegoChar* p_name)
 
 // FUNCTION: LEGO1 0x100a9410
 // FUNCTION: BETA10 0x1018b324
+#pragma optimize("y", off)
```
disinvite: Should these be disabled by an `#if` when we are not running the accuracy build?

disinvite (approved):

LEGO1/lego/sources/roi/legoroi.cpp
foxtacles: Thanks, this one slipped through. There shouldn't be any pragmas like that. I thought I already had a solution without it somewhere - let me fix that. Thanks!

LEGO1/viewmanager/viewlod.h
foxtacles: It is implemented, at the end of `viewmanager.cpp`. Its address is the lowest in that file, so a marker on the definition trips decomplint's order check, and moving the definition changes the code of `Pick`/`Remove`. Hence the by-name marker (like `SpawnLocation` in `islepathactor.h`); the compiler doesn't generate it, so not `SYNTHETIC`.

LEGO1/lego/sources/roi/legoroi.cpp
foxtacles: Gone from the source now. No source form reproduces this frame pointer under MSVC 4.2 (inline asm never disables frame-pointer omission there; try/catch, `_alloca` and `setjmp` leave bytes retail lacks), so the ReproBit records carry the directive into the effective source instead (ReproBit 0.1.6). The ordinary build compiles the function with frame-pointer omission, so no `#if` is needed.

PR #1784, opened 2026-09-07
Merge the race, actor and CONFIG dialog files into their original files

foxtacles:
`carrace.cpp` and `jetskirace.cpp` merge into `legorace.cpp`, `act2genactor.cpp` into `act2actor.cpp`, `buildingentity.cpp` into `buildings.cpp`, and CONFIG's about dialog into `MainDlg`. IslePathActor's spawn table becomes a table inside the class, as in the original, and the building table entry's `m_entity` field is renamed to `entity`.


LEGO1/lego/legoomni/include/islepathactor.h
```diff
@@ -26,65 +26,6 @@ class IslePathActor : public LegoPathActor {
 		c_spawnBit3 = 0x04
 	};
 
-	// SIZE 0x38
-	struct SpawnLocation {
```
disinvite, line 30: Why move this into the .cpp file? (I have a hunch as to the reason, but if claude answers I don't want to preload it with my explanation 😉)

LEGO1/lego/legoomni/include/islepathactor.h
```diff
@@ -144,4 +85,13 @@ class IslePathActor : public LegoPathActor {
...
 };
 
+// FUNCTION: LEGO1 0x1001a6f0
+// ??0SpawnLocation@@QAE@XZ
+
+// FUNCTION: LEGO1 0x1001b1b0
+// ??0SpawnLocation@@QAE@W4Area@LegoGameState@@PAVMxAtomId@@HPBDFMFMIW4Script@JukeboxScript@@@Z
+
+// FUNCTION: LEGO1 0x1001b230
+// ??4SpawnLocation@@QAEAAU0@ABU0@@Z
```
disinvite, line 95: Could these attach to the implementation sites in `islepathactor.cpp`?

LEGO1/lego/legoomni/include/islepathactor.h
foxtacles: Compiler state: `RegisterSpawnLocations` only matched once four more declarations preceded it after the `legovideomanager.h` include. Defining `SpawnLocation` at file scope here (its only user) supplies exactly those; measured in foxtacles/isle@783387e0.

LEGO1/lego/legoomni/include/islepathactor.h
foxtacles: They're inline members emitted at first use (`0x1001a6f0` sits right before `RegisterSpawnLocations`), so plain markers on the in-class definitions fail decomplint's order check, and decomplint rejects by-name markers in .cpp files (`BYNAME_FUNCTION_IN_CPP`). Hence the header. Same situation as `~ViewLOD` in #1778: https://github.com/isledecomp/isle/pull/1778#discussion_r3951281503

LEGO1/lego/legoomni/include/islepathactor.h
disinvite: Nice. For the record, I thought it was the opposite: distributing `SpawnLocations` to every actor translation unit caused entropy ripple effects.

disinvite (approved):

Issue #1613, opened 2025-07-09
STL `Vector` needs void constructor


disinvite:
We use the STL vector in three places:

- `vector<undefined*>` in `LegoAnimActorStruct`
- `vector<LegoAnimActorStruct*>` in `LegoAnimActor`
- `vector<const ROI*>` aka (`ROIList`) in `ViewManager`

In `BETA10`, there is a diff in the constructors for those three classes where we initialize the vector member:

```
0x10171cf5  -call <OFFSET3>
            +call RealtimeView::RealtimeView (FUNCTION)
0x10171cfa   mov byte ptr [ebp - 4], 1
0x10171cfe  -mov ecx, dword ptr [ebp - 0x18]
            +lea eax, [ebp - 0x10]
            +push eax
            +mov ecx, dword ptr [ebp - 0x14]
0x10171d01   add ecx, 0x18
0x10171d04  -call <OFFSET4>
            +call Vector<ROI const *>::Vector<ROI const *> (FUNCTION)
```

We're pushing a value into that function when we shouldn't be. It seems like we need a void constructor for `Vector` in `mxstl.h`, similar to the one for `List`:
```cpp
template<class _TYPE>
class List : public list<_TYPE, allocator<_TYPE> >
{
public:
  typedef List<_TYPE> _Myt;
  typedef allocator<_TYPE> _A;

  explicit List() : list<_TYPE, _A>()
  {}

  explicit List(size_type _N, const _TYPE& _V = _TYPE()) : list<_TYPE, _A>(_N, _V)
  {}
//...
```

Imitating those two functions in `Vector` improves the accuracy for those three constructors, but introduces some diff noise as with most changes to a header.

Can any C++ experts speak to the correctness of that approach?

PR #1786, opened 2026-09-07
Match the original omni source more closely

foxtacles:
`MXIOINFO` derives from `MMIOINFO`, the mxstl constructors take their 1997 overload form, loop variables are declared before their `for` loops as the old scoping rules required (modern MSVC gets `/Zc:forScope-`), globals return to their original order, asserts come back from BETA10, and `MxDSBuffer` and `PutFrame` keep the local variables the original code had.

### DIFF
--- a/LEGO1/mxstl/mxstl.h
+++ b/LEGO1/mxstl/mxstl.h
@@ -146,8 +146,10 @@ public:
   typedef Multiset<_K, _Pr> _Myt;
   typedef allocator<_K> _A;
 
-  explicit Multiset(const _Pr& _Pred = _Pr())
-    : multiset<_K, _Pr, _A>(_Pred)
+  explicit Multiset() : multiset<_K, _Pr, _A>()
+  {}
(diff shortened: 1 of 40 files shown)


LEGO1/mxstl/mxstl.h
```diff
@@ -168,7 +170,10 @@ class Vector : public vector<_TYPE, allocator<_TYPE> >
   typedef Vector<_TYPE> _Myt;
   typedef allocator<_TYPE> _A;
 
-  explicit Vector(const _A& _Al = _A()) : vector<_TYPE, _A>(_Al)
+  explicit Vector() : vector<_TYPE, _A>()
+  {}
```
disinvite, line 174: This probably resolves #1613

foxtacles: Yes!

Several other issues will be closed as well

disinvite (approved):

PR #1787, opened 2026-09-07
Match the original legoomni source: worlds, actors and races

foxtacles:
Asserts and names restored from BETA10, `const` restored on read-only tables, globals moved to the files that originally defined them, and small logic and member-initialization fixes confirmed by the byte-identical build. Covers the worlds, actors and race directories.

### DIFF
--- a/LEGO1/lego/legoomni/include/infocenter.h
+++ b/LEGO1/lego/legoomni/include/infocenter.h
@@ -31,7 +31,7 @@ public:
 		// e_9 = 9,
 		// e_10 = 10,
 		e_welcomeAnimation = 11,
-		e_exiting = 12,
+		e_quiting = 12,
 		e_playCredits = 13,
 		e_exitingToIsland = 14,
 		e_backToInfoAct1 = 15,
@@ -91,7 +91,7 @@ public:
 	Playlist m_returnDialogue[3];   // 0x20
 	Playlist m_leaveDialogue[3];    // 0x44
 	Playlist m_bricksterDialogue;   // 0x68
-	MxU32 m_state;                  // 0x74
+	MxU32 m_step;                   // 0x74
 	MxStillPresenter* m_letters[7]; // 0x78
 };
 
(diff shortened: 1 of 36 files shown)


LEGO1/lego/legoomni/include/registrationbook.h
```diff
@@ -46,15 +46,13 @@ class RegistrationBook : public LegoWorld {
...
-	undefined m_unk0xfd[3];           // 0xfd
-	MxStillPresenter* m_alphabet[26]; // 0x100
-	MxStillPresenter* m_name[10][7];  // 0x168
-	struct {
-		MxS16 m_letters[7];                // 0x00
-		MxS16 m_cursorPos;                 // 0x0e
-	} m_newName;                           // 0x280
+	MxS32 m_registerDialogueTimer;         // 0xf8
+	undefined m_unk0xfc;                   // 0xfc
+	undefined m_unk0xfd[3];                // 0xfd
```
disinvite, line 51: Did we usually pad these to the longest digit? Good opportunity to do it now if we still care about the style points. (In hindsight, we should have had a script that checked these.)

LEGO1/lego/legoomni/include/infocenter.h
```diff
@@ -31,7 +31,7 @@ class InfocenterState : public LegoState {
 		// e_9 = 9,
 		// e_10 = 10,
 		e_welcomeAnimation = 11,
-		e_exiting = 12,
+		e_quiting = 12,
```
disinvite, line 34: Is the spelling of "quiting" from an assert?

disinvite (approved):

LEGO1/lego/legoomni/include/infocenter.h
foxtacles: Yes. BETA10.DLL contains the assert string `m_state->m_step == InfocenterState::e_quiting` (from `\Lego\LegoOmni\Src\Worlds\Infocenter.cpp`), so the spelling is the original one.

LEGO1/lego/legoomni/include/registrationbook.h
foxtacles: There is no convention: only `isle.h` and `CONFIG/config.h` zero-pad (28 lines), while 31 headers mix 2- and 3-digit offsets unpadded (331 lines). This file was already unpadded on master; the diff only realigned the comment column. I would leave it here and introduce a convention later, with a script enforcing it in CI.

PR #1788, opened 2026-09-07
Match the original legoomni source: paths, entities and the model database

foxtacles:
Asserts and names restored from BETA10, unused code recovered from BETA10 (`ModelDbModel::Write`), and small logic and member-initialization fixes confirmed by the byte-identical build. Covers the paths and entity directories and modeldb.

### DIFF
--- a/LEGO1/lego/legoomni/src/paths/legoextraactor.cpp
+++ b/LEGO1/lego/legoomni/src/paths/legoextraactor.cpp
@@ -9,6 +9,8 @@
 #include "mxmisc.h"
 #include "mxtimer.h"
 
+#include <assert.h>
+
 DECOMP_SIZE_ASSERT(LegoExtraActor, 0x1dc)
 
 // GLOBAL: LEGO1 0x100f31d0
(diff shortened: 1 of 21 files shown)


LEGO1/lego/legoomni/src/paths/legoextraactor.cpp
```diff
@@ -41,21 +43,26 @@ LegoExtraActor::LegoExtraActor()
...
 	delete m_assAnim;
 	delete m_disAnim;
 }
 
 // FUNCTION: LEGO1 0x1002a720
+// FUNCTION: BETA10 0x10080b4c
 MxU32 LegoExtraActor::StepState(float p_time, Matrix4& p_transform)
 {
+	// GLOBAL: LEGO1 0x100d6b64
+	static const float g_hitAnimationDelay = 2000.0f;
```
disinvite, line 58: Does it only work as a function `static` instead of just a global `const`?

disinvite (approved):

LEGO1/lego/legoomni/src/paths/legoextraactor.cpp
foxtacles: Both compile the same for this function, but the function-local static is what keeps the rest of the file matching: at file scope, three later functions (`CheckPresenterAndActorIntersections`, `~_Tree`, `_Tree::erase`) come out differently, and placed directly above the function the operand order flips. We could move it to the top of the file to match act3ammo.cpp, but those functions would need re-matching.

PR #1789, opened 2026-09-07
Match the original legoomni source: managers and presenters

foxtacles:
Asserts and names restored from BETA10, unused code recovered from BETA10 (`ConvertRGBToHSV`), a reconstructed `LegoTestTimerConsoleControl` that stands in for code the shipped link discarded (explained in a comment), and small logic and member-initialization fixes confirmed by the byte-identical build. Covers the manager, presenter, audio, video, control and build sources.

### DIFF
--- a/LEGO1/lego/legoomni/src/common/legotesttimer.cpp
+++ b/LEGO1/lego/legoomni/src/common/legotesttimer.cpp
@@ -2,9 +2,12 @@
 
 #include "legoeventnotificationparam.h"
 #include "legoinputmanager.h"
+#include "legovideomanager.h"
+#include "realtime/matrix4d.inl.h"
 #include "misc.h"
 #include "mxnotificationparam.h"
 
+#include <conio.h>
 #include <stdio.h>
 
 // FUNCTION: BETA10 0x100d1030
@@ -153,4 +156,23 @@ MxLong LegoTestTimer::Notify(MxParam& p_param)
 	}
 
 	return 0;
+}
+
+// Reconstruction: not in BETA10 or the 1996 source. Retail's .reloc reservation (sized before
+// /OPT:REF) accounts for a discarded function with this switch table. Keys mirror Notify.
+void LegoTestTimerConsoleControl(LegoTestTimer* p_timer)
+{
+	while (!_kbhit()) {
+	}
+
+	switch (_getch()) {
+	case 's':
+	case 'S':
+		p_timer->ResetAtNextTick();
+		break;
+	case 'p':
+	case 'P':
+		p_timer->Print();
+		break;
+	}
 }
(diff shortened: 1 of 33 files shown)


LEGO1/lego/legoomni/src/common/legotesttimer.cpp
```diff
@@ -154,3 +157,20 @@ MxLong LegoTestTimer::Notify(MxParam& p_param)
 
 	return 0;
 }
+
+void LegoTestTimerConsoleControl(LegoTestTimer* p_timer)
+{
```
disinvite, line 164: No beta addr for this? Is it from the 96 source?

LEGO1/lego/legoomni/src/video/legovideomanager.cpp
```diff
@@ -448,10 +455,9 @@ void LegoVideoManager::DrawFPS()
 			m_arialFont = NULL;
 		}
 		else {
-			DWORD i;
 			char* ptr = (char*) surfaceDesc.lpSurface;
 
-			for (i = 0; i < surfaceDesc.dwHeight; i++) {
+			for (DWORD i = 0; i < surfaceDesc.dwHeight; i++) {
```
disinvite, line 460: not `MxU32`?

disinvite (approved):

LEGO1/lego/legoomni/src/common/legotesttimer.cpp
foxtacles: I think this was a test I added in order to satisfy a IAT import for the console functions that I couldn’t create in any other way. There is a similar instance of that in isleapp.cpp where the IAT proves there must have been an additional unused function (the linker strips the unused function, but the unused import remains). Let me double check on this one

foxtacles: Neither. BETA10 has no such function (only padding after `Notify`, and nothing in the beta calls `_kbhit`), and the 1996 source handles the same keys inside `Notify`. It's a reconstruction: LINK sizes `.reloc` before /OPT:REF strips unreferenced code, and retail's header is 7 relocations larger than its actual table, exactly this function's switch table. Removing it breaks byte identity. Comment added above it.

LEGO1/lego/legoomni/src/video/legovideomanager.cpp
foxtacles: Not new: master already had `DWORD i;` here. The PR only moves the declaration into the for-init, which is what puts `ptr` and `i` in the original's registers. DrawFPS isn't in BETA10 or the 1996 source, so either spelling is a guess; `DWORD` matches the `dwHeight` it's compared against.

PR #1803, opened 2026-09-10
Drop the last overlay edits of isle.cpp and act3.cpp

foxtacles:
This continues #1802 with the last edits repair could replace, and records what was tried and refused so the campaign can stop here.

**Edits.** isle.cpp and act3.cpp each lose their last declaration carrier; repair re-found the affected bodies (6 records updated). 58,678 → 58,553.

**What was tried and refused, so the campaign stops here.** Moving `FindSoundByKey`'s owning-key constructor (the form the file's overlay donor already compiles privately) into the real source: the declaration in the shared header shifts the line tables of legosoundmanager.cpp, whose `_Tree::_Erase` is a register-bijection record, a family without a measured-pin repair; repair stops there after fourteen passes. Retiring an expensive record (the 500-point mosaics of VTable0x30 and LinkEdgesAndFaces, the 250-point rewrite of FUN_10061010) so the widened discovery could re-author it: repair finds no fresh carrier state that reproduces those bodies and does not author rewriting records from scratch. Removing the remaining carriers of files that hold rewriting or mosaic records (legonavcontroller, legoroi, legolod, legoanimationmanager, legocharactermanager and others): the rewrite bodies move and the same gap applies. The remaining 300-odd effective-source edits pin data or section order, or carry functions no discovered state reproduces.

A cold `rbit verify` on the final tree is byte-identical for all three targets with a clean authenticity verdict.


foxtacles:
Folded into #1802 (same commit, one PR).

